PR #1, opened 2021-04-28
ADR 001: Umee Module

alexanderbez:


brentxu (commented):
The ADR looks good.  My main comments are around two areas:
1) ability to withdraw and re-delegate staking rewards.  One suggestion was to do this after a certain amount of epochs eg. withdraw and re-delegate staking rewards every 30-50 Epochs.   My understanding is that the UX of this would be a bit clumsy, though it would be very worth it for meToken users to be able to have some form of compounded proof of stake yield to better reflect real world continuous compounding.  

2) It would make sense for the meToken to mirrow/mimic the inflation that is happening with the staked Atoms.  eg. as staked Atoms are receiving staking rewards, the meToken uses some sort of oracle to also reflect the inflation of the staking rewards.  

The ADR in its current state already looks solid.  These two points I mentioned could be visited in the future, though I just wanted to voice them to make sure that we take them into consideration.

docs/architecture/adr-001-umee-module.md
```diff
@@ -0,0 +1,90 @@
...
+  `PrevPendingDelegationQueue`. This queue is locked as funds cannot be sent to
+  it. In addition, these are the funds that will be delegated to the source chain
+  during the current staking epoch.
+- Clear out `PendingDelegationQueue`.
+
+The validators in the current staking epoch, E<sub>1</sub>, will then construct,
+sign, and broadcast multisig transactions to the Umee network. Once enough
+transactions are received by the `x/umee` module, specifically 2/3 of total
+staking weight, the `x/umee` module will store the constructed multisig transaction
+and emit an event signalling a delegation transaction can be made to the source
```
fedekunze, line 63: to whom will the will the amounts be delegated? does the user define the validator address or is the address defined by the umee module? Is it a single validator or a "portfolio" of validators? Who is (are) this (these) validator(s) chosen?

docs/architecture/adr-001-umee-module.md
```diff
@@ -0,0 +1,90 @@
...
+module will perform the following:
+
+- Move the funds from the `PendingDelegationQueue` into another queue,
+  `PrevPendingDelegationQueue`. This queue is locked as funds cannot be sent to
+  it. In addition, these are the funds that will be delegated to the source chain
+  during the current staking epoch.
+- Clear out `PendingDelegationQueue`.
+
+The validators in the current staking epoch, E<sub>1</sub>, will then construct,
+sign, and broadcast multisig transactions to the Umee network. Once enough
```
fedekunze, line 60: what's the multisig threshold? what happens if a validator doesn't sign in the current epoch? are there any penalties / rewards for uptime in the signing process?

docs/architecture/adr-001-umee-module.md
```diff
@@ -0,0 +1,90 @@
...
+will perform the following:
+
+- Automatically send source ATOM tokens to the dedicated `ModuleAccount` pool
+  address instead of a user supplied address.
+- Mark the transfer's amount in a pending queue, `PendingDelegationQueue`, to be
+  later used for delegation on the source chain.
+- Track the sender and amount in the module's state.
+- Mint derivative meTokens and send them to the source sender's Umee account.
+
+At the beginning of a staking epoch, E<sub>1</sub> in `BeginBlock`, the `x/umee`
```
fedekunze, line 50: what's the proposed epoch value in blocks?

docs/architecture/adr-001-umee-module.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Consequences
+
+### Positive
+
+- Allows for cross-chain delegations in a decentralized manner.
+
+### Negative
+
+- Requires the maintenance of a validator controlled multisig account each epoch.
```
fedekunze, line 82: I assume this requires the a hot key too

docs/architecture/adr-001-umee-module.md
```diff
@@ -0,0 +1,90 @@
...
+a `x/umee` module that contains a dedicated `ModuleAccount` that represents the
+ATOM pool. In addition, we will implement a [custom IBC module](https://github.com/cosmos/ibc-go/blob/v1.0.0-alpha1/docs/custom.md).
+The custom IBC module will handle receiving custom packets on a custom Umee port.
+The custom module will be nearly identical to the ICS-20 module, except that it
+will perform the following:
+
+- Automatically send source ATOM tokens to the dedicated `ModuleAccount` pool
+  address instead of a user supplied address.
+- Mark the transfer's amount in a pending queue, `PendingDelegationQueue`, to be
+  later used for delegation on the source chain.
```
fedekunze, line 46: do we only support the cosmos hub? what happens if an external chain, say Iris, also connects (independently) to Umee and performs a transfer? I guess the chains supported by Umee will have to be registered in the parameters. Also we need to take into consideration the naming collision for the chain ID in this case

fedekunze: I guess we would also need to have a pending delegation queue by base denom type (ATOM, etc)

docs/architecture/adr-001-umee-module.md
```diff
@@ -0,0 +1,90 @@
...
+- Track the sender and amount in the module's state.
```
fedekunze, line 47: see note about delegation validator address

docs/architecture/adr-001-umee-module.md
```diff
@@ -0,0 +1,90 @@
...
+At the beginning of a staking epoch, E<sub>1</sub> in `BeginBlock`, the `x/umee`
+module will perform the following:
+
+- Move the funds from the `PendingDelegationQueue` into another queue,
+  `PrevPendingDelegationQueue`. This queue is locked as funds cannot be sent to
+  it. In addition, these are the funds that will be delegated to the source chain
+  during the current staking epoch.
+- Clear out `PendingDelegationQueue`.
+
+The validators in the current staking epoch, E<sub>1</sub>, will then construct,
```
fedekunze, line 59: Is this happening on the Hub or Umee?

docs/architecture/adr-001-umee-module.md
```diff
@@ -0,0 +1,90 @@
...
+sign, and broadcast multisig transactions to the Umee network. Once enough
+transactions are received by the `x/umee` module, specifically 2/3 of total
+staking weight, the `x/umee` module will store the constructed multisig transaction
```
fedekunze, line 62: staking weight of the validators signers of the multisig? or the total power? Can a validator not be part of the multisig?

fedekunze (commented):
Looks good, I left a bunch of questions for further clarification

docs/architecture/adr-001-umee-module.md
```diff
@@ -0,0 +1,90 @@
...
+Any new transfers during E<sub>1</sub> will be processed in the same manner during
+E<sub>2</sub>.
+
+## Consequences
+
+### Positive
+
+- Allows for cross-chain delegations in a decentralized manner.
+
+### Negative
```
fedekunze, line 80: I assume that the validator set can't be that large because of the multisig?

docs/architecture/adr-001-umee-module.md
alexanderbez: That is decided by Umee governance. @brentxu can explain more.

docs/architecture/adr-001-umee-module.md
alexanderbez: Yes, but there is no other choice AFAIK :-/

docs/architecture/adr-001-umee-module.md
alexanderbez: We will only support the Hub (for now). Multi-chain becomes much more complex.

docs/architecture/adr-001-umee-module.md
alexanderbez: Umee. All this logic happens on the umee-side.

docs/architecture/adr-001-umee-module.md
alexanderbez: 1/2. That's a good question. The multsig threshold will have to be power-based, but since we cannot actually construct a multsig like this, what we can do instead is the make the threshold 1/n. Then, on chain, we ensure 2/3 have signed. As for how and when it's created, each validator will need an account on the Hub with key `K` and they will use the same key `K` on umee for the multisig. When they create their validator on Umee, they will provide the account associated with `K`. This is just a rough idea.

3. Validators are changed each epoch, so if a validator wants to join, they'll need enough stake to enter the set in the _next_ epoch.

4. I haven't thought about this. We either do nothing or penalize them somehow via a slash or something. Or perhaps we keep a rolling window and slash if they don't sign "enough".

5. Rewards? No. Penalties? Perhaps. @brentxu have you thought about this?

docs/architecture/adr-001-umee-module.md
alexanderbez: Staking weight of the validators in the validator set, i.e. of total power. All validators have to be in the multisig per epoch. You cannot opt out.

docs/architecture/adr-001-umee-module.md
fedekunze: Makes sense, the only thing is that you somehow have to prevent other chains from doing the same (i.e only allow IBC connections with the umee module from the Hub)

docs/architecture/adr-001-umee-module.md
fedekunze: ack, it would be good to write that in the ADR

docs/architecture/adr-001-umee-module.md
fedekunze:
```suggestion
The Umee validators in the current staking epoch, E<sub>1</sub>, will then construct,
```

docs/architecture/adr-001-umee-module.md
fedekunze: 8 hours?

docs/architecture/adr-001-umee-module.md
brentxu: yep - Bez is correct - Umee governance votes determine which validators eg. they can decide 33% each to the top 3 validators.  Or some combination that takes into consideration other validators in the network.

docs/architecture/adr-001-umee-module.md
brentxu: Are there any examples of other networks that are staking in epochs? If not, would we be setting some sort of standard approach?

docs/architecture/adr-001-umee-module.md
brentxu: what do you think the limitations in size would be? could it feasibly expand in the future?

docs/architecture/adr-001-umee-module.md
khoslaventures: What are the pros and cons of having an epoch vs not having an epoch for pending delegations?

Why not just delegate immediately? Is this to synchronize on governance decisions?

If governance decisions don't change, it seems like you could delegate as soon as the next block?

docs/architecture/adr-001-umee-module.md
khoslaventures: There was an ADR for weighted multisig a while ago: https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-019-multisigs.md

This was back when the multisig was part of Tendermint.

Perhaps it makes sense to extend the following:
https://github.com/cosmos/cosmos-sdk/blob/master/crypto/types/multisig/multisignature.go

With the above design it seems like it would be possible to alter the weights of the multisig depending on staking power... However it would have to be deployed on the Hub and all other supported chains based on this design.

docs/architecture/adr-001-umee-module.md
khoslaventures: So when the validator joins, the multisig weights will have to be updated with a weighted multisig design. This seems really hard to do given that there are multisigs on every chain Umee supports.

It would be interesting to have a design that allows you to put a weighted multisig on Umee, but then proxies everything to other accounts on other chains. The concept would be:
- On Umee, I have a weighted multisig
- On Cosmos, I have an account that when it receives a threshold signature from the weighted multisig originated from Umee chain, it performs the action I want (i.e. Delegation)

docs/architecture/adr-001-umee-module.md
khoslaventures: If the multisig is weighted according to stake on Umee then essentially consensus security must be broken in order to compromise the multisig?

Reinforcements may be done since there's a chance that KMS cloud services will get FIPs 140-3 compliant and have ed25519 hardware signers in the cloud soon this year.

Wondering if it's possible to have a security model to prevent functionality other than delegations related transactions on the multisig and returning owed money to the user. Sort of like how proxy contracts work on Ethereum.

docs/architecture/adr-001-umee-module.md
khoslaventures: > We either do nothing or penalize them somehow via a slash or something

I think this is a safe default. Rolling window approach could also make sense, it all depends how much slack you'd want to give to the validators. If I understand correctly that would mean signing should give rewards, not signing means no reward, if not signing long enough then slashing.

docs/architecture/adr-001-umee-module.md
khoslaventures: Cosmos does not make use of cryptographic threshold signatures. This means for every m-of-n multisig there is m signatures. This would likely result in some big performance implications for consensus but ultimately it depends how often we are updating the delegation allocations of the multisig as well

Here is a [reference implementation](https://github.com/ZenGo-X/multi-party-eddsa/wiki/Aggregated-Ed25519-Signatures) for cryptographic threshold ed25519

docs/architecture/adr-001-umee-module.md
alexanderbez: Yes @khoslaventures that is the multi-sig we will use -- the standard multi-sig used by the SDK.

alexanderbez: Upon further thought, I think all validators will have to sign. We cannot do 1/n because that validator can simply just broadcast transactions themselves.

docs/architecture/adr-001-umee-module.md
alexanderbez: @brentxu There is no current SDK-based chain that uses epochs, but many plan to such as the Hub and Osmosis. I believe both will have epoch on the order of a few hours.

@khoslaventures  An epoch is required, you cannot do this without an epoch because without an epoch, the validator set can change at any block making it practically impossible to do the cross-chain delegation(s). In other words, it's required, even we interchain accounts finally land, I still believe epochs will be required.

docs/architecture/adr-001-umee-module.md
alexanderbez: The multisig is not weighted -- it's a simple n/n multisig. All validators are required to sign per epoch.

docs/architecture/adr-001-umee-module.md
```diff
@@ -0,0 +1,90 @@
...
+
+- April 28, 2021: Initial Draft (@alexanderbez)
+
+## Status
+
+Proposed
+
+## Context
+
+Umee is a Universal Capital Facility that can collateralize assets on one
```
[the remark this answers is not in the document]
alexanderbez: That's a question for @brentxu :) (I took this from the paper)

docs/architecture/adr-001-umee-module.md
alexanderbez: There are no weights. If the account already exists on the native ATOM chain, great, we do nothing. If it does not exist, it'll have to be created first, where the Umee network will send a tx to the native ATOM chain to create the account (by sending some funds to it). The multi-sig is not weighted, it's a simple n/n setup where all validators are required to sign.

docs/architecture/adr-001-umee-module.md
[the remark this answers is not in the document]
brentxu: Yes - it is pretty similar to cross chain money markets.  Though there are aspects of this protocol that are beyond money markets eg an ecosystem where different blockchains can interact with cross chain leverage.  A capital facility is like a borrowing facility used by central banks for extending credit.  Umee is a blockchain hub that can also extend leverage.  I'm open to figuring out the general messaging.  Thanks for pointing this out @khoslaventures

docs/architecture/adr-001-umee-module.md
khoslaventures: It seems like having a weighted multisig would be worth it and the sdk would have to be extended to support it - otherwise it would limit who can become a validator on the network

docs/architecture/adr-001-umee-module.md
khoslaventures: So let's say we go with the n/n approach - managing this multisig is still quite difficult across chains if you want validator set flexibility? Also with n of n you'd have to be stricter about who can become a validator because misoperations from anyone is opportunity cost and would stop progress.

It also seems dangerous, what happens if you have an n of n multisig and someone's key set up is destroyed? All the user funds are gone then because of one validator?

alexanderbez: > managing this multisig is still quite difficult across chains if you want validator set flexibility? 

What is being managed cross chain? There is a single validator set multisig per epoch which must exist on the native ATOM chain. Validator sets can change every epoch.

> Also with n of n you'd have to be stricter about who can become a validator because misoperations from anyone is opportunity cost and would stop progress.

Yes, if a single validator misses a tx, the delegation does not happen for that epoch. The chain progress normally though.

> It also seems dangerous, what happens if you have an n of n multisig and someone's key set up is destroyed? All the user funds are gone then because of one validator? 

I think we'd have the validators submit txs first, and then send funds to that account on the native ATOM chain once we received all signatures. Only then do we emit the event.

I get this process is not clean, but I cannot think of any other way to do cross-chain delegation. Do you have alternative ideas?

docs/architecture/adr-001-umee-module.md
alexanderbez: There is no limit to who can become a validator on Umee, but having power-based multi-sig would be nice.

docs/architecture/adr-001-umee-module.md
alexanderbez: Yes, the SDK would need to support this in order for us to use threshold signatures.

docs/architecture/adr-001-umee-module.md
```diff
@@ -0,0 +1,90 @@
...
+blockchain towards borrowing assets on another blockchain. The platform specializes
+in allowing staked assets from POS blockchains to be used as collateral for
+borrowing across blockchains. The platform uses a combination of algorithmically
+determined interest rates based on market driven conditions.
+
+For the initial MVP implementation of the Umee network, we require the ability
+for users to be able to send ATOM tokens to a dedicated pool in the Umee network
+via IBC. ATOM tokens deposited into the Umee pool will mint a derivative meToken
+in a one-to-one ratio, i.e. one ATOM mints one meToken derivative.
```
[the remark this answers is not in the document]
brentxu: Yeh thanks for pointing this out - we'll want to properly document naming conventions - pretty important for defi protocols considering all of the stuff that is moving around.

docs/architecture/adr-001-umee-module.md
khoslaventures: > What is being managed cross chain? There is a single validator set multisig per epoch which must exist on the native ATOM chain. Validator sets can change every epoch.

Let me rephrase - so there's epoch management on the Umee chain and a multisig that's managed on the ATOM chain by the Umee validators, which gets managed by the `x/umee` module if I understand correctly. If we have epochs, then validator set changes are easy. I am not sure how the multisig gets updated though through these epochs for staked assets, but the notion of an epoch is useful for synchronizing events to happen.

> I think we'd have the validators submit txs first, and then send funds to that account on the native ATOM chain once we received all signatures. Only then do we emit the event.

That makes sense. I think this fact is still scary:

> It also seems dangerous, what happens if you have an n of n multisig and someone's key set up is destroyed? All the user funds are gone then because of one validator?

Are weighted multisigs completely off the table? We'd have to add support to the SDK and Hub.

docs/architecture/adr-001-umee-module.md
```diff
@@ -34,54 +34,134 @@ synthetic ERC-20 meTokens.
...
+
+Note, we push the account into a queue via, `PushAddressToRegistrationQueue`, so
+that we can implement a hook where on packet ACK, the `x/umee` module can store
+and mark the account as registered<sup>1</sup>.
+
+Once the `ModuleAccount` pool address is registered, we are now able to perform
+cross-chain delegations on the destination ATOM chain. During each `EndBlock`
+execution in the `x/umee` module, we will construct a series of messages that will
+be executed on the source ATOM chain via `TryRunTx` as defined by the
+`Interchain Account` ICS standard. Finally, the `PendingDelegationQueue` will be
```
khoslaventures, line 91: I read through the interchain account standard and I didn't really see any mention of signers... 

all I see is a confusing paragraph on it from the spec

> It does not restrict how you can distinguish signers that was not made by the counterparty chain. But the most common way would be to record the account in state when the interchain account is registered and to verify that signers are recorded interchain account.

So does this mean that we still need to implement something that's effectively a weighted multisig? Is that what `ModuleAccount` is?

alexanderbez: Good question. **No, there no need for any multi-sig**. It's completely unnecessary. Validators are not responsible for signing anything in this context. Keep in mind, the multi-sig approach was always a temporary stop-gap approach until Interchain accounts became available.

Interchain accounts take care of authorization and authentication of cross-chain accounts and the messages they can execute. They do this through the Interchain account module, leveraging IBC's ports and channels.

That being said, a relayer is still responsible for signing and relaying the actual transactions. This is no different than any other IBC-type transaction.

fedekunze (approved):
ACK

Issue #11, opened 2021-08-04
e2e: enable min gas prices


alexanderbez:
Once https://github.com/PeggyJV/gravity-bridge/pull/121 is merged and released, update the e2e tests to set minimum gas prices per validator.

Issue #10, opened 2021-08-04
e2e: verify eth balances


alexanderbez:
Currently, in the e2e tests we have no direct way of getting the balance of a token contract for a particular address. In order to do this we'd either have to rely on an external tool or create ABI go bindings. I'd rather avoid the former and hope that the gravity client exposes a way to query a deployed ERC20 token balance.

Issue #8, opened 2021-08-04
deps: update to latest gravity and SDK versions


alexanderbez:
- Update gravity bridge to latest version compatible with SDK v0.43
- Update to latest SDK v0.43 RC release

PR #14, opened 2021-08-04
deps: update SDK and Gravity Versions

alexanderbez:
closes: #8
closes: #11 
closes: #10


app/app.go
```diff
@@ -213,7 +221,7 @@ func New(
 
 	base := baseapp.NewBaseApp(Name, logger, db, encodingConfig.TxConfig.TxDecoder(), baseAppOptions...)
 	base.SetCommitMultiStoreTracer(traceStore)
-	base.SetAppVersion(version.Version)
```
marbar3778, line 216: id recommend removing this until state sync is fixed.

alexanderbez: Thanks @marbar3778!!!

Issue #178, opened 2021-10-25
price-feeder: implement main loop


alexanderbez:
## Summary

Now that we have the basis of the `price-feeder` tool in place and scaffolded, we need need to implement the "oracle" part of it. Specifically, [`Oracle#Start`](https://github.com/umee-network/umee/blob/main/price-feeder/oracle/oracle.go#L36).

This method should create a blocking loop that essentially performs what Terra's main [`vote`](https://github.com/terra-money/oracle-feeder/blob/main/feeder/src/vote.ts#L301) loop does. In other words, we need to:

- continuously update prices via the provided set of `providers`
- get on-chain oracle params and determine what tx(s) we need to sign and broadcast

----

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #241, opened 2021-11-24
feat: Implement price feeder

adamewozniak:
## Description

closes: #178

----

### Author Checklist

I have...

### Reviewers Checklist

I have...


toteki:
Use `go mod tidy` to fix the failing CI check

price-feeder/oracle/client/client.go
```diff
@@ -0,0 +1,274 @@
...
+	"github.com/cosmos/cosmos-sdk/types/tx/signing"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	rpchttp "github.com/tendermint/tendermint/rpc/client/http"
+	tmjsonclient "github.com/tendermint/tendermint/rpc/jsonrpc/client"
+	umeeapp "github.com/umee-network/umee/app"
+	umeeparams "github.com/umee-network/umee/app/params"
+)
+
+type (
+	CosmosChain struct {
```
alexanderbez, line 27: - I would combine `CosmosChain` and `OracleClient` into a single structure.
- I would remove `Ctx` and `Factory` from the structure as that's a bit of a code-smell
  - `Context` and `TxFactory` should probably be generated on the fly via a method on `CosmosChain`. E.g. `CosmosChain#CreateContext` and `CosmosChain#CreateTxFactory`

price-feeder/oracle/client/client.go
```diff
@@ -0,0 +1,274 @@
...
+	Factory           tx.Factory
+}
+
+// Pre-vote and vote are separated out mainly for readability
+// Prevote doesn't need the timeout functionality that vote needs,
+// Because of the block timing validation on the node side
+
+// Ref : https://github.com/terra-money/oracle-feeder/blob/baef2a4a02f57a2ffeaa207932b2e03d7fb0fb25/feeder/src/vote.ts#L230
+
+func (oc OracleClient) BroadcastPrevote(msgs ...sdk.Msg) error {
```
alexanderbez, line 81:
```suggestion
// Pre-vote and vote are separated out mainly for readability.
// Pre-vote doesn't need the timeout functionality that vote needs,
// because of the block timing validation on the node side.
//
// Ref : https://github.com/terra-money/oracle-feeder/blob/baef2a4a02f57a2ffeaa207932b2e03d7fb0fb25/feeder/src/vote.ts#L230
func (oc OracleClient) BroadcastPrevote(msgs ...sdk.Msg) error {
```

price-feeder/oracle/client/client.go
```diff
@@ -0,0 +1,274 @@
...
+	return tx.BroadcastTx(oc.Ctx, oc.Factory, msgs...)
+}
+
+// Need to try this continually, since err is either timing issue or whitelist error
```
alexanderbez, line 85: I don't understand the godoc here.

price-feeder/oracle/client/client.go
```diff
@@ -0,0 +1,274 @@
...
+// Ref : https://github.com/terra-money/core/blob/746a15f1bd83d62cd284e4af9471dc58701b3e33/x/oracle/keeper/msg_server.go#L89
+
+func (oc OracleClient) BroadcastVote(nextBlockHeight int, timeoutHeight int, msgs ...sdk.Msg) error {
+
+	var err error
```
alexanderbez, line 90:
```suggestion
```

price-feeder/oracle/client/client.go
```diff
@@ -0,0 +1,274 @@
...
+	maxBlockHeight := nextBlockHeight + timeoutHeight
+	lastCheckHeight := nextBlockHeight - 1
+	height := 0
+
+	for height == 0 && lastCheckHeight < maxBlockHeight {
```
alexanderbez, line 95: Can we add a comment on what this loop is doing and what the timeouts mean? I feel uneasy about seeing random sleeps.

price-feeder/oracle/client/client.go
```diff
@@ -0,0 +1,274 @@
...
+
+		// set last check height to latest block height
+		lastCheckHeight = latestBlockHeight
+
+		// wait for indexing (not sure; but just for safety)
+		time.Sleep(500)
+
+		err = tx.BroadcastTx(oc.Ctx, oc.Factory, msgs...)
+
+		if err != nil {
```
alexanderbez, line 113: I don't get why we ignore some errors, panic on others. We need to avoid all panics. Errors should bubble up to the main caller.

price-feeder/oracle/client/client.go
```diff
@@ -0,0 +1,274 @@
...
+				panic(err)
+			}
+		}
+	}
+
+	return err
+
+}
+
+func (oc OracleClient) GetHeight() (int, error) {
```
alexanderbez, line 127: There's no need for this entire method. The RPC client already has a method to do this. E.g.: https://github.com/cosmos/cosmos-sdk/blob/master/client/rpc/block.go#L74

price-feeder/oracle/client/client.go
alexanderbez: Also, we can skip the `else` due to the error check.

price-feeder/oracle/client/client.go
```diff
@@ -0,0 +1,274 @@
...
+
+	if err != nil {
+		panic(err)
+	}
+
+	return height, nil
+
+}
+
+func NewCosmosChain(ChainID string,
```
alexanderbez, line 167: The constructor should be moved right below the type definition above 👍

price-feeder/oracle/client/client.go
```diff
@@ -0,0 +1,274 @@
...
+	chain.OracleAddr, err = sdk.AccAddressFromBech32(OracleAddrString)
+	chain.OracleAddrString = OracleAddrString
+	if err != nil {
+		return nil, err
+	}
+	chain.ValidatorAddr = sdk.ValAddress(ValidatorAddrString)
+	chain.ValidatorAddrString = ValidatorAddrString
+	chain.Encoding = umeeapp.MakeEncodingConfig()
+	// Static gas adjustment
+	chain.GasAdjustment = 1.15
```
alexanderbez, line 195: Is this not fetched from config?

alexanderbez (changes requested):
Nice @adamewozniak! I left some initial feedback on the first few files. Prior to reviewing the entire thing and adding too much noise to the PR, I review only a few things. Once we make progress on that, we can review the rest of it :)

price-feeder/config/config.go
```diff
@@ -59,6 +62,24 @@ type (
...
+
+	Keyring struct {
+		Backend string `toml:"backend" validate:"required"`
+		Pass    string `toml:"pass" validate:"required"`
+		Dir     string `toml:"dir" validate:"required"`
+	}
+
+	Rpc struct {
+		TMRPCEndpoint string `toml:"tmrpc_endpoint" validate:"required"`
+		GRPCEndpoint  string `toml:"grpc_endpoint" validate:"required"`
```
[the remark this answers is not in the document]
adamewozniak: Afaik we need this for the GetParams method to grab the VotePeriod, unless there's a better way to do this

price-feeder/go.mod
```diff
@@ -13,75 +13,112 @@ require (
...
 	github.com/stretchr/testify v1.7.0
+	github.com/tendermint/tendermint v0.34.14
 	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
 	gopkg.in/yaml.v2 v2.4.0
 )
 
 require (
+	github.com/umee-network/umee v0.0.0-20211104202437-f74129d075f4
+	google.golang.org/grpc v1.42.0
+)
```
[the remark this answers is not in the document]
adamewozniak: Good to know, ty

price-feeder/oracle/client/client.go
adamewozniak: Should probably ref https://github.com/terra-money/oracle-feeder/blob/baef2a4a02f57a2ffeaa207932b2e03d7fb0fb25/feeder/src/vote.ts#L230 or https://docs.terra.money/Reference/Terra-core/Module-specifications/spec-oracle.html#prevote-and-vote instead here to make more sense. The method has to perform re-tries because the voting has to occur in the voting period after the prevote, and sometimes will try this before next period begins

price-feeder/oracle/client/client.go
adamewozniak: Not yet, but I'll put it in there

price-feeder/oracle/client/client.go
adamewozniak: Got these timeouts from terra's feeder ( https://github.com/terra-money/oracle-feeder/blob/baef2a4a02f57a2ffeaa207932b2e03d7fb0fb25/feeder/src/vote.ts#L230 ) - seems like it takes time to be in the correct voting period. I added a retry here too ( why the error is ignored in this func ), but I'll remove that until we have the oracle module wired up 👍

price-feeder/config/config.go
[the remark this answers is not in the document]
alexanderbez: Ahh, indeed. Then yes, we'll need the Umee gRPC endpoint 👍

price-feeder/cmd/price-feeder.go
```diff
@@ -82,16 +84,51 @@ func priceFeederCmdHandler(cmd *cobra.Command, args []string) error {
...
-	oracle := oracle.New()
-
 	ctx, cancel := context.WithCancel(context.Background())
 	g, ctx := errgroup.WithContext(ctx)
 
+	// Set up chain
+
+	timeout, err := strconv.Atoi(cfg.RPC.RPCTimeout)
+	if err != nil {
+		trapSignal(cancel)
```
alexanderbez, line 94: We only need to call `trapSignal` once which we already do at the end of this function.

price-feeder/cmd/price-feeder.go
```diff
@@ -82,16 +84,51 @@ func priceFeederCmdHandler(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	oracle := oracle.New()
-
 	ctx, cancel := context.WithCancel(context.Background())
 	g, ctx := errgroup.WithContext(ctx)
 
+	// Set up chain
+
```
alexanderbez, line 91:
```suggestion
```

price-feeder/cmd/price-feeder.go
```diff
@@ -82,16 +84,51 @@ func priceFeederCmdHandler(cmd *cobra.Command, args []string) error {
...
+		cfg.Keyring.Dir,
+		cfg.Keyring.Pass,
+		cfg.RPC.TMRPCEndpoint,
+		time.Duration(timeout),
+		cfg.Account.Address,
+		cfg.Account.Validator,
+		cfg.GRPCEndpoint,
+		gasAdjustment,
+	)
+
```
alexanderbez, line 116:
```suggestion
```

price-feeder/cmd/price-feeder.go
```diff
@@ -82,16 +84,51 @@ func priceFeederCmdHandler(cmd *cobra.Command, args []string) error {
...
+	oracle := oracle.New(oracleClient)
+
 	g.Go(func() error {
+		// start the process that observes and publishes exchange prices
 		return startPriceFeeder(ctx, cfg, oracle)
 	})
 	g.Go(func() error {
-		return startPriceOracle(ctx, oracle)
+		// Starts the process that calculates oracle prices
+		// And also votes
```
alexanderbez, line 130:
```suggestion
		// start the process that calculates oracle prices and votes
```

price-feeder/cmd/price-feeder.go
```diff
@@ -165,10 +202,14 @@ func startPriceFeeder(ctx context.Context, cfg config.Config, oracle *oracle.Ora
 	}
 }
 
-func startPriceOracle(ctx context.Context, oracle *oracle.Oracle) error {
+func startPriceOracle(ctx context.Context, oracle *oracle.Oracle, cancel context.CancelFunc) error {
```
alexanderbez, line 205: `startPriceOracle` should not take a `CancelFunc`. It was originally just a stub, but `startPriceOracle` should look more like `startPriceFeeder`.

i.e.

```go
// ...

go func() {
    log.Info().Msg("starting price-feeder oracle...")
    srvErrCh <- oracle.Start(ctx)
}

// ...
for {
  select {
    case <-ctx.Done():
      // ...

    case err := <-srvErrCh:
      // ...
  }
}
```

price-feeder/config/config.go
```diff
@@ -41,6 +41,10 @@ type (
 	Config struct {
 		Server        Server         `toml:"server"`
 		CurrencyPairs []CurrencyPair `toml:"currency_pairs" validate:"required,gt=0,dive,required"`
+		Account       `toml:"account" validate:"required,gt=0,dive,required"`
```
alexanderbez, line 44: Let's not embed the types here, i.e. lets do `Account Account ...`

price-feeder/config/config.go
```diff
@@ -59,6 +63,28 @@ type (
...
+		Dir     string `toml:"dir" validate:"required"`
+	}
+
+	RPC struct {
+		TMRPCEndpoint string `toml:"tmrpc_endpoint" validate:"required"`
+		GRPCEndpoint  string `toml:"grpc_endpoint" validate:"required"`
+		RPCTimeout    string `toml:"rpc_timeout" validate:"required"`
+	}
+
+	Gas struct {
```
alexanderbez, line 85: I'm not sure if Gas warrants its own struct -- we can just probably add `GasAdjustment` directly to `Config`.

alexanderbez (changes requested):

price-feeder/cmd/price-feeder.go
adamewozniak: Ty, didn't know the appropriate way to handle with goroutines

price-feeder/cmd/price-feeder.go
```diff
@@ -82,21 +84,49 @@ func priceFeederCmdHandler(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	oracle := oracle.New()
-
 	ctx, cancel := context.WithCancel(context.Background())
 	g, ctx := errgroup.WithContext(ctx)
 
+	// listen for and trap any OS signal to gracefully shutdown and exit
+	trapSignal(cancel)
```
alexanderbez, line 91: curious, was there a reason to move this here?

price-feeder/cmd/price-feeder.go
```diff
@@ -82,21 +84,49 @@ func priceFeederCmdHandler(cmd *cobra.Command, args []string) error {
...
+
+	timeout, err := strconv.Atoi(cfg.RPC.RPCTimeout)
+	if err != nil {
+		return fmt.Errorf("failed to parse RPC timeout: %w", err)
+	}
+
+	gasAdjustment, err := strconv.ParseFloat(cfg.GasAdjustment, 32)
+	if err != nil {
+		return fmt.Errorf("failed to parse Gas Adjustment: %w", err)
```
alexanderbez, line 100:
```suggestion
		return fmt.Errorf("failed to parse gas adjustment: %w", err)
```

price-feeder/config/config.go
```diff
@@ -41,6 +41,10 @@ type (
 	Config struct {
 		Server        Server         `toml:"server"`
 		CurrencyPairs []CurrencyPair `toml:"currency_pairs" validate:"required,gt=0,dive,required"`
+		Account       Account        `toml:"account" validate:"required,gt=0,dive,required"`
+		Keyring       Keyring        `toml:"key_ring" validate:"required,gt=0,dive,required"`
+		RPC           RPC            `toml:"rpc" validate:"required,gt=0,dive,required"`
+		GasAdjustment string         `toml:"gas_adjustment" validate:"required"`
```
alexanderbez, line 47: Any reason we can't have `GasAdjustment` be a float64?

price-feeder/cmd/price-feeder.go
```diff
@@ -82,21 +84,49 @@ func priceFeederCmdHandler(cmd *cobra.Command, args []string) error {
...
 
-	oracle := oracle.New()
-
 	ctx, cancel := context.WithCancel(context.Background())
 	g, ctx := errgroup.WithContext(ctx)
 
+	// listen for and trap any OS signal to gracefully shutdown and exit
+	trapSignal(cancel)
+
+	timeout, err := strconv.Atoi(cfg.RPC.RPCTimeout)
```
alexanderbez, line 93: this should use `time.ParseDuration` 👍

price-feeder/cmd/price-feeder.go
adamewozniak: If it was below the `return fmt.Errorf` paths, it could've gone unused, which just seemed ugly

alexanderbez: Ok, I think this fine here too 👍

price-feeder/oracle/oracle.go
```diff
@@ -2,29 +2,38 @@ package oracle
...
+	logger          zerolog.Logger
+	closer          *pfsync.Closer
 	mtx             sync.RWMutex
 	lastPriceSyncTS time.Time
 	prices          map[string]sdk.Dec
+	oracleClient    *client.OracleClient
 }
 
-func New() *Oracle {
+func New(oc *client.OracleClient) *Oracle {
```
alexanderbez, line 32: We need to provide the `CurrencyPairs` from the config as another argument here and create a "reverse flattened" map of provider => exchange pairs, i.e. a `map[string]ExchangePair`.

price-feeder/oracle/oracle.go
```diff
@@ -33,15 +42,271 @@ func (o *Oracle) Stop() {
 	<-o.closer.Done()
 }
 
-func (o *Oracle) Start(ctx context.Context) {
+var denoms = []string{"ATOMUSDT"}
```
alexanderbez, line 45: Why do we need this?

price-feeder/oracle/oracle.go
```diff
@@ -33,15 +42,271 @@ func (o *Oracle) Stop() {
...
+
+// Should return a prices map that we can set
+// The oracle to. Needs to touch multiple providers,
+// And then average out.
+func GetPrices() (map[string]sdk.Dec, error) {
+
+	binanceProvider := provider.NewBinanceProvider()
+	krakenProvider := provider.NewKrakenProvider()
```
alexanderbez, line 53: A few thoughts on this one:

1. Godocs should start with the name, e.g. `// GetPrices ...`
2. We shouldn't assume anything about providers. See my previous on the constructor. The operator of the price-feeder provides what providers they want to use.

Based on (2) I imagine this method would completely change

alexanderbez (changes requested):

price-feeder/oracle/oracle.go
adamewozniak: I took a slightly different approach and used a `map[string][]string`, so I could just pass the string array into the `GetTickerPrices` method

price-feeder/oracle/oracle.go
```diff
@@ -33,15 +56,261 @@ func (o *Oracle) Stop() {
...
+				priceAverages[k] = sdk.NewDec(0)
+			}
+			priceAverages[k] = priceAverages[k].Add(price)
+			priceCounts[k]++
+		}
+	}
+
+	// Average out each one
+	for k, average := range priceAverages {
+		floatToMultiply := float64(1) / float64(priceCounts[k])
```
adamewozniak, line 106: @alexanderbez please tell me there's a better way to go from float64 -> sdk.Dec

price-feeder/oracle/oracle.go
```diff
@@ -33,15 +56,260 @@ func (o *Oracle) Stop() {
...
+		stringToMultiply := fmt.Sprintf("%f", floatToMultiply)
+		decToMultiply := sdk.MustNewDecFromStr(stringToMultiply)
+		average = average.Mul(decToMultiply)
```
toteki, line 108: Danger note - in general, sdk.Dec should be used for all floating point math, because of occasional (but fatal to determinism) differences in how some hardware handles floating points.

Assuming you just need to divide by priceCounts[k]
```suggestion
		average = average.QuoInt64(int64(priceCounts[k]))
```

price-feeder/oracle/oracle.go
toteki: @adamewozniak See multiline suggestion

price-feeder/oracle/oracle.go
adamewozniak: Exactly what I was looking for, tyvm!

price-feeder/cmd/price-feeder.go
```diff
@@ -82,21 +83,45 @@ func priceFeederCmdHandler(cmd *cobra.Command, args []string) error {
...
+		cfg.Account.Validator,
+		cfg.RPC.GRPCEndpoint,
+		cfg.GasAdjustment,
+	)
+	if err != nil {
+		return err
+	}
+
+	oracle := oracle.New(oracleClient,
+		cfg.CurrencyPairs)
```
alexanderbez, line 114:
```suggestion
	oracle := oracle.New(oracleClient, cfg.CurrencyPairs)
```

price-feeder/oracle/client/client.go
```diff
@@ -0,0 +1,248 @@
...
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	rpchttp "github.com/tendermint/tendermint/rpc/client/http"
+	tmjsonclient "github.com/tendermint/tendermint/rpc/jsonrpc/client"
+	umeeapp "github.com/umee-network/umee/app"
+	umeeparams "github.com/umee-network/umee/app/params"
+)
+
+// OracleClient defines a structure that interfaces
+// with the umee node
+type (
```
alexanderbez, line 25:
```suggestion
type (
  // OracleClient defines a structure that interfaces with the Umee node.
```

price-feeder/oracle/client/client.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	passReader struct {
+		pass string
+		buf  *bytes.Buffer
+	}
+)
+
+func NewOracleClient(ChainID string,
+	KeyringBackend string,
```
alexanderbez, line 51:
```suggestion
func NewOracleClient(
        chainID string,
	keyringBackend string,
```

arguments to functions and methods are not capitalized ;-)

price-feeder/oracle/client/client.go
```diff
@@ -0,0 +1,248 @@
...
+	KeyringDir string,
+	KeyringPass string,
+	TMRPC string,
+	RPCTimeout time.Duration,
+	OracleAddrString string,
+	ValidatorAddrString string,
+	GRPCEndpoint string,
+	GasAdjustment float64) (*OracleClient, error) {
```
alexanderbez, line 59:
```suggestion
	GasAdjustment float64,
) (*OracleClient, error) {
```

price-feeder/oracle/client/client.go
```diff
@@ -0,0 +1,248 @@
...
+		RPCTimeout:          RPCTimeout,
+		OracleAddr:          oracleAddr,
+		OracleAddrString:    OracleAddrString,
+		ValidatorAddr:       validatorAddr,
+		ValidatorAddrString: ValidatorAddrString,
+		Encoding:            umeeapp.MakeEncodingConfig(),
+		GasAdjustment:       GasAdjustment,
+		GRPCEndpoint:        GRPCEndpoint,
+	}, nil
+
```
alexanderbez, line 86:
```suggestion
```

price-feeder/oracle/client/client.go
```diff
@@ -0,0 +1,248 @@
...
+			return err
+		}
+
+	}
+
+	return nil
+
+}
+
+func (oc *OracleClient) CreateContext() (*client.Context, error) {
```
alexanderbez, line 172: This should non return a pointer -- contexts should not be mutable  👍

price-feeder/oracle/client/client.go
```diff
@@ -0,0 +1,248 @@
...
+		GenerateOnly:      false,
+		Offline:           false,
+		SkipConfirm:       true,
+	}
+
+	return &clientCtx, nil
+
+}
+
+func (oc *OracleClient) CreateTxFactory() (*tx.Factory, error) {
```
alexanderbez, line 231: Ditto -- we should not return a pointer.

price-feeder/oracle/client/client.go
```diff
@@ -0,0 +1,248 @@
...
+		OutputFormat:      "json",
+		UseLedger:         false,
+		Simulate:          false,
+		GenerateOnly:      false,
+		Offline:           false,
+		SkipConfirm:       true,
+	}
+
+	return &clientCtx, nil
+
```
alexanderbez, line 228:
```suggestion
	return clientCtx, nil
```

price-feeder/oracle/client/client.go
```diff
@@ -0,0 +1,248 @@
...
+		WithAccountRetriever(clientCtx.AccountRetriever).
+		WithChainID(oc.ChainID).
+		WithTxConfig(clientCtx.TxConfig).
+		WithGasAdjustment(oc.GasAdjustment).
+		WithGasPrices(oc.GasPrices).
+		WithKeybase(clientCtx.Keyring).
+		WithSignMode(signing.SignMode_SIGN_MODE_DIRECT).
+		WithSimulateAndExecute(true)
+
+	return &txFactory, nil
```
alexanderbez, line 247:
```suggestion
	return txFactory, nil
```

price-feeder/oracle/client/client.go
```diff
@@ -0,0 +1,248 @@
...
+
+		height, err = rpcClient.GetChainHeight(*ctx)
+		if err != nil {
+			return err
+		}
+
+	}
+
+	return nil
+
```
alexanderbez, line 169:
```suggestion
	return nil
```

price-feeder/oracle/client/client.go
```diff
@@ -0,0 +1,248 @@
...
+	return tx.BroadcastTx(*ctx, *factory, msgs...)
+}
+
+// Broadcast vote - tries to vote within the next voting period
+// Ref : https://github.com/terra-money/oracle-feeder/blob/baef2a4a02f57a2ffeaa207932b2e03d7fb0fb25/feeder/src/vote.ts#L230
+func (oc OracleClient) BroadcastVote(nextBlockHeight int64, timeoutHeight int64, msgs ...sdk.Msg) error {
+	maxBlockHeight := nextBlockHeight + timeoutHeight
+	lastCheckHeight := nextBlockHeight - 1
+	height := int64(0)
```
alexanderbez, line 131:
```suggestion

	var height int64
```

price-feeder/oracle/client/client.go
```diff
@@ -0,0 +1,248 @@
...
+
+	// Create Context, factory
```
alexanderbez, line 133:
```suggestion
```

price-feeder/oracle/client/client.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	factory, err := oc.CreateTxFactory()
+	if err != nil {
+		return err
+	}
+
+	// Re-try voting until timeout
+	for height == 0 && lastCheckHeight < maxBlockHeight {
+
```
alexanderbez, line 146:
```suggestion
```

price-feeder/oracle/client/client.go
```diff
@@ -0,0 +1,248 @@
...
+		latestBlockHeight, _ := rpcClient.GetChainHeight(*ctx)
```
alexanderbez, line 147: we should probably capture and return the error here, right?

price-feeder/oracle/client/client.go
```diff
@@ -0,0 +1,248 @@
...
+		err := tx.BroadcastTx(*ctx, *factory, msgs...)
+		if err != nil {
+			return err
+		}
+
+		height, err = rpcClient.GetChainHeight(*ctx)
+		if err != nil {
+			return err
+		}
+
```
alexanderbez, line 165:
```suggestion
```

alexanderbez (changes requested):
Nice work Woz, so so close!

price-feeder/oracle/oracle.go
```diff
@@ -33,15 +56,256 @@ func (o *Oracle) Stop() {
...
+			priceProvider = provider.NewKrakenProvider()
+		}
+
+		prices, err := priceProvider.GetTickerPrices(tickerList...)
+		if err != nil {
+			return err
+		}
+
+		providerPrices[providerName] = prices
+		wg.Done()
```
alexanderbez, line 83: I like your approach with the waitgroup! But a few things:

- wait groups are useful when used with goroutines (threads) -- we're not using any here.
- we should use [errgroup](https://pkg.go.dev/golang.org/x/sync/errgroup) instead since `GetTickerPrices` returns an error.
- So ultimately, this would look something like:

```go
func (o *Oracle) SetPrices() error {
	g := new(errgroup.Group)
	mtx := new(sync.Mutex)
	providerPrices := make(map[string]map[string]sdk.Dec)

	for providerName, currencyPairs := range o.providerPairs {
		var priceProvider provider.Provider

		switch providerName {
		case config.ProviderBinance:
			priceProvider = provider.NewBinanceProvider()

		case config.ProviderKraken:
			priceProvider = provider.NewKrakenProvider()
		}

		g.Go(func() error {
			ticker := make([]string, len(currencyPairs))
			for _, cp := range currencyPairs {
				ticker = append(ticker, cp.String())
			}

			prices, err := priceProvider.GetTickerPrices(ticker...)
			if err != nil {
				return err
			}

			// flatten and collect prices based on the base currency per provider
			//
			// e.g.: {ProviderKraken: {"ATOM": 34.03, "UMEE": 6.3}}
			mtx.Lock()
			for _, cp := range currencyPairs {
				if p, ok := prices[cp.String()]; ok {
					providerPrices[providerName][cp.Base] = p
				}
			}
			mtx.Unlock()

			return nil
		})
	}

	if err := g.Wait(); err != nil {
		return err
	}

	// consolidate the different provider maps into one for each exchange rate
	var (
		priceAverages = make(map[string]sdk.Dec)
		priceCounts   = make(map[string]int)
	)

	// TODO: Consider using Volume Weighted Average Price (VWAP).
	//
	// Ref: https://github.com/umee-network/umee/issues/251
	for _, prices := range providerPrices {
		for base, price := range prices {
			if _, ok := priceAverages[base]; !ok {
				priceAverages[base] = sdk.NewDec(0)
			}

			priceAverages[base] = priceAverages[base].Add(price)
			priceCounts[base]++
		}
	}

	for k, average := range priceAverages {
		average = average.QuoInt64(int64(priceCounts[k]))
	}

	o.prices = priceAverages

	return nil
}
```

Assume we have the following:

```go
type CurrencyPair struct {
  Base string
  Quote string
}

func (cp CurrencyPair) String() string {
  return cp.Base + cp.Quote
}
```

alexanderbez (changes requested):

price-feeder/oracle/oracle.go
```diff
@@ -2,29 +2,52 @@ package oracle
...
 
-func New() *Oracle {
+func New(oc *client.OracleClient,
+	CurrencyPairs []config.CurrencyPair) *Oracle {
+
+	providerPairs := make(map[string][]string)
+
+	for _, pair := range CurrencyPairs {
+		for _, provider := range pair.Providers {
+			providerPairs[provider] = append(providerPairs[provider], pair.Base+pair.Quote)
```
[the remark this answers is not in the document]
adamewozniak: Went with the struct : 
```go
	providerPairs[provider] = append(providerPairs[provider], CurrencyPair{
		Base:  pair.Base,
		Quote: pair.Quote,
	})
```

price-feeder/oracle/oracle.go
```diff
@@ -33,15 +66,281 @@ func (o *Oracle) Stop() {
...
+		average = average.QuoInt64(int64(priceCounts[k]))
+	}
+
+	o.prices = priceAverages
+
+	return nil
+}
+
+func (o *Oracle) GetParams() (*umeetypes.QueryParamsResponse, error) {
+	// Create a connection to the gRPC server.
```
alexanderbez, line 149:
```suggestion
```

price-feeder/oracle/oracle.go
```diff
@@ -33,15 +66,281 @@ func (o *Oracle) Stop() {
...
+	grpcConn, err := grpc.Dial(
+		o.oracleClient.GRPCEndpoint, // your gRPC server address.
```
alexanderbez, line 151:
```suggestion
		o.oracleClient.GRPCEndpoint,
```

price-feeder/oracle/oracle.go
```diff
@@ -33,15 +66,281 @@ func (o *Oracle) Stop() {
...
+		grpc.WithInsecure(),         // The Cosmos SDK doesn't support any transport security mechanism.
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	defer grpcConn.Close()
+	queryClient := umeetypes.NewQueryClient(grpcConn)
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
```
alexanderbez, line 161: I think a minute is way too high here. Let's do 15s.
```suggestion
	ctx, cancel := context.WithTimeout(context.Background(), 15 * time.Second)
```

price-feeder/oracle/oracle.go
```diff
@@ -33,15 +66,281 @@ func (o *Oracle) Stop() {
...
+	defer cancel()
+
+	queryResponse, err := queryClient.Params(ctx, &umeetypes.QueryParamsRequest{})
+	if err != nil {
+		return nil, err
+	}
+
+	return queryResponse, nil
```
alexanderbez, line 169: Maybe return `queryResponse.Params` instead. No need to return the wrapping response type ;-)

price-feeder/oracle/oracle.go
```diff
@@ -33,15 +66,281 @@ func (o *Oracle) Stop() {
...
+func (o *Oracle) generateSalt(length int) (string, error) {
+	if length == 0 {
+		return "", fmt.Errorf("can't generate empty salt")
+	}
+	n := length / 2
+	bytes := make([]byte, n)
+	if _, err := rand.Read(bytes); err != nil {
+		return "", err
+	}
+	return hex.EncodeToString(bytes), nil
```
alexanderbez, line 181:
```suggestion
func (o *Oracle) generateSalt(length int) (string, error) {
	if length == 0 {
		return "", fmt.Errorf("failed to generate salt: zero length")
	}
	
	n := length / 2
	bytes := make([]byte, n)
	
	if _, err := rand.Read(bytes); err != nil {
		return "", err
	}
	
	return hex.EncodeToString(bytes), nil
```

price-feeder/oracle/oracle.go
```diff
@@ -33,15 +66,281 @@ func (o *Oracle) Stop() {
...
+	SubmitBlockHeight int64
+}
+
+func NewPreviousPrevote() *PreviousPrevote {
+	return &PreviousPrevote{
+		Salt:              "",
+		ExchangeRates:     "",
+		SubmitBlockHeight: 0,
+	}
+}
```
alexanderbez, line 198: Let's move the type and constructor to the top of the file.

price-feeder/oracle/oracle.go
```diff
@@ -33,15 +66,281 @@ func (o *Oracle) Stop() {
...
+
+var previousPrevote *PreviousPrevote = nil
```
alexanderbez, line 200: `previousPrevote` and `previousVotePeriod` are private globals which are a big code smell. I know they did this in the JS version, but I think that's bad practice. It makes testing incredibly difficult and bugs are easy to occur.

Instead, can we move `previousVotePeriod` `previousPrevote` to be part of the `Oracle`?

price-feeder/oracle/oracle.go
```diff
@@ -33,15 +66,281 @@ func (o *Oracle) Stop() {
...
+
+func (o *Oracle) tick() error {
+	ctx, err := o.oracleClient.CreateContext()
+	if err != nil {
+		return err
+	}
+
+	err = o.SetPrices()
+	if err != nil {
```
alexanderbez, line 209: just a bit more tidy :P 

```suggestion
	if err := o.SetPrices(); err != nil {
```

alexanderbez (approved):
@adamewozniak I left a few more comments. Once those are addressed, we can merge this! I'm going to create a small followup PR just to tidy a few things up.

alexanderbez (approved):

Issue #99, opened 2021-09-13
x/leverage: Handle Liquidation 


alexanderbez:
## Summary

Ensure the `x/leverage` module supports liquidation of CDPs.

----

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #258, opened 2021-12-02
feat: Liquidation

### LINKED ISSUE #257: x/leverage: Improve events and logs from RepayAsset and Liquidate
## Summary

Improve logs and events emitted by functions in `x/leverage/keeper/msg_server.go`

## Problem Definition

1. Borrower repayment amount can be less than the `MsgRepay.Amount` if the submitted amount is greater than the total borrowed. (The transaction succeeds without overpaying.) Logs and events should reflect the amount actually repaid, not the larger amount attempted.
2. Liquidations also produce information not contained within `MsgLiquidate`. Namely, the reward amount (which is determined by exchange rates at the time of the transaction) is not in the original message but should be logged.

## Proposal

Additional return values can be added to `keeper.RepayAsset` and `keeper.Liquidate` which provide the desired information. They can then be used inside `msg_server.go` when composing the relevant events and logs.

----

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

toteki:
## Description

Liquidation message type and params
Core liquidation functionality
Oracle placeholders for computing token value and borrow limits

closes: #99
closes #257 

----

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/leverage/keeper/borrows.go
```diff
@@ -126,3 +126,45 @@ func (k Keeper) GetBorrowUtilization(ctx sdk.Context, denom string, totalBorrowe
...
+		if k.GetCollateralSetting(ctx, borrowerAddr, coin.Denom) {
+			collateral = collateral.Add(coin)
+		}
+	}
+
+	return collateral
+}
+
+// CalculateBorrowLimit uses the price oracle to determine the borrow limit (in USD) provided by
+// collateral sdk.Coins, using each token's uToken exchange rate and collateral weight.
```
toteki, line 147:
```suggestion
// collateral sdk.Coins, using each token's uToken exchange rate and collateral weight.
// An error is returned if any input coins are not uTokens or if value calculation fails.
```

x/leverage/keeper/keeper.go
```diff
@@ -270,3 +270,166 @@ func (k Keeper) GetCollateralSetting(ctx sdk.Context, borrowerAddr sdk.AccAddres
...
+
+	// Transfer uToken collateral reward from borrower to liquidator
+	if err := k.bankKeeper.SendCoins(ctx, borrowerAddr, liquidatorAddr, sdk.NewCoins(reward)); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// LiquidationParams allows for dynamic liquidation parameters based on collateral denomination,
```
toteki, line 403:
```suggestion
// LiquidationParams computes dynamic liquidation parameters based on collateral denomination,
```

x/leverage/types/params.go
```diff
@@ -67,3 +83,33 @@ func validateInterestEpoch(i interface{}) error {
...
+	if !ok {
+		return fmt.Errorf("invalid parameter type: %T", i)
+	}
+
+	if v.IsNegative() {
+		return fmt.Errorf("complete liquidation threshold cannot be negative: %d", v)
+	}
+	if v.GT(sdk.MustNewDecFromStr("0.1")) {
+		return fmt.Errorf("complete liquidation threshold cannot exceed 0.1: %d", v)
+	}
```
toteki, line 98:
```suggestion
```
Since this isn't hard limit (but strongly advisable) it may not belong in a Validation function.

x/leverage/keeper/borrows.go
```diff
@@ -25,7 +25,7 @@ func (k Keeper) GetBorrow(ctx sdk.Context, borrowerAddr sdk.AccAddress, denom st
 	return owed
 }
 
-// SetBorrow sets the amount borrowed by an address in a given denom to an amount
+// SetBorrow sets the amount borrowed by an address in a given denom
```
alexanderbez, line 28:
```suggestion
// SetBorrow sets the amount borrowed by an address in a given denom.
```

x/leverage/keeper/borrows.go
```diff
@@ -126,3 +126,46 @@ func (k Keeper) GetBorrowUtilization(ctx sdk.Context, denom string, totalBorrowe
...
+	limit := sdk.ZeroDec()
+
+	for _, coin := range collateral {
+		// Convert uToken collateral to baseAssets assets
+		baseAssets, err := k.ExchangeUTokens(ctx, coin)
+		if err != nil {
+			return sdk.ZeroDec(), err
+		}
+		// Get USD value of base assets
+		value, err := k.GetValue(ctx, baseAssets)
```
alexanderbez, line 159: let's use `Price` or `ExchangeRate` nomenclature instead 👍

toteki: In this case the functions in the file would be
```
k.Price(sdk.Coin)
k.TotalPrice(sdk.Coins)
```
(and some renaming of `k.EquivalentValue`) right?

Note that these are all "total USD value" functions which multiply by coin.Amount, not the "pure exchange rate" functions that are exposed by oracle and will be invoked in the body of what is currently called `k.GetValue`

alexanderbez: ACK. I like `TotalPrice` 👍

x/leverage/keeper/borrows.go
```diff
@@ -126,3 +126,46 @@ func (k Keeper) GetBorrowUtilization(ctx sdk.Context, denom string, totalBorrowe
...
+
+	fullBalance := k.bankKeeper.GetAllBalances(ctx, borrowerAddr)
+	for _, coin := range fullBalance {
+		// If a denom is a valid uToken and enabled as collateral by this borrower addr
+		if k.IsAcceptedUToken(ctx, coin.Denom) && k.GetCollateralSetting(ctx, borrowerAddr, coin.Denom) {
+			collateral = collateral.Add(coin)
+		}
+	}
+
+	return collateral
```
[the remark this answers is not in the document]
toteki: Not as far as I can tell - expected uses of this function other than for (display purposes, borrow limit calculation) are order insensitive.

x/leverage/keeper/borrows.go
```diff
@@ -126,3 +126,46 @@ func (k Keeper) GetBorrowUtilization(ctx sdk.Context, denom string, totalBorrowe
...
+
+// CalculateBorrowLimit uses the price oracle to determine the borrow limit (in USD) provided by
+// collateral sdk.Coins, using each token's uToken exchange rate and collateral weight.
+// An error is returned if any input coins are not uTokens or if value calculation fails.
+func (k Keeper) CalculateBorrowLimit(ctx sdk.Context, collateral sdk.Coins) (sdk.Dec, error) {
+	limit := sdk.ZeroDec()
+
+	for _, coin := range collateral {
+		// Convert uToken collateral to baseAssets assets
+		baseAssets, err := k.ExchangeUTokens(ctx, coin)
```
alexanderbez, line 154: I pushed some minor cleanup to `x/leverage/keeper/exchange_rate.go`. 

```suggestion
		baseAsset, err := k.ExchangeUTokens(ctx, coin)
```

x/leverage/keeper/keeper.go
```diff
@@ -270,3 +270,166 @@ func (k Keeper) GetCollateralSetting(ctx sdk.Context, borrowerAddr sdk.AccAddres
...
+	// Transfer uToken collateral reward from borrower to liquidator
+	if err := k.bankKeeper.SendCoins(ctx, borrowerAddr, liquidatorAddr, sdk.NewCoins(reward)); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// LiquidationParams computes dynamic liquidation parameters based on collateral denomination,
+// borrowed value, and borrow limit
```
alexanderbez, line 405:
```suggestion
// borrowed value, and borrow limit.
```

alexanderbez: Can we describe what we're actually returning since we're turning multiple values of the same type.

x/leverage/keeper/keeper.go
```diff
@@ -270,3 +270,166 @@ func (k Keeper) GetCollateralSetting(ctx sdk.Context, borrowerAddr sdk.AccAddres
...
+
+	// liquidation incentive is determined by collateral reward denom
+	liquidationIncentive, err := k.GetLiquidationIncentive(ctx, reward)
+	if err != nil {
+		return sdk.ZeroDec(), sdk.ZeroDec(), err
+	}
+
+	// close factor scales linearly between MinimumCloseFactor and 1.0,
+	// reaching max value when (borrowed / limit) = CompleteLiquidationThreshold
+	closeFactor := k.GetParams(ctx).MinimumCloseFactor
```
alexanderbez, line 422: call `k.GetParams(ctx)` once and store it. i.e. `params := k.GetParams(ctx)`

alexanderbez: Recall, all queries cost gas for the end user ;-)

x/leverage/keeper/keeper.go
```diff
@@ -270,3 +270,166 @@ func (k Keeper) GetCollateralSetting(ctx sdk.Context, borrowerAddr sdk.AccAddres
...
+	if repayment.Amount.GTE(borrowed.AmountOf(repayment.Denom)) {
+		repayment.Amount = borrowed.AmountOf(repayment.Denom)
+	}
+
+	// Repayment cannot exceed borrowed value * close factor
+	repayValue, err := k.GetPrice(ctx, repayment)
+	if err != nil {
+		return err
+	}
+	if repayValue.GTE(borrowValue.Mul(closeFactor)) {
```
alexanderbez, line 343: We call `borrowValue.Mul(closeFactor)` twice. Let's put `maxRepayValue := borrowValue.Mul(closeFactor)` outside this `if` statement and use that in both places.

x/leverage/keeper/keeper.go
```diff
@@ -271,3 +271,166 @@ func (k Keeper) GetCollateralSetting(ctx sdk.Context, borrowerAddr sdk.AccAddres
...
+		return err
+	}
+
+	// Apply liquidation incentive
+	reward.Amount = reward.Amount.ToDec().Mul(sdk.OneDec().Add(liquidationIncentive)).TruncateInt()
+
+	// Reward amount cannot exceed available collateral
+	if reward.Amount.GTE(collateral.AmountOf(rewardDenom)) {
+		// only pay what can be correctly compensated
+		repayment.Amount = repayment.Amount.Mul(collateral.AmountOf(rewardDenom)).Quo(reward.Amount)
```
alexanderbez, line 368: I find that modifying `repayment.Amount` in two different places to be confusing. Can we comment on why we do this and how it differs from above?

x/leverage/keeper/keeper.go
```diff
@@ -271,3 +271,166 @@ func (k Keeper) GetCollateralSetting(ctx sdk.Context, borrowerAddr sdk.AccAddres
...
+	key := types.CreateLoanKey(borrowerAddr, repayment.Denom)
+	if owed.IsZero() {
+		store.Delete(key) // Completely repaid
+	} else {
+		bz, err := owed.Marshal()
+		if err != nil {
+			return err
+		}
+		store.Set(key, bz) // Partially repaid
+	}
```
alexanderbez, line 394: We should move this entire thing into a keeper method, e.g. `UpdateBorrowed` or something

x/leverage/keeper/keeper.go
toteki: Comment updated, LMK if it's good this time around

x/leverage/keeper/keeper.go
toteki: Godoc updated for this function

x/leverage/types/tx.go
```diff
@@ -142,3 +142,44 @@ func (msg *MsgRepayAsset) GetSigners() []sdk.AccAddress {
...
+		Reward:     reward,
+	}
+}
+
+func (msg *MsgLiquidate) ValidateBasic() error {
+	liquidator, err := sdk.AccAddressFromBech32(msg.GetLiquidator())
+	if err != nil {
+		return err
+	}
+	if liquidator.Empty() {
```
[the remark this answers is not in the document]
toteki: Coinfirmed, `AccAddressFromBech32` would cover it. I'll remove any such checks.

x/leverage/keeper/keeper.go
toteki: Done, just had to make my existing `SetBorrow` handle the `amount == 0` case and then use it.

x/leverage/keeper/oracle.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+
+	"github.com/umee-network/umee/x/leverage/types"
+)
+
+// Price returns the USD value of an sdk.Coin containing base assets
+func (k Keeper) Price(ctx sdk.Context, coin sdk.Coin) (sdk.Dec, error) {
```
adamewozniak, line 11: Maybe GetUSDValue?

x/leverage/keeper/msg_server.go
```diff
@@ -187,21 +187,26 @@ func (s msgServer) RepayAsset(
 		return nil, err
 	}
 
-	if err := s.keeper.RepayAsset(ctx, borrowerAddr, msg.Amount); err != nil {
+	repaid, err := s.keeper.RepayAsset(ctx, borrowerAddr, msg.Amount)
+	if err != nil {
 		return nil, err
 	}
 
+	actualRepayment := sdk.NewCoin(msg.Amount.Denom, repaid)
```
adamewozniak, line 195: I think you're trying to differentiate between what was successfully repaid and what the msg wanted to be repaid `msg.Amount` 

```suggestion
	repaidCoin := sdk.NewCoin(msg.Amount.Denom, repaid)
```

x/leverage/keeper/msg_server.go
```diff
@@ -212,3 +217,56 @@ func (s msgServer) RepayAsset(
...
+	if err != nil {
+		return nil, err
+	}
+
+	repaid, reward, err := s.keeper.LiquidateBorrow(ctx, liquidatorAddr, borrowerAddr, msg.Repayment, msg.RewardDenom)
+	if err != nil {
+		return nil, err
+	}
+
+	actualRepayment := sdk.NewCoin(msg.Repayment.Denom, repaid)
```
adamewozniak, line 243: Same suggestion as above ^

x/leverage/keeper/msg_server.go
```diff
@@ -187,21 +187,26 @@ func (s msgServer) RepayAsset(
 		return nil, err
 	}
 
-	if err := s.keeper.RepayAsset(ctx, borrowerAddr, msg.Amount); err != nil {
+	repaid, err := s.keeper.RepayAsset(ctx, borrowerAddr, msg.Amount)
```
adamewozniak, line 190:
```suggestion
	successfullyRepaid, err := s.keeper.RepayAsset(ctx, borrowerAddr, msg.Amount)
```

adamewozniak (changes requested):

x/leverage/keeper/oracle.go
toteki: Currently it's been changed from `GetValue/GetTotalValue` to `Price/TotalPrice`, so that's probably how it'll stay for now.

x/leverage/keeper/msg_server.go
toteki: Changed to `repaidCoin` and/or `rewardCoin` everywhere.

x/leverage/keeper/msg_server.go
toteki: I think `repaid` should be sufficient in this case

x/leverage/keeper/keeper.go
```diff
@@ -271,3 +265,181 @@ func (k Keeper) GetCollateralSetting(ctx sdk.Context, borrowerAddr sdk.AccAddres
...
+	baseRewardDenom := k.FromUTokenToTokenDenom(ctx, rewardDenom)
+	liquidationIncentive, closeFactor, err := k.LiquidationParams(ctx, baseRewardDenom, borrowValue, borrowLimit)
+	if err != nil {
+		return sdk.ZeroInt(), sdk.ZeroInt(), err
+	}
+
+	// Get liquidator's available balance of base asset to repay
+	liquidatorBalance := k.bankKeeper.GetBalance(ctx, liquidatorAddr, repayment.Denom)
+	// Return specific error on zero liquidator balance
+	if liquidatorBalance.Amount.IsZero() {
```
alexanderbez, line 326: note, the x/bank module will never have a balance of zero for an address. So this check doesn't hurt, but it's not necessary.

x/leverage/keeper/keeper.go
```diff
@@ -271,3 +265,181 @@ func (k Keeper) GetCollateralSetting(ctx sdk.Context, borrowerAddr sdk.AccAddres
...
+		return sdk.ZeroInt(), sdk.ZeroInt(), sdkerrors.Wrap(types.ErrLiquidatorBalanceZero, liquidatorBalance.String())
+	}
+	// Repayment cannot exceed liquidator's available balance
+	repayment.Amount = sdk.MinInt(repayment.Amount, liquidatorBalance.Amount)
```
alexanderbez, line 330: wdyt about not modifying the input param `repayment` and just keeping track internally of `repaymentAmt` as an `sdk.Int`?

x/leverage/keeper/keeper.go
```diff
@@ -271,3 +265,181 @@ func (k Keeper) GetCollateralSetting(ctx sdk.Context, borrowerAddr sdk.AccAddres
...
+		ctx, liquidatorAddr,
+		types.ModuleName,
+		sdk.NewCoins(repayment),
+	); err != nil {
+		return sdk.ZeroInt(), sdk.ZeroInt(), err
+	}
+
+	// Store the remaining borrowed amount in keeper
+	owed := borrowed.AmountOf(repayment.Denom).Sub(repayment.Amount)
+	if err := k.SetBorrow(ctx, borrowerAddr, repayment.Denom, owed); err != nil {
```
alexanderbez, line 386: what about just passing an `sdk.Coin` here instead?

x/leverage/keeper/msg_server.go
alexanderbez: yeah `repaid` should be fine to convey context.

x/leverage/types/params.go
```diff
@@ -3,18 +3,23 @@ package types
...
 var (
-	KeyInterestEpoch = []byte("InterestEpoch")
+	KeyInterestEpoch                = []byte("InterestEpoch")
+	KeyCompleteLiquidationThreshold = []byte("CompleteLiquidationThreshold")
+	KeyMinimumCloseFactor           = []byte("MinimumCloseFactor")
 )
 
 var (
-	defaultInterestEpoch = int64(100)
+	defaultInterestEpoch                = int64(100)
```
alexanderbez, line 20: should we discuss these defaults with the team?

x/leverage/types/params.go
```diff
@@ -67,3 +83,30 @@ func validateInterestEpoch(i interface{}) error {
 	}
 	return nil
 }
+
+func validateLiquidationThreshold(i interface{}) error {
```
alexanderbez, line 87: Is there no maximum value?

x/leverage/types/params.go
```diff
@@ -67,3 +83,30 @@ func validateInterestEpoch(i interface{}) error {
...
+func validateMinimumCloseFactor(i interface{}) error {
+	v, ok := i.(sdk.Dec)
+	if !ok {
+		return fmt.Errorf("invalid parameter type: %T", i)
+	}
+
+	if v.IsNegative() {
+		return fmt.Errorf("minimum close factor cannot be negative: %d", v)
+	}
+	if v.GT(sdk.MustNewDecFromStr("1")) {
```
alexanderbez, line 108: You can use `sdk.OneDec()` 👍

x/leverage/keeper/oracle.go
```diff
@@ -0,0 +1,60 @@
+package keeper
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+
+	"github.com/umee-network/umee/x/leverage/types"
+)
+
```
alexanderbez, line 9: These methods are all tested via the keeper tests right? We probably should have direct and small tests `oracle_test.go` that test just these. Can we add those? We can update them later once we have a real oracle

alexanderbez (changes requested):
Looks great @toteki. Just a few small remarks.

x/leverage/types/params.go
toteki: fixed

x/leverage/types/params.go
toteki: We should - they've been picked somewhat arbitrarily at the moment.

x/leverage/types/params.go
toteki: There's no hard value at which it becomes invalid, though it would be unreasonable to put it above, say, `0.2`

Since that's more of a soft cutoff than a "math breaks at param=x" one, I left that out of the validator.

x/leverage/keeper/keeper.go
toteki: done

x/leverage/keeper/oracle.go
toteki: Small tests have been added in `keeper_test.go` for now. They are also used indirectly during the liquidate tests.

If I separate them into `oracle_test.go`,  should they be written to avoid being methods of `(s *IntegrationTestSuite)`? That was possible with `math_test.go`, for example.

x/leverage/keeper/keeper.go
toteki: That would work, the new `repaymentAmt` would take the place of `repayment.Amount` everywhere, e.g.

```go
// was repayValue, err := k.Price(ctx, repayment)
repayValue, err := k.Price(ctx, sdk.NewCoin(repaymentAmt, repayment.Denom))
```

Is this just for clarity?

I double-checked that the input `sdk.Coin` isn't acting like a pointer, so outside functions passing in a coin don't have to worry about the amount being modified. That would have been a problem in `msg_server.go`, where the both original (attempted) and final (successful) repayment amounts are used for logging.

alexanderbez: Yeah I just feel uneasy about modifying input parameters. I would attempt to use a local variable instead, either `Int` or `Coin`.

x/leverage/types/params.go
alexanderbez: Ok, let's create an issue and discuss during our call. I've updated the agenda too.

alexanderbez (approved):
ACK. Let's just cleanup `LiquidateBorrow` a bit (see comment). Nice work 🚀

x/leverage/types/params.go
toteki: Issue is created #276

adamewozniak (approved):

Issue #251, opened 2021-11-30
Price Feeder: Use Volume Weighted Average Price


alexanderbez:
## Summary

Currently, we just average the prices fetched from all the providers specified by the operator of a price-feeder process.

## Proposal

We should instead, average by Volume Weighted Average Price (VWAP) as ChainLink does. 

ref: https://www.investopedia.com/terms/v/vwap.asp

We would need to refactor the `Provider` interface to something like the following:

```go
type WeightedPrice struct {
	Price  sdk.Dec
	Volume sdk.Int
}

type Provider interface {
	// ...
	GetTickerPrices(ticker ...string) (map[string]WeightedPrice, error)
}
```

And then the oracle can use the volume in conjunction with the price in determining the average per base currency.

----

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #260, opened 2021-12-02
refactor: Price Feeder VWAP 

alexanderbez:
## Description

closes: #251 

----

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/price-feeder/oracle/provider/binance.go
+++ b/price-feeder/oracle/provider/binance.go
@@ -12,7 +12,8 @@ import (
 )
 
 const (
-	binanceBaseURL = "https://api.binance.com"
+	binanceBaseURL        = "https://api.binance.com"
+	binanceTickerEndpoint = "/api/v3/ticker/24hr"
 )
 
 var _ Provider = (*BinanceProvider)(nil)
@@ -73,7 +74,7 @@ func (p BinanceProvider) GetTickerPrices(tickers ...string) (map[string]TickerPr
 }
 
 func (p BinanceProvider) getTickerPrice(ticker string) (TickerPrice, error) {
-	path := fmt.Sprintf("%s/api/v3/ticker/24hr?symbol=%s", p.baseURL, ticker)
+	path := fmt.Sprintf("%s%s?symbol=%s", p.baseURL, binanceTickerEndpoint, ticker)
 
 	resp, err := p.client.Get(path)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


price-feeder/oracle/provider/binance.go
```diff
@@ -69,44 +72,49 @@ func (p BinanceProvider) GetTickerPrices(tickers ...string) (map[string]sdk.Dec,
 	return tickerPrices, nil
 }
 
-func (p BinanceProvider) getTickerPrice(ticker string) (sdk.Dec, error) {
-	path := fmt.Sprintf("%s/api/v3/ticker/price?symbol=%s", p.baseURL, ticker)
+func (p BinanceProvider) getTickerPrice(ticker string) (TickerPrice, error) {
+	path := fmt.Sprintf("%s/api/v3/ticker/24hr?symbol=%s", p.baseURL, ticker)
```
adamewozniak, line 76: Endpoint should probably be in the BinanceProvider struct

price-feeder/oracle/provider/kraken.go
```diff
@@ -91,7 +93,7 @@ func (p KrakenProvider) GetTickerPrices(tickers ...string) (map[string]sdk.Dec,
 		)
 	}
 
-	tickerPrices := make(map[string]sdk.Dec, len(tickers))
+	tickerPrices := make(map[string]TickerPrice, len(tickers))
 	for _, t := range tickers {
 		// TODO: We may need to transform 't' prior to looking it up in the response
```
adamewozniak, line 98: Not sure I understand this todo - do we still need to transform this?

alexanderbez: The TODO is because Kraken might not necessarily treat ATOM as ATOM. E.g. they treat BTC as XBTC: https://api.kraken.com/0/public/Ticker?pair=XBTUSD

So in the case of Bitcoin, the value provided to this function would be `BTCUSD`, however, the response contains `XXBTZUSD`, the lookup would fail! Hence, we'd need to hardcode a transformation from XXBTZUSD to BTCUSD and thus the TODO I added. The reason we haven't done anything yet, is because BTC is the only ticker I know of for Kraken that they do this for and we won't support BTC.

price-feeder/oracle/provider/binance.go
alexanderbez: Ehh I don't think there's must merit to adding it to the type, but we can define a const instead 👍

adamewozniak (approved):
only approving because of the commit messages

Issue #250, opened 2021-11-30
x/oracle: Cleanup


adamewozniak:
## Summary

Need to cleanup the x/oracle module to put it in a workable state before we can continue with implementation 

## Problem Definition

The current [oracle module](https://github.com/umee-network/umee/tree/main/x/oracle) in our codebase isn't organized according to the folder structure used by [peggy](https://github.com/umee-network/umee/tree/main/x/peggy) / [leverage](https://github.com/umee-network/umee/tree/main/x/leverage), and it contains code we won't need per proposed [ADR-006](https://github.com/umee-network/umee/pull/249)

## Proposal

1. Reorganize file structure, add `module.go` & `genesis.go` files
2. Remove unnecessary functions having to do with `Tobin Tax`, `MsgExchangeRate[Pre]Vote`, and other Terra-specific code that might be laying around there 
3. Update `genesis.go` with keeper functions that we still need 

----

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #261, opened 2021-12-03
feat: Oracle cleanup

adamewozniak:
## Description

Removes tobintaxery and updates `genesis.go` with keeper functions

closes: #250

----

### Author Checklist

I have...

### Reviewers Checklist

I have...


toteki (approved):
Looks good - the PR is mostly removals so it didn't take as long to review as I though it would at first glance.

x/oracle/genesis.go
```diff
@@ -9,12 +11,107 @@ import (
...
+	}
+
+	for _, av := range genState.AggregateExchangeRateVotes {
+		valAddr, err := sdk.ValAddressFromBech32(av.Voter)
+		if err != nil {
+			panic(err)
+		}
+
+		keeper.SetAggregateExchangeRateVote(ctx, valAddr, av)
+	}
```
toteki, line 58: Does this mean that genesis state contains initial prices, as a consequence of hard-coded votes from validator addresses?

proto/umee/oracle/v1beta1/genesis.proto
```diff
@@ -32,12 +31,4 @@ message FeederDelegation {
...
-}
-
-// TobinTax defines an denom and tobin_tax pair used in
-// oracle module's genesis state
-message TobinTax {
-  string denom     = 1;
-  string tobin_tax = 2
-      [(gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec", (gogoproto.nullable) = false];
-}
+}
```
alexanderbez, line 34: nit: newline.

You add the following to your VSCode config:

```json
"files.insertFinalNewline": true,
"files.trimFinalNewlines": true,
"files.eol": "\n"
```

and it'll do everything for you on save 👍

x/oracle/genesis.go
alexanderbez: I think the idea is this happens on export/import network upgrades. This process is old and somewhat antiquated, in favor of live upgrades via `x/upgrade`. So TLDR; in practice this code will never run except in maybe some tests cases.

x/oracle/keeper/reward.go
```diff
@@ -18,7 +18,7 @@ func (k Keeper) RewardBallotWinners(
 	ballotWinners map[string]types.Claim,
 ) {
 	rewardDenoms := make([]string, len(voteTargets)+1)
-	rewardDenoms[0] = types.MicroUmeeDenom
+	rewardDenoms[0] = types.UmeeDenom
```
alexanderbez, line 21: this actually need to be the base denom, i.e. `uumee`.

Just a quick note/recap on denominations in the state-machine:

- The state-machine only deals with and knows about "base-level" denominations, e.g. `uumee`, `uatom`, `uosmo`, etc...In other words, these "base-level" denoms are not divisible, they cannot be broken down any further.
- Denomination metadata exists in `x/bank` and such metadata is used to let users and clients know how to deal with and display "higher-level" denominations, e.g. `1 umee = 10^6 uumee (1000000uumee)`.
- The `x/oracle` module gets prices for high-level denominations, e.g. `atom` and thus we will use the metadata to do the conversions.

x/oracle/types/asset.go
```diff
@@ -6,16 +6,16 @@ import (
...
-	USDDenom        = "USD"
-	BlocksPerMinute = uint64(10)
-	BlocksPerHour   = BlocksPerMinute * 60
-	BlocksPerDay    = BlocksPerHour * 24
-	BlocksPerWeek   = BlocksPerDay * 7
-	BlocksPerMonth  = BlocksPerDay * 30
-	BlocksPerYear   = BlocksPerDay * 365
-	MicroUnit       = int64(1e6)
+	AtomDenom       string = "ATOM"
+	UmeeDenom       string = "UMEE"
```
alexanderbez, line 10: IDK how these are used yet in the module, but if they're used in regards to actually moving funds around, they need to be the "base-level" denoms, e.g. `uumee` and `uatom`.

Also, I doubt we need `AtomDenom` at all.

alexanderbez (changes requested):
LGTM! Just one quick note on the denoms 👍

x/oracle/genesis.go
adamewozniak: It contains the initial exchange rate and votes for them, but the votes don't turn into exchange rates until after the ballot is counted -- the exchange rates will always be at least one vote period behind the votes

alexanderbez (approved):

Issue #256, opened 2021-12-01
price-feeder: evaluate oracle ticker timeout


alexanderbez:
## Summary

Currently, we have a hard-coded 10ms tick timeout between each oracle prevote/vote execution in the price-feeder.

https://github.com/umee-network/umee/blob/main/price-feeder/oracle/oracle.go#L112

## Problem Definition

We need to evaluate if 10ms is too short or too long and what a reasonable value should be.

/cc @adamewozniak 

----

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #274, opened 2021-12-08
fix: oracle ticker timeout & retry logic 

adamewozniak:
## Description

Sets an appropriate timeout for the oracle, fixes the loop so it'll retry for both the votes and pre-votes, and adds some logging for the oracle client 

closes: #256

----

### Author Checklist

I have...

### Reviewers Checklist

I have...


price-feeder/oracle/client/client.go
```diff
@@ -105,33 +110,10 @@ func (r *passReader) Read(p []byte) (n int, err error) {
...
-}
-
-// BroadcastVote attempts to broadcast a vote transaction within the next voting
-// period.
-//
-// Ref: https://github.com/terra-money/oracle-feeder/blob/baef2a4a02f57a2ffeaa207932b2e03d7fb0fb25/feeder/src/vote.ts#L230
-func (oc OracleClient) BroadcastVote(nextBlockHeight int64, timeoutHeight int64, msgs ...sdk.Msg) error {
-	var height int64
-
+func (oc OracleClient) BroadcastTx(nextBlockHeight int64, timeoutHeight int64, msgs ...sdk.Msg) error {
```
adamewozniak, line 116: Needed to use this for both the voting and pre-voting because in every "voting period" there should be a vote and a pre-vote (in that order), and we can get errors here because of async transactions (e.g. getting back "account sequence mismatch, expected...")

price-feeder/oracle/oracle.go
```diff
@@ -105,11 +105,7 @@ func (o *Oracle) Start(ctx context.Context) error {
 
 			o.lastPriceSyncTS = time.Now()
 
-			// XXX/TODO: We need to evaluate if this is the correct value to wait for
-			// between loop executions.
-			//
-			// Ref: https://github.com/umee-network/umee/issues/256
-			time.Sleep(10 * time.Millisecond)
+			time.Sleep(2250 * time.Millisecond)
```
adamewozniak, line 108: Came to this by looking at terra's timeout and testing it locally - pretty consistently doesn't run into the async broadcast issue, any lower and it does have a few once in a while. Want to keep this on the lower limit because we have retry logic in place & too long could skip voting periods

price-feeder/oracle/client/client.go
```diff
@@ -105,33 +110,10 @@ func (r *passReader) Read(p []byte) (n int, err error) {
 	return n, err
 }
 
-// BroadcastPrevote attempts to broadcast a prevote transaction. Note, submiting
-// a prevote transaction does not require timeout functionality that vote
-// transactions require due to the fact that block timing validation exists on
-// the Umee node.
-//
+// BroadcastTx attempts to broadcast a transaction
+// a few times until it gets timed out
```
alexanderbez, line 114:
```suggestion
// BroadcastTx attempts to broadcast a signed transaction. If it fails, a few re-attempts
// will be made until the transaction succeeds or ultimately times out or fails.
```

price-feeder/oracle/client/client.go
alexanderbez: Shouldn't it be pre-vote and then vote, as voting reveals the salt and commitment to the exchange rates.

price-feeder/oracle/client/client.go
```diff
@@ -161,16 +143,15 @@ func (oc OracleClient) BroadcastVote(nextBlockHeight int64, timeoutHeight int64,
 
 		err = tx.BroadcastTx(ctx, factory, msgs...)
```
alexanderbez, line 144: So this isn't related to the PR, but I looked at this function and it's 99% of what we want. That method is meant to be used by the CLI.

The reason I bring this up is because we want the response object of the tx, so we can for example log the tx hash and block it was committed in.

Let's address this in a separate PR though, i.e. copy the entire contents of this method except the last line where we print it.

price-feeder/oracle/client/client.go
```diff
@@ -161,16 +143,15 @@ func (oc OracleClient) BroadcastVote(nextBlockHeight int64, timeoutHeight int64,
...
-			return err
+			oc.Logger.Err(err).Msg("error broadcasting, retrying...")
+			continue
 		}
 
-		height, err = rpcClient.GetChainHeight(ctx)
-		if err != nil {
-			return err
-		}
+		oc.Logger.Info().Msg("successfully broadcasted")
```
alexanderbez, line 150: we're gonna later include the tx hash and block in this log.

price-feeder/oracle/oracle.go
alexanderbez: ACK. Can we do two things:

1. Move this value into a `const` at the top of the file.
2. Add a comment above the `const` stating the rationale for the value picked. We can use what you just said above.

alexanderbez: Also, can you describe what the objective is? E.g. do we want the loop to trigger/tick X # of times per voting period...something along those lines?

price-feeder/oracle/oracle.go
```diff
@@ -321,7 +317,14 @@ func (o *Oracle) tick() error {
 	}
 
 	if isPrevoteOnlyTx {
-		if err := o.oracleClient.BroadcastPrevote(msg); err != nil {
+		// this timeout could be as small as oracleVotePeriod-indexInVotePeriod,
+		// but we give it some extra time just in case
```
alexanderbez, line 321:
```suggestion
		// This timeout could be as small as oracleVotePeriod-indexInVotePeriod,
		// but we give it some extra time just in case.
```

price-feeder/oracle/oracle.go
```diff
@@ -321,7 +317,14 @@ func (o *Oracle) tick() error {
...
+		// Ref : https://github.com/terra-money/oracle-feeder/blob/baef2a4a02f57a2ffeaa207932b2e03d7fb0fb25/feeder/src/vote.ts#L222
+		if err := o.oracleClient.BroadcastTx(
+			nextBlockHeight,
+			oracleVotePeriod*2,
+			msg,
```
alexanderbez, line 326: Can we call this `preVoteMsg`?

price-feeder/oracle/oracle.go
adamewozniak: Twice per voting period, each with a vote and then a prevote (except for the first tick, which will wait until the next voting period) - will add comments about this and a const

price-feeder/oracle/client/client.go
adamewozniak: Overall, yes, but not in each voting period. In the first voting period there will be only one transaction, a pre-vote, but in subsequent voting periods there will then be a vote, and then a pre-vote for the next voting period

| pre-vote | vote, pre-vote | vote, pre-vote | ... 

I'll clarify this in the comment

price-feeder/oracle/oracle.go
```diff
@@ -25,6 +25,19 @@ import (
...
 
+// Came to this by looking at terra's timeout and
+// testing it locally - pretty consistently doesn't run
+// into the async broadcast issue, any lower and it does
+// have a few once in a while. Want to keep this on the
+// lower limit because we have retry logic in place, and making
+// it too long could skip voting periods. The goal is to have ticker
+// run twice per voting period (one vote for the current period,
+// and one pre-vote for the next), except for the first tick, which
+// will just have a pre-vote.
```
alexanderbez, line 36: We should use 80 or 120 char limit and try to avoid first person 👌

```suggestion
// We define tickerTimeout as the minimum timeout between each oracle loop. We
// define this value empirically based on enough time to collect exchange rates,
// and broadcast pre-vote and vote transactions such that they're committed in a
// block during each voting period.
```

alexanderbez (changes requested):

alexanderbez (approved):

alexanderbez (approved):

PR #277, opened 2021-12-09
feat: x/oracle ballot counting

adamewozniak:
## Description

Adding the ballot counting for the oracle, turning votes -> accepted exchange rates 

----

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/oracle/abci.go
```diff
@@ -0,0 +1,155 @@
...
+	return ((uint64)(ctx.BlockHeight())+1)%blocksPerPeriod == 0
+}
+
+// EndBlocker is called at the end of every block
+func EndBlocker(ctx sdk.Context, k keeper.Keeper) error {
+	defer telemetry.ModuleMeasureSince(types.ModuleName, time.Now(), telemetry.MetricKeyEndBlocker)
+
+	params := k.GetParams(ctx)
+	if IsPeriodLastBlock(ctx, params.VotePeriod) {
+
```
alexanderbez, line 24:
```suggestion
```

x/oracle/abci.go
```diff
@@ -0,0 +1,155 @@
...
+		})
+
+		// Organize votes to ballot by denom
+		// NOTE: **Filter out inactive or jailed validators**
+		// NOTE: **Make abstain votes to have zero vote power**
+		voteMap := k.OrganizeBallotByDenom(ctx, validatorClaimMap)
+
+		// Iterate through ballots and update exchange rates; drop if not enough votes have been achieved.
+		for denom, ballot := range voteMap {
+
```
alexanderbez, line 67:
```suggestion
```

x/oracle/abci.go
```diff
@@ -0,0 +1,155 @@
...
+			return false
+		})
+
+		// Organize votes to ballot by denom
+		// NOTE: **Filter out inactive or jailed validators**
+		// NOTE: **Make abstain votes to have zero vote power**
+		voteMap := k.OrganizeBallotByDenom(ctx, validatorClaimMap)
+
+		// Iterate through ballots and update exchange rates; drop if not enough votes have been achieved.
+		for denom, ballot := range voteMap {
```
alexanderbez, line 66: Is this taken from Terra's codebase? This seems very problematic to me. We're iterating over a map and updating state which is non-deterministic because map iteration order is non-deterministic.

x/oracle/abci.go
```diff
@@ -0,0 +1,155 @@
...
+			exchangeRate, err := Tally(ctx, ballot, params.RewardBand, validatorClaimMap)
+			if err != nil {
+				return err
+			}
+
+			// Set the exchange rate, emit ABCI event
+			k.SetExchangeRateWithEvent(ctx, denom, exchangeRate)
+		}
+
+		//---------------------------
```
alexanderbez, line 78:
```suggestion
```

x/oracle/abci.go
```diff
@@ -0,0 +1,155 @@
...
+		// Do miss counting & slashing
```
alexanderbez, line 79:
```suggestion
		// update miss counting & slashing
```

x/oracle/abci.go
```diff
@@ -0,0 +1,155 @@
...
+		)
+
+		// Clear the ballot
+		k.ClearBallots(ctx, params.VotePeriod)
+
+		// Update vote targets
+		k.ApplyWhitelist(ctx, params.Whitelist, voteTargets)
+	}
+
+	// Do slash who did miss voting over threshold and
```
alexanderbez, line 112:
```suggestion
	// Slash oracle providers who missed voting over the threshold and
```

x/oracle/keeper/keeper.go
```diff
@@ -67,8 +67,8 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 	return ctx.Logger().With("module", fmt.Sprintf("x/%s", types.ModuleName))
 }
 
-// GetExchangeRate gets the consensus exchange rate of USD denominated in the
-// denom asset from the store.
+// GetExchangeRate gets the consensus exchange rate of an asset
```
alexanderbez, line 70: The former comment was more correct. It's the exchange rate against the USD denomination.

x/oracle/types/denom.go
```diff
@@ -14,7 +14,8 @@ func (d Denom) String() string {
 
 // Equal implements equal interface
 func (d Denom) Equal(d1 *Denom) bool {
-	return d.Name == d1.Name
+	return d.BaseDenom == d1.BaseDenom &&
+		d.SymbolDenom == d1.SymbolDenom
```
alexanderbez, line 18: should we compare decimals as well?

x/oracle/types/params.go
```diff
@@ -32,9 +32,12 @@ const (
 var (
 	DefaultVoteThreshold = sdk.NewDecWithPrec(50, 2) // 50%
 	DefaultRewardBand    = sdk.NewDecWithPrec(2, 2)  // 2% (-1, 1)
-	DefaultTobinTax      = sdk.NewDecWithPrec(25, 4) // 0.25%
 	DefaultWhitelist     = DenomList{
-		{Name: UmeeDenom},
+		{
```
alexanderbez, line 36: So a few things:

1. We shouldn't store ATOM in the default list
2. We should only store umee
3. ATOM won't be interpreted as "atom", it'll be the IBC denom, e.g. `ibc/CCFAF...`

alexanderbez (changes requested):
Nice work @adamewozniak

alexanderbez:
Another question I have is, do we normalize the base denominations at all in the exchange rate ballot/tally counting logic? i.e. I want to make sure we're treating `ATOM` and `atom` as the same.

adamewozniak:
> Another question I have is, do we normalize the base denominations at all in the exchange rate ballot/tally counting logic? i.e. I want to make sure we're treating `ATOM` and `atom` as the same.

Not yet, but I'll add that in 👍

x/oracle/abci.go
```diff
@@ -0,0 +1,154 @@
...
+				return err
+			}
+
+			// Set the exchange rate, emit ABCI event
+			k.SetExchangeRateWithEvent(ctx, ballotDenom.Denom, exchangeRate)
+		}
+
+		// update miss counting & slashing
+		voteTargetsLen := len(voteTargets)
+		for _, claim := range validatorClaimMap {
```
alexanderbez, line 80: same comment as before -- we can't iterate over maps and have any state mutation, it'll lead to non-determinism.

x/oracle/abci.go
```diff
@@ -0,0 +1,154 @@
...
+		var voteTargetDenoms []string
+		for _, v := range params.Whitelist {
+			voteTargetDenoms = append(voteTargetDenoms, v.BaseDenom)
+		}
+
+		// Distribute rewards to ballot winners
+		k.RewardBallotWinners(
+			ctx,
+			(int64)(params.VotePeriod),
+			(int64)(params.RewardDistributionWindow),
```
alexanderbez, line 99: just do `int64(variable)` instead.

x/oracle/abci.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/umee-network/umee/x/oracle/keeper"
+	"github.com/umee-network/umee/x/oracle/types"
+
+	"github.com/cosmos/cosmos-sdk/telemetry"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// IsPeriodLastBlock returns true if we are at the last block of the period
+func IsPeriodLastBlock(ctx sdk.Context, blocksPerPeriod uint64) bool {
+	return ((uint64)(ctx.BlockHeight())+1)%blocksPerPeriod == 0
```
alexanderbez, line 15:
```suggestion
	return (uint64(ctx.BlockHeight())+1)%blocksPerPeriod == 0
```

x/oracle/types/asset.go
```diff
@@ -7,6 +7,8 @@ import (
 // Denomination constants
 const (
 	UmeeDenom       string = "uumee"
+	UmeeSymbol      string = "umee"
```
alexanderbez, line 10: we technically don't need these. We store all of them in `app.go`, so we should just reference them here too 👍

x/oracle/types/ballot.go
```diff
@@ -132,6 +133,26 @@ func (pb ExchangeRateBallot) Swap(i, j int) {
...
 
+// BallotDenom is a convenience wrapper for setting rates deterministically.
+type BallotDenom struct {
+	Ballot ExchangeRateBallot
+	Denom  string
+}
+
+// BallotMapToSlice returns an array of sorted exchange rate ballots.
+func BallotMapToSlice(votes map[string]ExchangeRateBallot) (b []BallotDenom) {
+	for denom, ballot := range votes {
```
alexanderbez, line 144: we can make this slightly more efficient by:

```suggestion
func BallotMapToSlice(votes map[string]ExchangeRateBallot) ([]BallotDenom) {
	b := make([]BallotDenom, len(votes))

	for denom, ballot := range votes {
```

x/oracle/types/vote.go
```diff
@@ -91,7 +91,7 @@ func ParseExchangeRateTuples(tuplesStr string) (ExchangeRateTuples, error) {
 		denom := denomAmountStr[0]
 
 		tuples[i] = ExchangeRateTuple{
-			Denom:        denom,
+			Denom:        strings.ToUpper(denom),
```
alexanderbez, line 94: nice!

alexanderbez (changes requested):

x/oracle/types/asset.go
adamewozniak: It's done this way to avoid import cycles - is there a way to get around that?

alexanderbez: ahhh you're totally right. Unfortunately, not. Good point

x/oracle/abci.go
```diff
@@ -0,0 +1,155 @@
...
+			}
+
+			// Increase miss counter
+			k.SetMissCounter(ctx, claim.Recipient, k.GetMissCounter(ctx, claim.Recipient)+1)
+		}
+
+		var voteTargetDenoms []string
+		for _, v := range params.Whitelist {
+			voteTargetDenoms = append(voteTargetDenoms, v.BaseDenom)
+		}
```
alexanderbez, line 94: This is the 2nd loop over `params.WhiteList`.

1. Can we rename this to `AcceptList` (per recent events 😉 )?
2. Loop over the `params.AcceptList` once and collect `voteTargetDenoms` and `voteTargets` at the same time.

proto/umee/oracle/v1beta1/oracle.proto
```diff
@@ -23,8 +23,8 @@ message Params {
     (gogoproto.nullable)   = false
   ];
   uint64         reward_distribution_window = 4 [(gogoproto.moretags) = "yaml:\"reward_distribution_window\""];
-  repeated Denom whitelist                  = 5 [
-    (gogoproto.moretags)     = "yaml:\"whitelist\"",
+  repeated Denom acceptlist                  = 5 [
```
alexanderbez, line 26:
```suggestion
  repeated Denom accept_list                  = 5 [
```

proto/umee/oracle/v1beta1/oracle.proto
```diff
@@ -23,8 +23,8 @@ message Params {
...
+    (gogoproto.moretags)     = "yaml:\"acceptlist\"",
```
alexanderbez, line 27:
```suggestion
    (gogoproto.moretags)     = "yaml:\"accept_list\"",
```

x/oracle/abci.go
```diff
@@ -0,0 +1,152 @@
...
+				validatorClaimMap[valAddr.String()] = types.NewClaim(validator.GetConsensusPower(powerReduction), 0, 0, valAddr)
+				i++
+			}
+		}
+
+		// These are the denoms we want votes on
+		// TODO : Get these from the x/leverage module's accepted list of tokens
+		// Ref https://github.com/umee-network/umee/issues/281
+		var voteTargets []string
+		var voteTargetDenoms []string
```
alexanderbez, line 49:
```suggestion
		var (
		  voteTargets []string
		  voteTargetDenoms []string
		)
```

x/oracle/types/ballot.go
```diff
@@ -132,6 +133,27 @@ func (pb ExchangeRateBallot) Swap(i, j int) {
...
 
+// BallotDenom is a convenience wrapper for setting rates deterministically.
+type BallotDenom struct {
+	Ballot ExchangeRateBallot
+	Denom  string
+}
+
+// BallotMapToSlice returns an array of sorted exchange rate ballots.
+func BallotMapToSlice(votes map[string]ExchangeRateBallot) []BallotDenom {
+	b := make([]BallotDenom, len(votes))
```
alexanderbez, line 144:
```suggestion
	b := make([]BallotDenom, len(votes))
	i := 0
```

x/oracle/types/ballot.go
```diff
@@ -132,6 +133,27 @@ func (pb ExchangeRateBallot) Swap(i, j int) {
...
+	for denom, ballot := range votes {
+		b = append(b, BallotDenom{
```
alexanderbez, line 146: So we have to use a index variable here since we've pre-allocated the slice.
```suggestion
		b[i] = BallotDenom{
```

x/oracle/types/ballot.go
```diff
@@ -149,3 +171,20 @@ func NewClaim(power, weight, winCount int64, recipient sdk.ValAddress) Claim {
 		Recipient: recipient,
 	}
 }
+
+// ClaimMapToSlice returns an array of sorted exchange rate ballots.
+func ClaimMapToSlice(claims map[string]Claim) []Claim {
+	c := make([]Claim, len(claims))
```
alexanderbez, line 177: same -- we must use an index variable here.

x/oracle/keeper/keeper.go
```diff
@@ -80,10 +80,33 @@ func (k Keeper) GetExchangeRate(ctx sdk.Context, denom string) (sdk.Dec, error)
...
+	decProto := sdk.DecProto{}
+	k.cdc.MustUnmarshal(b, &decProto)
 
-	return dp.Dec, nil
+	return decProto.Dec, nil
+}
+
+// GetExchangeRateBase gets the consensus exchange rate of an asset
+// in the base denom (e.g. ATOM -> uatom)
+func (k Keeper) GetExchangeRateBase(ctx sdk.Context, denom string) (sdk.Dec, error) {
```
alexanderbez, line 91: So we gotta be careful with this method. The caller, `x/leverage`, will only have the base denom, e.g. `uumee` or `ibc/XXXX...` (uatom), so the `denom` argument here will be the base denom.

What denom do we use for storing exchange rates via `GetExchangeRateKey`? The base or symbol? If it's the symbol (my guess), then we'll need to have a reverse index from base -> symbol. Then we use that value for the loop check below.

adamewozniak: We're using the symbol - we could use the whitelist to get to the base, or add a map to the store when we do ballot counting

x/oracle/keeper/keeper.go
```diff
@@ -70,6 +70,15 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 // GetExchangeRate gets the consensus exchange rate of USD denominated in the
 // denom asset from the store.
 func (k Keeper) GetExchangeRate(ctx sdk.Context, denom string) (sdk.Dec, error) {
+	// Translate the base denom -> symbol
```
alexanderbez, line 73:
```suggestion
	var symbol string

	// Translate the base denom -> symbol
```

x/oracle/keeper/keeper.go
```diff
@@ -70,6 +70,15 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
...
+	params := k.GetParams(ctx)
+	for i := range params.AcceptList {
+		if params.AcceptList[i].BaseDenom == denom {
+			denom = params.AcceptList[i].SymbolDenom
```
alexanderbez, line 77: let's use another variable instead of overriding input to make it clear.

```suggestion
			symbol = params.AcceptList[i].SymbolDenom
```

x/oracle/keeper/keeper.go
```diff
@@ -70,6 +70,15 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 // GetExchangeRate gets the consensus exchange rate of USD denominated in the
 // denom asset from the store.
 func (k Keeper) GetExchangeRate(ctx sdk.Context, denom string) (sdk.Dec, error) {
```
alexanderbez, line 72: Looks good 👌

x/oracle/keeper/keeper.go
```diff
@@ -80,10 +89,33 @@ func (k Keeper) GetExchangeRate(ctx sdk.Context, denom string) (sdk.Dec, error)
...
+// in the base denom (e.g. ATOM -> uatom)
+func (k Keeper) GetExchangeRateBase(ctx sdk.Context, denom string) (sdk.Dec, error) {
+	store := ctx.KVStore(k.storeKey)
+	b := store.Get(types.GetExchangeRateKey(denom))
+	if b == nil {
+		return sdk.ZeroDec(), sdkerrors.Wrap(types.ErrUnknownDenom, denom)
+	}
+
+	decProto := sdk.DecProto{}
+	k.cdc.MustUnmarshal(b, &decProto)
```
alexanderbez, line 108: This still is technically not correct as the `denom` input is the "base" not the "symbol". However, `GetExchangeRate` already does all the work for us! So we can make this much simpler:

```suggestion
	exchangeRate, err := k.GetExchangeRate(ctx, denom)
	if err != nil {
	    // ...
	}
```

Now we have the exchange rate for the symbol that corresponds to the input `denom`. We can then use this for the remainder of the method 👍

x/oracle/keeper/keeper.go
```diff
@@ -80,10 +89,33 @@ func (k Keeper) GetExchangeRate(ctx sdk.Context, denom string) (sdk.Dec, error)
...
+
+	params := k.GetParams(ctx)
+
+	for _, acceptedDenom := range params.AcceptList {
+		if denom == acceptedDenom.SymbolDenom {
+			return decProto.Dec.QuoInt64(int64(10 ^ acceptedDenom.Exponent)), nil
+		}
```
alexanderbez, line 115:
```suggestion
		if denom == acceptedDenom.BaseDenom {
			return exchangeRate.Dec.QuoInt64(int64(10 ^ acceptedDenom.Exponent)), nil
		}
```

x/oracle/keeper/keeper.go
```diff
@@ -74,16 +74,49 @@ func (k Keeper) GetExchangeRate(ctx sdk.Context, denom string) (sdk.Dec, error)
 		return sdk.OneDec(), nil
 	}
 
+	var symbol string
+
+	// Translate the base denom -> symbol
+	params := k.GetParams(ctx)
+	for i := range params.AcceptList {
```
alexanderbez, line 81: nit, you don't have to use the index, you can just do `for _, item := range params.AcceptList`

alexanderbez (approved):

Issue #329, opened 2021-12-23
x/leverage: ensure safety against empty exchange rates


alexanderbez:
Currently, there is no direct positive incentive for validators to submit exchange rates to the oracle module (AFAIK -- correct me if I'm wrong @adamewozniak).

However, there is a direct negative incentive to not run one, which is validators will get slashed if they do not submit votes via `SlashAndResetMissCounters`. So we should technically never be in a situation where we have missing exchange rates for x/leverage registered assets.

However, I just want to be sure that we're safe against the case where exchange rates may be zero or missing. So any message execution should fail when the exchange rate is zero or missing.

In the case of Begin/EndBlock logic, we might screwed? Can you confirm @toteki ?

toteki:
I'm thinking of putting extra lines of defense on `GetPrice`:

- Recover panics from oracle and turn them into an error (also helps with the bug I'm on)
- Add an error if an exchange rate is ever zero or nil

A second helpful fact: I think `GetPrice` is called during message-based transactions, but never automatically (like on an epoch or during begin/endblock). I'll have to confirm that.

If that's the case, then empty or zero exchange rates will cause transactions which rely on exchange rates to error, but will not halt the chain, and operations can resume normally once exchange rates exist. That will also prevent instant liquidation or other serious problems when an exchange rate is zero.

toteki:
Also, I confirmed that oracle-based functions are not called during begin/end block. The PR above should be adequate.

alexanderbez:
Perfect -- this lines up with what I was expecting in the execution flow.

PR #331, opened 2021-12-24
fix: ensure safety against empty exchange rates

toteki:
## Description

`TokenPrice`, which is the leverage module's single path to `x/oracle`, is not protected against nil/negative/zero exchange rates, and oracle panics. 

It will return errors in those cases.

closes: #329

----

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/x/leverage/keeper/oracle.go
+++ b/x/leverage/keeper/oracle.go
@@ -15,7 +15,6 @@ import (
 // This function will not panic or return non-positive exchange rates, preferring
 // to error instead.
 func (k Keeper) TokenPrice(ctx sdk.Context, denom string) (price sdk.Dec, err error) {
-
 	if !k.IsAcceptedToken(ctx, denom) {
 		return sdk.ZeroDec(), sdkerrors.Wrap(types.ErrInvalidAsset, denom)
 	}
--- a/x/leverage/keeper/oracle.go
+++ b/x/leverage/keeper/oracle.go
@@ -12,23 +12,14 @@ import (
 // the base and display/symbol denominations for each exchange pair. E.g. it must
 // know about the UMEE/USD exchange rate along with the uumee base denomination
 // and the exponent.
-// This function will not panic or return non-positive exchange rates, preferring
+// This function will not return non-positive exchange rates, preferring
 // to error instead.
-func (k Keeper) TokenPrice(ctx sdk.Context, denom string) (price sdk.Dec, err error) {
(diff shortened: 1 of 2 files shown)


x/leverage/keeper/oracle.go
```diff
@@ -12,12 +12,33 @@ import (
 // the base and display/symbol denominations for each exchange pair. E.g. it must
 // know about the UMEE/USD exchange rate along with the uumee base denomination
 // and the exponent.
-func (k Keeper) TokenPrice(ctx sdk.Context, denom string) (sdk.Dec, error) {
+// This function will not panic or return non-positive exchange rates, preferring
+// to error instead.
+func (k Keeper) TokenPrice(ctx sdk.Context, denom string) (price sdk.Dec, err error) {
+
```
alexanderbez, line 18:
```suggestion
```

x/leverage/keeper/oracle.go
```diff
@@ -12,12 +12,33 @@ import (
...
 	if !k.IsAcceptedToken(ctx, denom) {
 		return sdk.ZeroDec(), sdkerrors.Wrap(types.ErrInvalidAsset, denom)
 	}
 
-	return k.oracleKeeper.GetExchangeRateBase(ctx, denom)
+	defer func() {
+		// This will catch panics from the function body
+		// and instead return a general error.
+		if pan := recover(); pan != nil {
```
alexanderbez, line 26: This is not good -- where does the panic come from? I would much rather prefer to not panic in these code path(s) and instead return an error.

toteki: The panic would be coming from the lines in the function body - so the call to the oracle module.

Changing the caught panic into an error here allows us to distinguish between oracle-related and leverage-related panics, as the current state of things gives us something like "panic message hidden to protect potentially sensitive content".

Hopefully oracle never panics at all, but if it does, the error would be recognizable.

alexanderbez: > The panic would be coming from the lines in the function body - so the call to the oracle module.

Then we need to revise where in x/oracle this could panic (as a result) of calling it and have it return an error instead.

toteki: Removed the additional `recover()` logic, so the remaining change is just a check against negative/zero exchange rates. That should cover everything.

alexanderbez (approved):
🎉

PR #366, opened 2022-01-05
fix: borrow limit test behavior

toteki:
## Description

For some reason, updated collateral weight is not taking effect during tests. Needs investigation.

----

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/leverage/client/tests/util.go
```diff
@@ -132,7 +133,68 @@ func (s *IntegrationTestSuite) UpdateRegistry(
...
+	s.Require().NoError(err)
+
+	// Verify that the RegisteredTokens query is returning an updated collateral weight
+	registry := &types.QueryRegisteredTokensResponse{}
+	s.Require().NoError(clientCtx.Codec.UnmarshalJSON(out.Bytes(), registry), out.String())
+	for _, token := range registry.GetRegistry() {
+		if token.BaseDenom == baseDenom {
+			s.Require().Equal(collateralWeight, token.CollateralWeight, "Collateral weight not updated")
+		}
+	}
```
toteki, line 197: @alexanderbez I'm currently blocked at this behavior: After the `UpdateRegistry` utility function is called to update collateral weight, further `QueryRegisteredToken` calls continue to return the old `token.CollateralWeight` instead of the updated one. The current code checks this once but I've also tried a `Require().Eventually` of 2 minutes to wait for the right collateral weight to be returned, to no success. In both cases, the liquidation tests fail in a way consistent with the collateral weight not having changed.

Since the `UpdateRegistry` function itself isn't failing, the proposal is definitely being accepted. Do you know what could be preventing it from being applied to the Token Registry?

alexanderbez: Hey @toteki I'm looking into this right now.

alexanderbez: So I just gave this a shot and it seems to work just fine.

https://pastebin.com/9v252N9Q

Notice I changed the collateral weight. It was 0.05 before and then 0.07 after. I queried before and after and you can see the token registry changing. So I'm not sure what issues you're running into.

toteki: That's odd - the error I'm getting matches the failing `test-unit-cover` output on this PR. I'll look over vs. yours to see if I missed some detail.

So all of `make test-unit` succeeded on your machine when you tested?

alexanderbez: I didn't run any tests. I just made sure `UpdateRegistry` updated the token registry as it is intended.

toteki: Nvm, I think I found the issue. I'll push the solution if it works

alexanderbez: Is the issue that tests are failing or the token registry is not being updated?

alexanderbez: Nice, what was the issue??

toteki: The one I was stuck on seems to be modifying a copy of the registry then passing the original into UpdateRegistry, leading to no changes. There were other fails after that bug so I'm working on them now.

toteki:
Current output of `make test-unit`

```
 --- FAIL: TestIntegrationTestSuite/TestQueryMarketSize (5.11s)
        --- FAIL: TestIntegrationTestSuite/TestQueryMarketSize/valid_asset (0.00s)
            suite.go:114: 
                	Error Trace:	suite.go:114
                	            				suite.go:77
                	Error:      	Received unexpected error:
                	            	rpc error: code = InvalidArgument desc = UMEE: unknown denom: invalid request
                	Test:       	TestIntegrationTestSuite/TestQueryMarketSize/valid_asset

```
but the problem only occurs when the `TestQueryMarketSize` is run after `TestCmdLiquidate`, likely meaning my two registry updates (which change and then reset `uumee`'s collateral factor) are indirectly causing the problem.

Currently in wtf-space as to _how_, but this should be the final bug.

alexanderbez (approved):

x/leverage/client/tests/util.go
```diff
@@ -15,10 +15,13 @@ import (
 	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
 	tmcli "github.com/tendermint/tendermint/libs/cli"
 
+	"github.com/umee-network/umee/app"
 	"github.com/umee-network/umee/x/leverage/client/cli"
 	"github.com/umee-network/umee/x/leverage/types"
 )
 
+var proposalCounter = sdk.ZeroInt()
```
alexanderbez, line 23: we can just a regular int here 👍

x/leverage/client/tests/util.go
```diff
@@ -132,7 +135,53 @@ func (s *IntegrationTestSuite) UpdateRegistry(
...
+	}
+
+	// Increment proposalCounter so we don't re-use proposal title
+	proposalCounter = proposalCounter.Add(sdk.OneInt())
+
+	// Update token registry using the modified token registry - waits for proposal accepted
+	s.UpdateRegistry(
+		clientCtx,
+		types.NewUpdateRegistryProposal(
+			"collateral weight update "+proposalCounter.String(),
```
alexanderbez, line 174:
```suggestion
			fmt.Sprintf("collateral weight update - %d", proposalCounter)
```

x/leverage/client/tests/util.go
alexanderbez, line 23:
```suggestion
var proposalCounter int
```

x/leverage/client/tests/util.go
```diff
@@ -132,7 +135,53 @@ func (s *IntegrationTestSuite) UpdateRegistry(
...
+	// Replace the collateral weight of the selected token with the new value
+	newTokens := resp.GetRegistry()
+	for i := range newTokens {
+		if newTokens[i].BaseDenom == baseDenom {
+			newTokens[i].CollateralWeight = collateralWeight
+		}
+	}
+
+	// Increment proposalCounter so we don't re-use proposal title
+	proposalCounter = proposalCounter.Add(sdk.OneInt())
```
toteki, line 168:
```suggestion
	proposalCounter++
```

alexanderbez (approved):

Issue #276, opened 2021-12-08
x/leverage: Choose default params


toteki:
## Summary

Default params are ready to be chosen in `x/leverage/types/params.go`

## Problem Definition

`DefaultInterestEpoch` The number of blocks (e.g. `100`) between interest calculations.

`DefaultCompleteLiquidationThreshold` How far over borrow limit (e.g. `0.1` = 10% over limit) a borrower must be to be liquidated fully in a single transaction).

`DefaultMinimumCloseFactor` The portion of a borrower's borrowed value (e.g. `0.01` = 1% borrow value) that can be liquidated in a single transaction, when the borrower is just over their borrow limit)

`DefaultOracleRewardFactor` The portion of accrued interest (e.g. `0.01` = 1%) on all borrows that funds the `x/oracle` reward pool.

----

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

toteki:
DefaultOracleRewardFactor added to issue description

alexanderbez:
@toteki I would just your discretion here to set default values and then close this issue.

The values we decide to use in genesis for mainnet we'll discuss internally 👍

PR #415, opened 2022-01-22
chore: tweak default interest epoch

toteki:
## Description

Changes default interest epoch to 1000 blocks, which is 1-2 hours when block time is 3-5 seconds.

The other default params should be fine where they are - closes: #276

----

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/x/leverage/spec/07_params.md
+++ b/x/leverage/spec/07_params.md
@@ -4,7 +4,7 @@ The leverage module contains the following parameters:
 
 | Key                          | Type    | Example |
 | -----------------------------| ------- | ------- |
-| InterestEpoch                | int64   | 100     |
+| InterestEpoch                | int64   | 1000    |
 | CompleteLiquidationThreshold | sdk.Dec | 0.1     |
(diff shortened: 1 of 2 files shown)


alexanderbez:
What's the rationale for such a value? By the way, are APYs calculated and set every interest epoch or every block? If it's the former, wouldn't that mean our rates would be fairly inelastic?

toteki:
APYs are calculated on epoch, so there is the risk of not updating for an hour or so in response to a massive shift in borrow utilization.

I suspect as market size increases, swings in utilization will become slower and less severe, and the amount of calculations required for each epoch increase. In reality we could start with very fast epochs and transition to slower ones over time.

In any case the amount of interest accrual contained within an hour's APY is negligible, so the question is mainly one of responsiveness vs. computation.

Do you think we should aim for the ~1hour default epoch or keep ~6minutes?

toteki:
Closing - insufficient reason to change it looks like

PR #470, opened 2022-02-01
chore: double cpu cores in packer build to reduce failure

dreamcodez:
## Description

Double the cpu cores available to packer for building to reduce failure rate (was 1 core, now is 2); this will also provide more memory and network to the gcp instance during image building.

## Justification

1. The docker image build (running on github actions runners) does not ever fail -- well, the github runner has 2 cpus.
2. The packer (gcp) image builder defaults to using a 1 core machine to do the work, and I believe this is why it is so flaky right now, it is resource starved.


dreamcodez:
First build failure is transient/ignorable failure (known issue). (This is ok, i'm trying to fix the flakiness in the build, not the image overwrite issue)

adamewozniak (approved):

PR #475, opened 2022-02-01
fix: wait for cloud init during packer gcp build to avoid apt race condition

dreamcodez:
## Description
There is a race condition between cloud-init and the packer code which is causing the intermittent apt failures (hypothesis).

## Research on this issue:

https://askubuntu.com/questions/663296/intermittent-apt-get-install-issues-when-creating-an-ami-using-packer

https://github.com/hashicorp/packer/issues/2639

suggested fix:
```
provisioners:
  - type: shell
    inline:
    - /usr/bin/cloud-init status --wait
```

### DIFF
--- a/infra/packer/umee-node-dist.pkr.hcl
+++ b/infra/packer/umee-node-dist.pkr.hcl
@@ -34,10 +34,16 @@ source "googlecompute" "final" {
 
 build {
   name = "umee-node"
+
   sources = [
     "source.docker.final",
     "source.googlecompute.final"
   ]
+
+  provisioner "shell" {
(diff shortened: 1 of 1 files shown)


dreamcodez:
error:

```
==> umee-node.googlecompute.final: Error waiting for image: googleapi: Error 409: The resource 'projects/umeedefi/global/images/umee-node-7c59ec26dab59c5acddc10446386c7bd7b36e3a9' already exists, alreadyExists
```

will tackle idempotency in a separate effort

adamewozniak (approved):

Issue #492, opened 2022-02-03
x/oracle integer overflow


adamewozniak:
## Summary of Bug

Validators may send large exchange rates, crashing other nodes via a panic due an integer overflow when performing ballot arithemetic via sdk.Dec types..

```golang
 for _, v := range pb { 
 	deviation := v.ExchangeRate.Sub(median) 
 	sum = sum.Add(deviation.Mul(deviation)) 
 } 
```

While there are overflow bound checks that exist on individual exchange rate votes, there aren't overflow checks on the computation performed above.

Found during the ToB audit 
Was fixed in an [issue](https://github.com/terra-money/core/issues/667) opened up on Terra Core

----

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

toteki:
Terra fixed temporarily by doing 

```
	// The deferred recover here is what they added
	defer func() {
		if e := recover(); e != nil {
			standardDeviation = sdk.ZeroDec()
		}
	}()

	sum := sdk.ZeroDec()
	for _, v := range pb {
		deviation := v.ExchangeRate.Sub(median)
		sum = sum.Add(deviation.Mul(deviation))
	}
	variance := sum.QuoInt64(int64(len(pb)))
	floatNum, _ := strconv.ParseFloat(variance.String(), 64)
	floatNum = math.Sqrt(floatNum)
	standardDeviation, _ = sdk.NewDecFromStr(fmt.Sprintf("%f", floatNum))
	return
```

The error only existed because they were trying to parse `sdk.Dec`, which cannot overflow, into `float64`, which can overflow.

But since we have already removed such parsing by keeping things as `sdk.Dec`
```
	sum := sdk.ZeroDec()
	for _, v := range pb {
		deviation := v.ExchangeRate.Sub(median)
		sum = sum.Add(deviation.Mul(deviation))
	}

	variance := sum.QuoInt64(int64(len(pb)))

	standardDeviation, err := variance.ApproxSqrt()

	if err != nil {
		return sdk.ZeroDec(), err
	}

	return standardDeviation, nil
```
I don't think any changes are required - we're already immune.

Does that look right?

adamewozniak:
> Does that look right?

I'm pretty sure this is an issue, this test panics : 

```golang
func TestPBStandardDeviationOverflow(t *testing.T) {
[...]

PR #493, opened 2022-02-03
fix: x/oracle integer overflow

adamewozniak:
## Description

Adds a check for sdk.Dec overflowing when calculating standard deviation in `ExchangeRateBallot.StandardDeviation`, and a test to prove it. 

closes: #492

----

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/oracle/types/ballot.go
```diff
@@ -86,6 +86,11 @@ func (pb ExchangeRateBallot) StandardDeviation() (sdk.Dec, error) {
 	sum := sdk.ZeroDec()
 	for _, v := range pb {
 		deviation := v.ExchangeRate.Sub(median)
+		// check for overflow
+		err := checkMulOverflow(deviation, deviation)
+		if err != nil {
+			continue
```
adamewozniak, line 92: Pretty sure this is the more-correct thing to do - in terra's implementation the standard deviation becomes zero : 

```golang
	defer func() {
		if e := recover(); e != nil {
			standardDeviation = sdk.ZeroDec()
		}
	}()
```

Which forces the ballot rewarding logic to use `RewardBand` instead of `𝜎` ([spec here](https://github.com/umee-network/umee/blob/main/x/oracle/spec/01_concepts.md#reward-band))

This seems maybe unintended or at least confusing, so I think we should just throw it out, use the others to calculate `𝜎`, and then let the `𝜀 = max(𝜎, R/2)` logic run regularly

x/oracle/types/ballot.go
```diff
@@ -84,12 +84,19 @@ func (pb ExchangeRateBallot) StandardDeviation() (sdk.Dec, error) {
...
+		err := checkMulOverflow(deviation, deviation)
+		if err != nil {
+			overflowCount++
+			continue
+		}
 		sum = sum.Add(deviation.Mul(deviation))
 	}
 
-	variance := sum.QuoInt64(int64(len(pb)))
+	variance := sum.QuoInt64(int64(len(pb) - overflowCount))
```
adamewozniak, line 99: Not a big fan of `overflowCount` being a separate variable, but this seemed like the cleanest way to account for skipped rates

x/oracle/types/ballot_test.go
```diff
@@ -247,3 +247,34 @@ func TestPBStandardDeviation(t *testing.T) {
 		require.Equal(t, tc.standardDeviation, stdDev)
 	}
 }
+
+func TestPBStandardDeviation_Overflow(t *testing.T) {
```
adamewozniak, line 251: Keeping this out of `TestPBStandardDeviation` since that test set uses `float64` -> `sdk.Dec`, which we should probably refactor in a separate PR

x/oracle/types/ballot.go
alexanderbez: Ehhh yeah I'm not sure how I feel about having to port over decimal logic from the SDK just to check for overflow bounds here. Maybe the panic recovery is the cleaner option along with your suggestion above?

toteki: If we can use panic-recover for code cleanliness, but have the logic throw out the offending vote and keep the rest, that seems like it would minimize an attacker's power to manipulate reward behavior (like a single validator being able to stop `𝜎` calculation and force use of `RewardBand`).

x/oracle/types/ballot.go
adamewozniak: Never mind, realized a cleaner way

alexanderbez (approved):
LGTM!

alexanderbez:
@toteki wanna give this a quick look?

x/oracle/types/ballot_test.go
```diff
@@ -247,3 +247,34 @@ func TestPBStandardDeviation(t *testing.T) {
...
+			overflowRate,
+			UmeeSymbol,
+			valAddr,
+			1,
+		),
+	}
+
+	deviation, err := pb.StandardDeviation()
+	require.NoError(t, err)
+	expectedDevation, _ := sdk.NewDecFromStr("871.862661203013097586")
```
toteki, line 278:
```suggestion
	expectedDevation := sdk.MustNewDecFromStr("871.862661203013097586")
```

toteki (approved):

PR #499, opened 2022-02-04
test: leverage module test coverage

toteki:
## Description

This is a subset of the test improvements I'm writing.

- New tests in `borrows_test.go`, `collateral_test.go`
- Some tests `x/leverage/keeper/keeper_test.go` moved to new test files
- Type `TestKeeper` in `internal_test.go` which can be used to call un-exported methods on leverage keeper during tests, allowing for coverage of some hard-to-reach error branches.
- Multiple method names of leverage keeper changed to lowercase, so that they are not exposed anymore outside of testing.

----

### Author Checklist

I have...

### Reviewers Checklist

I have...


toteki:
To avoid clutter, I'm marking this PR as R4R at the current size.

It contains some new borrow tests, the leverage `TestKeeper `, some extra safeguards against invalid input on any function containing `store.Set`, and miscellaneous housekeeping.

Next PR or two will be more of the same: leverage tests and housekeeping.

x/leverage/keeper/store.go
```diff
@@ -38,13 +38,16 @@ func (k Keeper) getAdjustedTotalBorrowed(ctx sdk.Context, denom string) sdk.Dec
...
+// of computing it from real borrowed amount. Should only be used by genesis and setBorrow, as other
+// functions deal in non-adjusted amounts using setBorrow. Also updates AdjustedTotalBorrowed by the
 // resulting changes in borrowed amount. If either amount to store is zero, any stored value is cleared.
 func (k Keeper) setAdjustedBorrow(ctx sdk.Context, addr sdk.AccAddress, borrow sdk.DecCoin) error {
 	if err := borrow.Validate(); err != nil {
 		return err
 	}
+	if addr.Empty() {
+		return types.ErrEmptyAddress
+	}
```
toteki, line 50: Empty addresses are especially dangerous when interacting with store keys of the structure `prefix | lengthPrefixed(addr) | denom | 0x00` because the function `address.MustLengthPrefix(addr)` will return `[]byte{}` on empty input.

That would cause our getters and iterators to interpret the first byte of `denom` as the length prefix to an address, likely reaching out of slice bounds for the address, or interpreting part of a long IBC denom as an address, and the remainder as its own unique denom, causing all kinds of unpredictable behavior.

Since `addr.Empty` is computationally cheap I put it in all `x/leverage` functions with `store.Set` that take address as an input, even though such bad input should already be stopped before reaching such code.

x/leverage/keeper/internal_test.go
```diff
@@ -0,0 +1,63 @@
+package keeper
```
alexanderbez, line 1: maybe use `keeper_test` here instead?

toteki: This has to be `package keeper` so it can access unexported methods. (`keeper_test` would have to import `keeper`)

But since the filename ends in `_test.go` it doesn't build outside of tests, and it's not accessible to packages other than `keeper_test` (even `x/leverage/simulation` can't use `NewTestKeeper`).

This seemed to be the way to do it, to allow for testing of unexported methods using the test suite.

alexanderbez: ACK 👍

alexanderbez (approved):

Issue #542, opened 2022-02-11
price-feeder: change to use last traded price volume-weighted


RafilxTenfen:
## Summary

- Today in the providers we use the last traded price, the proposal is for us to use a volume-weighted price for the trades that occur in the last N minutes   

## Problem Definition

- We could use this feature to have a more concise price using all the traded made by that ticker in the last N minutes 

## Proposal

- The idea of this implementation comes from [terra oracle](https://github.com/terra-money/oracle-feeder)
   - They maintain all the trades that happen in the last 60 min (don't know why soo much)
   - Filter for the ones that happened last 3 minutes 
   - If it has one it uses `tvwap(time volume-weighted average price)`
   - in the end, it is `sum(volume * price) / (total volume)`

- For this to happen we need to: 
   - Change the Provider structs to store all the trades that happened in the last N minutes 
   - Compute the price taking into consideration all the trades that happened in all the providers in the last N minutes
   

----

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

adamewozniak:
Right direction, although maybe we don't store the last 60, maybe just the last 10? This should probably be implemented after we switch to websockets, seems like it'd be easier to do

adamewozniak:
note, the last bits we need here are the resolution to [this comment](https://github.com/umee-network/umee/pull/592#discussion_r818084703) and the addition of faulty price feeder detection for tvwap. 

cc : @RafilxTenfen

adamewozniak:
Closing this since #609 & #610 have been merged in - will address problems with warning messaging in #619

PR #586, opened 2022-02-25
feat: price-feeder: candlestick websocket subscriptions

adamewozniak:
## Description

Adds subscriptions to the candlestick websockets for our providers. Necessary for implementing tvwap.
Doing calculations in a separate PR to keep this tidy. 

progress on: #542

----

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/price-feeder/oracle/provider/binance.go
+++ b/price-feeder/oracle/provider/binance.go
@@ -89,6 +89,7 @@ func NewBinanceProvider(ctx context.Context, logger zerolog.Logger, pairs ...typ
 		wsClient:        wsConn,
 		logger:          logger.With().Str("module", "oracle").Logger(),
 		tickers:         map[string]BinanceTicker{},
+		candles:         map[string]BinanceCandle{},
 		subscribedPairs: pairs,
 	}
 
--- a/price-feeder/oracle/provider/binance.go
+++ b/price-feeder/oracle/provider/binance.go
@@ -190,7 +190,7 @@ func (p *BinanceProvider) subscribeTickers(cps ...types.CurrencyPair) error {
 	return p.wsClient.WriteJSON(subsMsg)
 }
 
-// subscribeTickers subscribe to all candles
+// subscribeCandles subscribe to all candles
 func (p *BinanceProvider) subscribeCandles(cps ...types.CurrencyPair) error {
 	params := make([]string, len(cps))
 
--- a/price-feeder/oracle/provider/binance.go
+++ b/price-feeder/oracle/provider/binance.go
@@ -15,9 +15,8 @@ import (
 )
 
 const (
-	binanceHost           = "stream.binance.com:9443"
-	binancePath           = "/ws/umeestream"
-	binanceConnectionTime = time.Hour * 23 //  should be < 24
+	binanceHost = "stream.binance.com:9443"
+	binancePath = "/ws/umeestream"
 )
 
 var _ Provider = (*BinanceProvider)(nil)
--- a/price-feeder/oracle/provider/binance.go
+++ b/price-feeder/oracle/provider/binance.go
@@ -148,10 +148,10 @@ func (p *BinanceProvider) messageReceived(messageType int, bz []byte) {
 
 	// sometimes the message received is not a ticker or a candle response.
 	if err := json.Unmarshal(bz, &tickerResp); err != nil {
-		p.logger.Err(err).Msg("could not unmarshal ticker response")
+		p.logger.Debug().Err(err).Msg("could not unmarshal ticker response")
 	}
 	if err := json.Unmarshal(bz, &candleResp); err != nil {
-		p.logger.Err(err).Msg("could not unmarshal candle response")
+		p.logger.Debug().Err(err).Msg("could not unmarshal candle response")
 	}
 
 	if len(tickerResp.LastPrice) != 0 {
--- a/price-feeder/oracle/provider/binance.go
+++ b/price-feeder/oracle/provider/binance.go
@@ -51,9 +51,9 @@ type (
 	}
 
 	BinanceCandleMetadata struct {
-		CloseTime  int64  `json:"T"` // Close time in unix epoch ex.: 1645756200000
-		ClosePrice string `json:"c"` // Price at close
-		Volume     string `json:"v"` // Volume during period
+		Close     string `json:"c"` // Price at close
+		TimeStamp int64  `json:"T"` // Close time in unix epoch ex.: 1645756200000
+		Volume    string `json:"v"` // Volume during period
 	}
 
 	BinanceCandle struct {
@@ -157,7 +157,7 @@ func (p *BinanceProvider) messageReceived(messageType int, bz []byte) {
 	if len(tickerResp.LastPrice) != 0 {
 		p.setTickerPair(tickerResp)
 	}
-	if len(candleResp.Metadata.ClosePrice) != 0 {
+	if len(candleResp.Metadata.Close) != 0 {
 		p.setCandlePair(candleResp)
 	}
 }
--- a/price-feeder/oracle/provider/binance.go
+++ b/price-feeder/oracle/provider/binance.go
@@ -15,8 +15,9 @@ import (
 )
 
 const (
-	binanceHost = "stream.binance.com:9443"
-	binancePath = "/ws/umeestream"
+	binanceHost           = "stream.binance.com:9443"
+	binancePath           = "/ws/umeestream"
+	binanceConnectionTime = time.Hour * 23 //  should be < 24
 )
 
 var _ Provider = (*BinanceProvider)(nil)
@@ -26,12 +27,14 @@ type (
 	// API.
 	//
 	// REF: https://binance-docs.github.io/apidocs/spot/en/#individual-symbol-mini-ticker-stream
+	// REF: https://binance-docs.github.io/apidocs/spot/en/#kline-candlestick-streams
 	BinanceProvider struct {
 		wsURL           url.URL
 		wsClient        *websocket.Conn
 		logger          zerolog.Logger
 		mu              sync.Mutex
 		tickers         map[string]BinanceTicker // Symbol => BinanceTicker
+		candles         map[string]BinanceCandle // Symbol => BinanceCandle
 		subscribedPairs []types.CurrencyPair
 	}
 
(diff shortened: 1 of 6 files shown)


price-feeder/oracle/provider/huobi.go
```diff
@@ -31,19 +31,21 @@ type (
...
 		tickers         map[string]HuobiTicker // market.$symbol.ticker => HuobiTicker
+		candles         map[string]HuobiCandle // market.$symbol.kline.$period => HuobiCandle
 		subscribedPairs []types.CurrencyPair
 	}
 
 	// HuobiTicker defines the response type for the channel and
 	// the tick object for a given ticker/symbol.
 	HuobiTicker struct {
-		CH   string    `json:"ch"` // Data belonged channel，Format：market.$symbol.ticker
+		CH   string    `json:"ch"` // Channel name.$symbol.ticker
```
toteki, line 48: [Docs](https://huobiapi.github.io/docs/spot/v1/en/#market-ticker) seem to suggest `market.$symbol.ticker`
Is that correct?

price-feeder/oracle/provider/huobi.go
```diff
@@ -157,17 +190,27 @@ func (p *HuobiProvider) messageReceived(messageType int, bz []byte, reconnectTic
...
+
+	if err := json.Unmarshal(bz, &candleResp); err != nil {
+		p.logger.Err(err).Msg("failed to unmarshal message")
 		return
 	}
 
-	p.setTickerPair(tickerResp)
+	if candleResp.Tick.Open != 0 {
+		p.setCandlePair(candleResp)
+	}
```
toteki, line 213: Does anything need to be done if the response is neither ticker nor candle? Currently it would return without error but also without setting anything it looks like

price-feeder/oracle/provider/kraken.go
```diff
@@ -197,42 +230,40 @@ func (p *KrakenProvider) messageReceived(messageType int, bz []byte) {
...
-		return
+		return fmt.Errorf("sent an unexpected structure")
 	}
 
-	channelName, ok := tickerMessage[2].(string)
+	channelName, ok := tickerMessage[1].(string)
 	if !ok || channelName != "ticker" {
-		p.logger.Debug().Msg("sent an unexpected channel name")
-		return
+		return fmt.Errorf("sent an unexpected channel name")
```
toteki, line 248: Possibly "sent" -> "received" on these error messages?

price-feeder/oracle/provider/okx.go
```diff
@@ -158,9 +186,21 @@ func (p *OkxProvider) messageReceived(messageType int, bz []byte) {
...
 
 	var tickerResp OkxTickerResponse
+	var candleResp OkxCandleResponse
+
+	// sometimes it returns other messages which are not
+	// tickerResponses or candleResponses
 	if err := json.Unmarshal(bz, &tickerResp); err != nil {
-		// sometimes it returns other messages which are not tickerResponses
-		p.logger.Err(err).Msg("could not unmarshal")
+		if err := json.Unmarshal(bz, &candleResp); err != nil {
```
toteki, line 194: Does the placement of this `json.Unmarshal(bz, &candleResp)` inside of the error case cause the candle unmarshal to be missed for candle responses?

If I recall correctly, unmarshaling json of struct A into struct B will ignore the fields of A that are not present in B, without returning an error.

[sauce](https://cs.opensource.google/go/go/+/refs/tags/go1.17.7:src/encoding/json/decode.go)
```go
// To unmarshal JSON into a struct, Unmarshal matches incoming object
// keys to the keys used by Marshal (either the struct field name or its tag),
// preferring an exact match but also accepting a case-insensitive match. By
// default, object keys which don't have a corresponding struct field are
// ignored (see Decoder.DisallowUnknownFields for an alternative).
```

price-feeder/oracle/provider/huobi.go
adamewozniak: it is `market.$symbol.ticker`, fixed that bit

price-feeder/oracle/provider/huobi.go
adamewozniak: I don't think so, the original implementation just would return here without an error if there were no ticker to set. 

cc : @RafilxTenfen

price-feeder/oracle/provider/okx.go
adamewozniak: Nope, the error you get from the first attempt is due to a type disagreement : 
```
error(*encoding/json.UnmarshalTypeError) *{Value: "array", Type: reflect.Type(*reflect.rtype) *{size: 48, ptrdata: 40, hash: 1242516269, tflag: tflagUncommon|tflagExtraStar|tflagNamed (7), align: 8, fieldAlign: 8, kind: 25, equal: type..eq.github.com/umee-network/umee/price-feeder/oracle/provider.OkxTickerPair, gcdata: *21, str: 247223, ptrToThis: 1015424}, Offset: 60, Struct: "OkxTickerResponse", Field: "data"}
```
This isn't a super clear way to do it, maybe it'd be better to do : 
```golang
	// sometimes it returns other messages which are not
	// tickerResponses or candleResponses
	if err := json.Unmarshal(bz, &tickerResp); err != nil {
		p.logger.Debug().Err(err).Msg("could not unmarshal ticker resp")
	}
	if err := json.Unmarshal(bz, &candleResp); err != nil {
		p.logger.Debug().Err(err).Msg("could not unmarshal candle resp")
	}

	for _, tickerPair := range tickerResp.Data {
		p.setTickerPair(tickerPair)
	}
	for _, candlePair := range candleResp.Data {
		p.setCandlePair(candlePair, candleResp.ID.InstID)
	}
```

a bit less efficient, but way more readable

price-feeder/oracle/provider/huobi.go
```diff
@@ -157,17 +190,27 @@ func (p *HuobiProvider) messageReceived(messageType int, bz []byte, reconnectTic
 	}
 
 	var tickerResp HuobiTicker
+	var candleResp HuobiCandle
```
alexanderbez, line 193: nit: group the var :)

price-feeder/oracle/provider/huobi.go
```diff
@@ -54,6 +56,21 @@ type (
...
+	// HuobiCandle defines the response type for the channel and
+	// the tick object for a given ticker/symbol.
+	HuobiCandle struct {
+		CH   string          `json:"ch"` // Channel name. Format：market.$symbol.kline.$period
+		Tick HuobiCandleTick `json:"tick"`
+	}
+
+	// HuobiCandleTick defines the response type for the candle.
+	HuobiCandleTick struct {
+		Open      float64 `json:"open"`   // Opening price during this period
```
alexanderbez, line 68: Just checking that we only define fields that we actually use

alexanderbez (changes requested):
Nice work @adamewozniak

price-feeder/oracle/provider/binance.go
```diff
@@ -48,6 +50,19 @@ type (
 		C         uint64 `json:"C"` // Statistics close time
 	}
 
+	BinanceCandleMetadata struct {
+		StartTime  int64  `json:"t"` // Start time in unix epoch ex.: 1645756200000
```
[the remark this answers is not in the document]
adamewozniak: Took out all except close price, volume, and timestamp

price-feeder/oracle/provider/okx.go
```diff
@@ -157,15 +186,28 @@ func (p *OkxProvider) messageReceived(messageType int, bz []byte) {
...
-		p.setTickerPair(tickerPair)
+	if tickerResp.ID.Channel == "tickers" {
+		for _, tickerPair := range tickerResp.Data {
+			p.setTickerPair(tickerPair)
+		}
+	}
+	if candleResp.ID.Channel == "candle1m" {
+		for _, candlePair := range candleResp.Data {
+			p.setCandlePair(candlePair, candleResp.ID.InstID)
+		}
```
adamewozniak, line 210: This bit of logic is necessary for okx since if the response is a ticker response, it will unmarshal into the `candleResp.Data` as a string array where [0] is an empty string array.

We could handle it by checking lengths, but this seemed more straightforward

price-feeder/oracle/provider/okx.go
```diff
@@ -175,6 +217,22 @@ func (p *OkxProvider) setTickerPair(tickerPair OkxTickerPair) {
 	p.tickers[tickerPair.InstId] = tickerPair
 }
 
+func (p *OkxProvider) setCandlePair(pairData []string, instID string) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	ts, err := strconv.ParseInt(pairData[0], 10, 64)
+	if err != nil {
+		ts = 0
```
alexanderbez, line 226: do we still want to have this set if we can't parse the time?

price-feeder/oracle/provider/provider.go
```diff
@@ -17,13 +17,12 @@ const (
...
 
-var (
-	ping = []byte("ping")
-)
+var ping = []byte("ping")
 
 // Provider defines an interface an exchange price provider must implement.
 type Provider interface {
 	GetTickerPrices(...types.CurrencyPair) (map[string]TickerPrice, error)
+	// GetVolumes(...types.CurrencyPair) (map[string]TickerVolume, error)
```
alexanderbez, line 25: Let's remove this until/if it's ever needed.

price-feeder/oracle/provider/okx.go
adamewozniak: returned here instead 👍

price-feeder/oracle/provider/kraken.go
```diff
@@ -180,59 +208,64 @@ func (p *KrakenProvider) messageReceived(messageType int, bz []byte) {
...
-		return
-	case krakenEventSubscriptionStatus:
-		p.messageReceivedSubscriptionStatus(bz)
+	if err := p.messageReceivedTickerPrice(bz); err != nil {
+		// msg is not a ticker, it will try to marshal to candle message
+		p.logger.Debug().Err(err).Msg("unable to unmarshal ticker")
+	} else {
 		return
 	}
+	if err := p.messageReceivedCandle(bz); err != nil {
```
alexanderbez, line 230: This will be executed if `messageReceivedTickerPrice` errors correct?

alexanderbez (approved):

price-feeder/oracle/provider/kraken.go
adamewozniak: Yup, logic is if `messageReceivedTickerPrice` errors, log and continue to attempt parsing as a candl; else, return and don't attempt to parse as a candle

price-feeder/oracle/provider/huobi.go
RafilxTenfen: it depends on the provider to send different messages, on the huobi one I only saw messages sent that we subscribed for

PR #754, opened 2022-04-05
chore: v2 module replace

toteki:
Replaces umee `go.mod` path and imported path to include v2.

Compare to [osmosis v7 go.mod](https://github.com/osmosis-labs/osmosis/blob/main/go.mod) and [import paths](https://github.com/osmosis-labs/osmosis/blob/main/x/gamm/keeper/keeper.go)

Ref: [How to upgrade to a major version in Go](https://golang.cafe/blog/how-to-upgrade-to-a-major-version-in-go.html)


toteki:
Note: `price-feeder` import paths still need to be figured out.

toteki:
Also note: doing the following in proto files causes unexpected behavior in make proto-gen. Looking into it.
```
option go_package = "github.com/umee-network/umee/v2/x/leverage/types";
```

contrib/scripts/protocgen.sh
```diff
@@ -30,5 +30,5 @@ for dir in $proto_dirs; do
     $(find "${dir}" -maxdepth 1 -name '*.proto')
 done
 
-cp -r ./github.com/umee-network/umee/* ./
+cp -r ./github.com/umee-network/umee/v*/x/* x/
```
toteki, line 33: Moves `make proto-gen` output, such as `./v2/x/leverage/types/genesis.pb.go`, to the intended output folder without `v2`.

The previous line was doing something similar for non-v2 output.

toteki:
> How many files would you like to change?

Yes.

RafilxTenfen (approved):

PR #853, opened 2022-04-28
chore: add code to build terra for base infra image

dreamcodez:


adamewozniak (approved):
lgtm, just adding the terrad binary to the image 👍 

also thanks for the PF / hermes updates ❤️

Issue #869, opened 2022-05-05
x/leverage: Lack of token validation on setRegisteredToken ( RV #18 )


toteki:
# Runtime Verification Audit Item 18

Issue: line 52 in keeper/token.go, in function SetRegisteredToken(ctx, token), no validation was performed on the token. Furthermore, there is no validation in InitGenesis() and handleUpdateRegistryProposalHandler() where setRegisteredToken() are called.

Effect: some of the token parameters may not meet expectation, for example, BaseBorrowRate is non-negative and KinkUtilizationRate is positive and less than 1.0.

Recommendation: call Validate() function in types/token.go before writing the token into the registry.

Issue #724, opened 2022-03-30
x/leverage: accepted -> unaccepted assets behavior


toteki:
Suppose that a user has deposited `1000 uabcd` and received `800 u/uabcd` (a token:uToken exchange rate of `1.25`). They may or may not have enabled `uabcd` as collateral.

Other users may have borrowed `uabcd` as well.

Then, the leverage module's token registry is updated, and for whatever reason, `uabcd` has been removed as an accepted asset.

EDIT: Oracle prices will still be available for symbol `ABCD`

## Causes

An asset could be removed from the accepted asset list in two scenarios:

- Governance decision (token registry update made with intent to remove asset)
- Accident (token registry update missing an entry)

## Consequences

| How should this scenario be treated? | What changed to the code are required? |
| - | - |
| User should be able to withdraw `uabcd` that has been lent or set as collateral if no borrow limit concerns overrule. | `WithdrawAsset` and `ExchangeUToken` need to function for assets which are not in the token registry, but whose uToken exchange rates still exist in store, as would be the case. |
| Users should be able to repay, but not borrow, `uabcd` | Alter `RepayAsset` |
| User should be able to disable, but not enable, `uabcd` as collateral if no borrow limit concers overrule.| Rearrange `SetCollateralSetting` |
| Calculations involving `BorrowLimit`, `LiquidationLimit`, or `BorrowedValue` should omit unregistered tokens from the total USD value, or even fail altogether. | TBD below |
| Without oracle prices, liquidations of `uabcd` borrows or collateral are not possible. Liquidation of the borrower's other denoms of debt and collateral may be possible. | TBD below |
| `EndBlocker` had better not crash | No changes - interest iterates over accepted assets only. |

Inevitably, anyone who was using `uabcd` as collateral may instantly drop below their borrow or liquidation limits. This would also happen if an asset's collateral weight were reduced or zeroed by governance, an action which would occur before intentional (but not accidental) versions of the scenario.

Anyone borrowing `uabcd` could be off the hook if `uabcd` no longer contributes USD value to their borrowed value. Alternately, users could be prevented from borrow-limit-dependent actions (withdraw, borrow, unset collateral) until all their borrows containing unaccepted asset `uabcd` have been repaid. This is cleaner, as long as `uabcd` can still be obtained.

## Decisions

| Question | Initial Conclusion |
| - | - |
[...]

PR #913, opened 2022-05-15
feat: add enabled/blacklist fields to tokens

toteki:
## Description

Adds various options for disabling borrow, lend, and oracle price calculations of asset types.

Prevents token registry update from EVER removing base denoms from token registry. Therefore:
- Gov updates only need to include the tokens they're actually adding/changing.
- Removing tokens intentionally can be done by disabling borrowing/lending, or setting the `Blacklist` field depending on the situation. Their metadata and params remain in the token registry.

Also adds validation to `SetRegisteredToken`

closes: #869
closes: #724 

----

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/leverage/gov_handler.go
```diff
@@ -22,12 +22,10 @@ func NewUpdateRegistryProposalHandler(k keeper.Keeper) govtypes.Handler {
...
-	if err := k.DeleteRegisteredTokens(ctx); err != nil {
-		return err
-	}
-
 	for _, token := range p.Registry {
-		k.SetRegisteredToken(ctx, token)
+		if err := k.SetRegisteredToken(ctx, token); err != nil {
+			return err
+		}
 	}
```
toteki, line 29: Note the functional difference in update token registry - now any existing tokens will remain or be updated, instead of being deleted. It is no longer possible to delete registered tokens entirely.

x/leverage/keeper/oracle.go
```diff
@@ -14,11 +14,16 @@ import (
 // the base and display/symbol denominations for each exchange pair. E.g. it must
 // know about the UMEE/USD exchange rate along with the uumee base denomination
 // and the exponent.
-// This function will not return non-positive exchange rates, preferring
-// to error instead.
+// This function will not return non-positive exchange rates unless a token is
+// blacklisted, in which case it will return zero.
```
toteki, line 18:
```suggestion
// This function will only return positive exchange rates or errors, unless a
// token is blacklisted, in which case it will return zero.
```

x/oracle/keeper/hooks_test.go
```diff
@@ -9,6 +9,12 @@ func (s *IntegrationTestSuite) TestHooks_AfterTokenRegistered() {
...
 	s.Require().Len(s.app.OracleKeeper.AcceptList(s.ctx), 1)
 
+	// TODO: These hooks need to respond to the Token.Blacklist field
+	// (as well as update symbol denom and exponent if they ever change).
+	// Blacklisting in particular should actually eliminate the base
+	// denom from the oracle (and un-blacklisting, for whatever reason,
+	// should do the opposite.)
+
 	// require that an existing token does not change the accept list
 	h.AfterTokenRegistered(s.ctx, leveragetypes.Token{
```
toteki, line 19: @adamewozniak Could you add the necessary oracle changes to this PR after reviewing?

x/leverage/gov_handler_test.go
```diff
@@ -30,30 +30,93 @@ func TestUpdateRegistryProposalHandler(t *testing.T) {
...
 
+		// ensure that unaffected token was not deleted by the gov proposal
 		token, err := k.GetRegisteredToken(ctx, "uatom")
 		require.NoError(t, err)
+
+		// ensure that the existing token was updated by the gov proposal
+		token, err = k.GetRegisteredToken(ctx, "uosmo")
+		require.NoError(t, err)
 		require.Equal(t, "0.020000000000000000", token.BaseBorrowRate.String())
 	})
```
robert-zaremba, line 121: let's add a test checking that a token with `enableLend=true, blacklisted=true` will fail to process - according to proto, when blacklisted=true then enable* must be false.

x/oracle/keeper/hooks_test.go
adamewozniak: @toteki I don't believe this update is necessary; the oracle module should be relatively agnostic of the x/leverage module's functionality. 

For example, if we used the oracle as a service for other blockchains over IBC, we would not want a leverage "blacklist" to remove those from the oracle's params.

In the case where we **definitely do** want to remove a denom from the oracle, a parameter governance proposal is still possible.

adamewozniak: @toteki I don't believe this update is necessary; the oracle module should be relatively agnostic of the x/leverage module's functionality. 

For example, if we used the oracle as a service for other blockchains over IBC, we would not want a leverage "blacklist" to remove those from the oracle's params.

In the case where we **definitely do** want to remove a denom from the oracle, a parameter governance proposal is still possible.

x/leverage/keeper/keeper_test.go
```diff
@@ -618,8 +643,13 @@ func (s *IntegrationTestSuite) TestLiqudateBorrow_Valid() {
 		MaxBorrowRate:        sdk.MustNewDecFromStr("1.0"),
 		KinkUtilizationRate:  sdk.MustNewDecFromStr("0.8"),
 		LiquidationIncentive: sdk.MustNewDecFromStr("0.1"),
+		SymbolDenom:          "UMEE",
+		Exponent:             6,
+		EnableLend:           true,
+		EnableBorrow:         true,
+		Blacklist:            false,
 	}
```
robert-zaremba, line 651: let's try to reduce copy pasted definitions and define them once in a suite.

x/leverage/gov_handler_test.go
```diff
@@ -30,30 +30,93 @@ func TestUpdateRegistryProposalHandler(t *testing.T) {
...
 
 		tokens := k.GetAllRegisteredTokens(ctx)
-		require.Len(t, tokens, 2)
+		require.Len(t, tokens, 3)
 
-		_, err := k.GetRegisteredToken(ctx, "uosmo")
-		require.Error(t, err)
-
-		_, err = k.GetRegisteredToken(ctx, "uumee")
+		// ensure that new tokens was added
```
adamewozniak, line 109:
```suggestion
```

x/leverage/gov_handler_test.go
```diff
@@ -30,30 +30,93 @@ func TestUpdateRegistryProposalHandler(t *testing.T) {
...
+		_, err := k.GetRegisteredToken(ctx, "uumee")
 		require.NoError(t, err)
 
+		// ensure that unaffected token was not deleted by the gov proposal
```
adamewozniak, line 113:
```suggestion
		// ensure that only the specified token was deleted
```

x/leverage/gov_handler_test.go
```diff
@@ -30,30 +30,93 @@ func TestUpdateRegistryProposalHandler(t *testing.T) {
...
 		token, err := k.GetRegisteredToken(ctx, "uatom")
 		require.NoError(t, err)
+
+		// ensure that the existing token was updated by the gov proposal
```
adamewozniak, line 117:
```suggestion
```

x/leverage/keeper/keeper.go
```diff
@@ -67,8 +67,8 @@ func (k Keeper) ModuleBalance(ctx sdk.Context, denom string) sdk.Int {
 // exchange for uTokens. If asset type is invalid or account balance is
 // insufficient, we return an error.
 func (k Keeper) LendAsset(ctx sdk.Context, lenderAddr sdk.AccAddress, loan sdk.Coin) error {
-	if !k.IsAcceptedToken(ctx, loan.Denom) {
-		return sdkerrors.Wrap(types.ErrInvalidAsset, loan.String())
+	if err := k.RequireLendEnabled(ctx, loan.Denom); err != nil {
```
adamewozniak, line 70: I think using the boolean logic is a bit cleaner than having a function return an error if false: 
```suggestion
	if !k.IsLendingEnabled(ctx, loan.Denom) {...
```

x/leverage/keeper/keeper.go
```diff
@@ -220,8 +220,8 @@ func (k Keeper) BorrowAsset(ctx sdk.Context, borrowerAddr sdk.AccAddress, borrow
 		return sdkerrors.Wrap(types.ErrInvalidAsset, borrow.String())
 	}
 
-	if !k.IsAcceptedToken(ctx, borrow.Denom) {
-		return sdkerrors.Wrap(types.ErrInvalidAsset, borrow.String())
+	if err := k.RequireBorrowEnabled(ctx, borrow.Denom); err != nil {
```
adamewozniak, line 223: Same here:
```suggestion
	if !k.IsBorrowEnabled(ctx, borrow.Denom) {...
```

x/leverage/keeper/keeper.go
```diff
@@ -375,9 +371,6 @@ func (k Keeper) SetCollateralSetting(ctx sdk.Context, borrowerAddr sdk.AccAddres
 
 // GetCollateralSetting checks if a uToken denom is enabled for use as collateral by a single borrower.
 func (k Keeper) GetCollateralSetting(ctx sdk.Context, borrowerAddr sdk.AccAddress, denom string) bool {
-	if !k.IsAcceptedUToken(ctx, denom) {
-		return false
-	}
```
adamewozniak, line 380: Why was this removed?

x/leverage/keeper/keeper.go
```diff
@@ -400,6 +393,9 @@ func (k Keeper) LiquidateBorrow(
 	if !desiredRepayment.IsValid() {
 		return sdk.ZeroInt(), sdk.ZeroInt(), sdkerrors.Wrap(types.ErrInvalidAsset, desiredRepayment.String())
 	}
+	if err := k.RequireNotBlacklisted(ctx, desiredRepayment.Denom); err != nil {
+		return sdk.ZeroInt(), sdk.ZeroInt(), err
```
adamewozniak, line 397:
```suggestion
	if !k.IsBlacklisted(ctx, desiredRepayment.Denom) {
		return sdk.ZeroInt(), sdk.ZeroInt(), fmt.Errorf...
```

x/oracle/keeper/hooks_test.go
toteki: Ok - in that case, we can remove the TODO for removing oracle accepted assets as a result of leverage module blacklisting.

Would it be ok to make a separate umee repo issue to cover "update of SymbolDenom / Exponent" situation that was mentioned?

adamewozniak (commented):
left a few suggestions 🥳

x/oracle/keeper/hooks_test.go
```diff
@@ -9,6 +9,12 @@ func (s *IntegrationTestSuite) TestHooks_AfterTokenRegistered() {
 	h := s.app.OracleKeeper.Hooks()
 	s.Require().Len(s.app.OracleKeeper.AcceptList(s.ctx), 1)
 
+	// TODO: These hooks need to respond to the Token.Blacklist field
+	// (as well as update symbol denom and exponent if they ever change).
+	// Blacklisting in particular should actually eliminate the base
+	// denom from the oracle (and un-blacklisting, for whatever reason,
+	// should do the opposite.)
+
```
toteki, line 17:
```suggestion
```

x/leverage/keeper/keeper.go
toteki: I started with that function signature actually, but got persuaded by two points:
- I was immediately turning around to make an error "lending not enabled" on seeing the boolean
- It was no longer possible to distinguish "token does not exit" with "lending disabled" on boolean return

x/oracle/keeper/hooks_test.go
adamewozniak: Imo that's a similar scenario, they could just be updated via param change proposals

x/leverage/keeper/keeper.go
toteki: Redundant check mostly - a collateral setting would only be true if an asset was acceptable when it was enabled.

In the cases where the asset is no longer accepted (specifically lending disabled or blacklisted now) retuning true if a 0x01 was stored in state doesn't seem harmful.

x/leverage/gov_handler_test.go
toteki: no tokens were deleted - I'll clarify comment in a different way

x/leverage/keeper/keeper.go
adamewozniak: Gotcha, makes sense with multiple codepaths - I'd suggest renaming this function to something like `[Check/Validate]LendingEnabled` rather than use the `Required` syntax then 😄

x/leverage/gov_handler_test.go
toteki: Part of that test exists in types/token_test.go - it ensures that token.Validate fails on that condition.

Also just added a test in this file that ensures the gov proposal fails on invalid token.

x/leverage/gov_handler_test.go
```diff
@@ -30,30 +30,93 @@ func TestUpdateRegistryProposalHandler(t *testing.T) {
...
+					LiquidationThreshold: sdk.MustNewDecFromStr("0.25"),
+					BaseBorrowRate:       sdk.MustNewDecFromStr("0.02"),
+					KinkBorrowRate:       sdk.MustNewDecFromStr("0.22"),
+					MaxBorrowRate:        sdk.MustNewDecFromStr("1.52"),
+					KinkUtilizationRate:  sdk.MustNewDecFromStr("0.8"),
+					LiquidationIncentive: sdk.MustNewDecFromStr("0.1"),
+					EnableLend:           true,
+					EnableBorrow:         true,
+					Blacklist:            false,
+				},
```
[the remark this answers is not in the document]
toteki: added to `gov_handler_test.go` - might do a more robust one in `keeper_test.go` next

proto/umee/leverage/v1/leverage.proto
```diff
@@ -117,4 +117,22 @@ message Token {
   // list of allowed tokens.
   string symbol_denom = 10 [(gogoproto.moretags) = "yaml:\"symbol_denom\""];
   uint32 exponent     = 11 [(gogoproto.moretags) = "yaml:\"exponent\""];
+  
+  // Allows lending of this token. Note that withdrawing is always enabled.
+  // Disabling lending would be one step in phasing out an asset type.
+  bool enable_lend = 12;
```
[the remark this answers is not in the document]
toteki: See comment on enable_borrow below - if staking derivatives were registered with lending enabled and borrowing disabled, then we would have created a working `meToken`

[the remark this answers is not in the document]
robert-zaremba: With staking derivatives, you would stake `sATOM` and lend `sATOM`.... unless we do something more sophisticated.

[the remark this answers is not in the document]
toteki: My impression was that we would not allow meTokens to be borrowed - but they can be used as collateral. That would be one implementation enabled by keeping those booleans separate.

CHANGELOG.md
```diff
@@ -46,6 +46,11 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ## [Unreleased]
 
+### State Machine Breaking
+
+- [913](https://github.com/umee-network/umee/pull/913) Add LendEnabled, BorrowEnabled, and Blacklist to Token struct.
```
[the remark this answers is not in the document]
toteki: Done

proto/umee/leverage/v1/leverage.proto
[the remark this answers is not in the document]
robert-zaremba: but here you have `lend` and `borrow`. lend != collateral.
IMO, if an asset is eligible for lending then it's eligible for borrowing.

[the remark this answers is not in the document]
toteki: okay, so mainly a naming issue. Either way, this one needs to allow `MsgLend` and `MsgSetCollateral`, and the other needs to allow `MsgBorrow`

x/leverage/keeper/keeper.go
toteki: _Note: Resolved the other two conversations to group them here - whatever conclusion we reach here will apply to RequireBorrowEnabled and RequireNotBlacklisted_

proto/umee/leverage/v1/leverage.proto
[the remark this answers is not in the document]
robert-zaremba: So let's rename. This must be clear for users and devs. I see 2 flags:
1. enabled for lending & borrowing
2. enabled as a collateral
`disabled` is implicit = it requires that the 2 flag above are set to false.

proto/umee/leverage/v1/leverage.proto
```diff
@@ -117,4 +117,22 @@ message Token {
...
+  // only, like meTokens.
+  bool enable_borrow = 13;
+
+  // This should only be used to eliminate an asset completely. A blacklisted
+  // asset is treated as though its oracle price is zero, and thus ignored by
+  // calculations such as collateral value and borrow limit. Can still be repaid
+  // or withdrawn, but not liquidated. A blacklisted token must have enable_lend
+  // and enable_borrow set to false. Such tokens can be safely removed from the
+  // by the oracle and price feeder as well.
+  bool blacklist = 14;
```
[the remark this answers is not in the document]
toteki: Runs the risk of being interpreted as just a combination of the other two being disabled.

This is more serious in that it also forces the value of the coin to zero in all calculations, which would have all kinds of consequences - we might look for a name that communicates that, but `blacklist` at least shows the severity.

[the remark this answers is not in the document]
robert-zaremba: In https://github.com/umee-network/umee/pull/913#discussion_r875948663 I was thinking to remove this field. 

What it would take to unblackist an asset?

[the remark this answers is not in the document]
toteki: Technically, only a governance proposal with the `blacklisted` field set back to false.

In reality, it seems unlikely that an accepted asset could become so nasty we have to blacklist it, but then recover to a point where we need to reverse the decision.

proto/umee/leverage/v1/leverage.proto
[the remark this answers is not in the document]
toteki: Would changing `(EnableLend, EnableBorrow)` to `(EnableSupply, EnableBorrow)` be better?

One param allows Lend & Collateral actions, and the other allows borrowing.

And separately from this, an asset can also have collateral weight equal to zero (that covers the "borrow and lend, but not collateral" case.)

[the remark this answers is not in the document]
robert-zaremba: I don't understand why you want to have `EnableSupply = Lend & Collateral`?
I think it should be:
* EnableCollateral = asset enabled as a collateral for borrowing
* EnableLending = asset enabled for borrowing and lending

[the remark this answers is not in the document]
toteki: Only `uTokens` can be collateral, and the only way to get a `uToken` is by `MsgLend`

x/leverage/keeper/keeper.go
toteki: Renamed all to assert - resolving conversation.

proto/umee/leverage/v1/leverage.proto
```diff
@@ -117,4 +117,22 @@ message Token {
   // list of allowed tokens.
   string symbol_denom = 10 [(gogoproto.moretags) = "yaml:\"symbol_denom\""];
   uint32 exponent     = 11 [(gogoproto.moretags) = "yaml:\"exponent\""];
+  
+  // Allows lending of this token. Note that withdrawing is always enabled.
+  // Disabling lending would be one step in phasing out an asset type.
```
robert-zaremba, line 122:
```suggestion
  // Allows lending and setting a collateral using this token. Note that withdrawing 
  // is always enabled. Disabling lending would be one step in phasing out an asset type.
```

proto/umee/leverage/v1/leverage.proto
```diff
@@ -117,4 +117,22 @@ message Token {
...
+  bool enable_lend = 12;
+
+  // Allows borrowing of this token. Note that repaying is always enabled.
+  // Disabling borrowing would be one step in phasing out an asset type, but
+  // could also be used from the start for asset types meant to be collateral
+  // only, like meTokens.
+  bool enable_borrow = 13;
```
robert-zaremba, line 129:
```suggestion
  bool enable_msg_borrow = 13;
```

proto/umee/leverage/v1/leverage.proto
robert-zaremba, line 123:
```suggestion
  // Allows lending of this token. Note that withdrawing is always enabled.
  // Disabling lending would be one step in phasing out an asset type.
  bool enable_msg_lend = 12;
```

robert-zaremba (approved):
LGTM. Pending proto comments update

proto/umee/leverage/v1/leverage.proto
```diff
@@ -117,4 +117,22 @@ message Token {
   // list of allowed tokens.
   string symbol_denom = 10 [(gogoproto.moretags) = "yaml:\"symbol_denom\""];
   uint32 exponent     = 11 [(gogoproto.moretags) = "yaml:\"exponent\""];
+  
+  // Allows lending and setting a collateral using this token. Note that withdrawing 
+  // is always enabled. Disabling lending would be one step in phasing out an asset type.
+  bool enable_lend = 12;
```
toteki, line 123:
```suggestion
  bool enable_msg_lend = 12;
```

PR #928, opened 2022-05-19
fix: facus' testnet halt gb patch

dreamcodez:
This is what the testnet needed to unhalt... in progress is the upstream contribution of this bugfix.


toteki (changes requested):
This needs to be done cleaner - it nukes 159 files

dreamcodez:
> This needs to be done cleaner - it nukes 159 files

its wrong.

PR #946, opened 2022-05-25
chore: Remove image publishing from main umee repo (moving)

dreamcodez:
Given how much the complexity of this has grown -- I don't think it really belongs in this repo anymore -- I need to refactor this in any case for our next testnet and I think its better if i move it to the infra repo.

----

### Reviewers Checklist

I have...


toteki:
Is the failing image publish CI on this PR a fluke or something that needs fixing?

adamewozniak:
Alternatively could we just remove terra from the docker build altogether? Afaik we don't need it for testnet anymore

dreamcodez:
> Alternatively could we just remove terra from the docker build altogether? Afaik we don't need it for testnet anymore

this failure has nothing to do with terra.. some sort of permissions issue strange because nothing changed. could be a fluke. retrying build....

dreamcodez:
> Is the failing image publish CI on this PR a fluke or something that needs fixing?

more info, will troubleshoot more tomorrow:

```
umee-node.docker.final (docker-push): denied: Permission "artifactregistry.repositories.uploadArtifacts" denied on resource "projects/umeedefi/locations/us/repositories/stack" (or it may not exist)
```

dreamcodez:
Given how much the complexity of this has grown -- I don't think it really belongs in this repo anymore -- I need to refactor this in any case for our next testnet and I think its better if i move it to the infra repo.

toteki:
I'm in favor of the move to infra - would like second opinion from other humans too though

adamewozniak (approved):
Agree - this stuff could be useful in another public repo, but this is not where it belongs

PR #984, opened 2022-05-28
build(deps): copy price feeder folder from main to 2xx, then go mod tidy

toteki:
## Description

This is a draft PR to see what it would take to sync the 2.x.x branch's price feeder folder to main branch. I want to see if the CI passes.

Main reason I'm looking into this is that the version of the price feeder in `release/2.x.x` still imports umee `v0.8.1`, and I'm not sure what the consequences of that are.


toteki:
Closing: forgot about main branch oracle changes that make price feeder here not compile

Issue #543, opened 2022-02-13
Beta-enabled chain reaches consensus failure on binary rebuild and restart


adamewozniak:
## Summary of Bug

When running the beta version of the chain locally, rebuilding the binary in beta mode, stopping the chain, and then restarting, it reaches consensus failure via an error from the `x/leverage` module : 

`ERR CONSENSUS FAILURE!!! err="-0.000002092846270928 years: negative time elapsed since last interest time" module=consensus ...`

Not sure if this is a major issue, but still worth jotting down 

## Version

Please provide the output of the following commands:

- `$ umeed version`
price-feeder/v0.1.0-c66f8961
- `$ go version`
go version go1.17.3 darwin/arm64
- `$ uname -a`
Darwin Adams-MacBook-Pro-2.local 21.3.0 Darwin Kernel Version 21.3.0: Wed Jan  5 21:37:58 PST 2022; root:xnu-8019.80.24~20/RELEASE_ARM64_T6000 arm64

## Steps to Reproduce

1. Launch beta umee network
```
UMEE_ENABLE_BETA=true starport chain serve -c ./starport.ci.beta.yml -v --reset-once
```
2. Rebuild binary with beta support
```
UMEE_ENABLE_BETA=true make install
```
3. Stop & Restart umee network 
```
UMEE_ENABLE_BETA=true starport chain serve -c ./starport.ci.beta.yml -v
```

----

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

alexanderbez:
Interesting. @toteki does anything stick out here?

toteki:
We were talking about it. and the only way for this to happen is for the `LastInterestTime` in state to be less than `ctx.BlockTime().Unix()` - it comes down to the behavior of `BlockTime` and `EndBlocker` (during which `AccrueAllInterest` sets `LastInterestTime` to `ctx.BlockTime().Unix()` after accruing interest.)

The error message posted, by the way, comes out to be a 3 minute time reversal, so much longer than a block time, and might be consistent with the time the chain was stopped before relaunch. Still can't figure out why `BlockTime` would reverse though.

RafilxTenfen:
@adamewozniak This still happen? I couldn't reproduce

toteki:
[...]

PR #1018, opened 2022-06-14
fix: Return nil if negative time elapsed from last block happens

RafilxTenfen:
## Description

Implemented Option 3: Modify AccrueAllInterest to treat "negative time elapsed" as "zero time elapsed", aborting the function without modifying the state until an EndBlock where BlockTime > LastInterestTime.

closes: #543 

----

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 
 ![Logo!](assets/umee-logo.png)
 
-[![Project Status: WIP – Initial development is in progress, but there has not yet been a stable, usable release suitable for the public.](https://img.shields.io/badge/repo%20status-WIP-yellow.svg?style=flat-square)](https://www.repostatus.org/#wip)
+[![Project Status: Active – The project has reached a stable, usable state and is being actively developed.](https://www.repostatus.org/badges/latest/active.svg)](https://www.repostatus.org/#wip)
 [![GoDoc](https://img.shields.io/badge/godoc-reference-blue?style=flat-square&logo=go)](https://godoc.org/github.com/umee-network/umee)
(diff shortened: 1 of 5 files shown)


x/leverage/keeper/interest.go
```diff
@@ -73,7 +73,13 @@ func (k Keeper) AccrueAllInterest(ctx sdk.Context) error {
...
 	// calculate time elapsed since last interest accrual (measured in years for APR math)
 	if currentTime < prevInterestTime {
-		return types.ErrNegativeTimeElapsed.Wrap(fmt.Sprintf("current: %d, prev: %d", currentTime, prevInterestTime))
+		k.Logger(ctx).With("AccrueAllInterest will wait for block time > prevInterestTime").Error(
+			types.ErrNegativeTimeElapsed.Error(),
+			"current", currentTime,
+			"prev", prevInterestTime,
+		)
+
+		return nil
```
robert-zaremba, line 82: Log is not enough. I think we should either return some message or keep the error.

RafilxTenfen: What do you think about [these options](https://github.com/umee-network/umee/issues/543#issuecomment-1152706616)?

toteki (approved):

RafilxTenfen:
@robert-zaremba So, do you think we should change from `Error` to `Warn`?

toteki:
Message can keep `Error` log level for now. Once tendermint fixes, we can return error instead of `nil`, which would bring it back to causing chain halt (which we want in this case).

robert-zaremba:
OK

x/leverage/keeper/interest.go
robert-zaremba: OK

robert-zaremba (approved):

Issue #827, opened 2022-04-21
x/leverage: liquidation_targets query is inefficient


toteki:
Need validators to be able to disable this specific query, or to make it more efficient, to avoid creating a vulnerability.

Should be done or mitigated before mainnet.

PR #1028, opened 2022-06-16
fix!: remove liquidation targets query

toteki:
## Description

Removes a query which was capable of seriously slowing down nodes on testnet (i.e. freeze 40sec per usage)

closes: #827 

----

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/price-feeder/CHANGELOG.md
+++ b/price-feeder/CHANGELOG.md
@@ -47,6 +47,7 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ### Improvements
 
+- [#1032](https://github.com/umee-network/umee/pull/1032) Update the accepted tvwap period from 3 minutes to 5 minutes.
 - [#978](https://github.com/umee-network/umee/pull/978) Cleanup the oracle package by moving deviation & conversion logic.
 
 ### Features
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -46,6 +46,9 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ## [Unreleased]
 
+### API Breaking
+TODO
```
toteki, line 50:
```suggestion

- [1028](https://github.com/umee-network/umee/pull/1028) Remove liquidation targets query
```

toteki:
Might need someone to run `make-proto-gen` on their machine for this one. My computer is failing to produce the required changes (removing liquidation targets query from `.pb.gw.go`) when I do it locally.

toteki:
> Looks like the `.pb.gw.go` file has been updated - is there more to be removed?
> 
> Also, is this query still necessary for liquidation targeting? @toteki

I did the file manually to compile so there might be diffs if gen were run properly.

We will need some other way to do this query on the liquidators side, but without exposing it on real nodes because of the huge burden it creates. Maybe reintroduce later with a flag or enable?

toteki:
Maybe we can enable the query only when a runtime flag is set, so nobody has it by accident.

In that's the case, then this PR with the proto removal can be closed, and a new one with the flag opened.

adamewozniak:
@toteki can we paginate this with an upper limit instead?

That way all clients could expect the same type of queries to work for all nodes, and some of the burden is reduced

Issue #1031, opened 2022-06-16
Price-feeder: Allow for config to override hard-coded API endpoints


adamewozniak:
## Summary

In case the endpoints of these APIs become deprecated or moved, validators should be able to swap them out without having to download a new binary. So, we should add the option for validators to override these in the `config.toml`

----

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1038, opened 2022-06-20
feat!: make the price feeder's uris configurable

adamewozniak:
## Description

Adds a feature which would allow validators to customize which websocket / rest URIs that they'd like to use. Very useful in case one of these is removed or renamed 😄 

closes: #1031

----

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/price-feeder/oracle/provider/coinbase.go
+++ b/price-feeder/oracle/provider/coinbase.go
@@ -99,9 +99,9 @@ func NewCoinbaseProvider(
 	endpoints config.ProviderEndpoint,
 	pairs ...types.CurrencyPair,
 ) (*CoinbaseProvider, error) {
-	if endpoints.Name != "coinbase" {
+	if endpoints.Name != config.ProviderCoinbase {
 		endpoints = config.ProviderEndpoint{
-			Name:      "coinbase",
+			Name:      config.ProviderCoinbase,
 			Rest:      coinbaseRestHost,
 			Websocket: coinbaseWSHost,
 		}
--- a/price-feeder/oracle/provider/coinbase.go
+++ b/price-feeder/oracle/provider/coinbase.go
@@ -21,12 +21,12 @@ import (
 )
 
 const (
-	coinbaseWSHost           = "ws-feed.exchange.coinbase.com"
-	coinbasePingCheck        = time.Second * 28 // should be < 30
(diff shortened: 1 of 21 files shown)


adamewozniak:
Docs for this to be added to the readme as a part of https://github.com/umee-network/umee/issues/1039 😄

adamewozniak:
Would like a tested ACK from @RafilxTenfen before this gets merged in pls 😉

RafilxTenfen (approved):
LGTM, 

I would ask you to add an example of `provider_endpoints` in the `price-feeder/price-feeder.example.toml`

and giving an incorrect endpoint to any provider would be an error of the guy who wanted to use a different endpoint, right? 

Because at the moment we do not raise any warning saying it is an invalid endpoint (or is different than the default), like this: 
image

price-feeder/oracle/provider/coinbase.go
```diff
@@ -94,11 +96,19 @@ type (
...
+	endpoints config.ProviderEndpoint,
 	pairs ...types.CurrencyPair,
 ) (*CoinbaseProvider, error) {
+	if endpoints.Name != config.ProviderCoinbase {
+		endpoints = config.ProviderEndpoint{
+			Name:      config.ProviderCoinbase,
+			Rest:      coinbaseRestHost,
+			Websocket: coinbaseWSHost,
+		}
+	}
```
toteki, line 108: This condition could throw an error instead if we wanted to avoid silently defaulting when passing the wrong (as opposed to just empty) endpoint.

toteki (approved):
Looks good. Only requested change involves clarifying behavior when `endpoints.Name != intended`

Example scenario:
```go
{
  Name:      "GTAE", // typo
  Rest:      "custom1",
  Websocket: "custom2",
}
```
would currently result (without an error/log) in the use of
```go
{
  Name:      config.ProviderGate,
  Rest:      gateRestHost,
  Websocket: gateWSHost,
}
```
so the debugging experience would be
> Why aren't my custom settings taking? It keeps using defaults

instead of the more direct
> I got an error saying "GTAE != GATE"

toteki:
That should work. Also, do you think in regular usage that `endpoints.Name == ""` would be common? If not, we can still return error instead of overwriting `endpoints` where I suggested.

PR #1062, opened 2022-06-23
chore: API and token registry renames

robert-zaremba:
## Description

+ API cleanup
+ removing unnecessary functions (some functions were used only once - and that one time was in a test)
+ renaming

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/leverage/keeper/genesis.go
```diff
@@ -79,17 +79,28 @@ func (k Keeper) InitGenesis(ctx sdk.Context, genState types.GenesisState) {
...
+	panicOnErr := func(err error) {
+		if err != nil {
+			panic(err)
+		}
+	}
+	tokens, err := k.GetAllRegisteredTokens(ctx)
+	panicOnErr(err)
+	reserves, err := k.GetAllReserves(ctx)
+	panicOnErr(err)
+	usuppy, err := k.GetAllUTokenSupply(ctx)
```
toteki, line 91:
```suggestion
	usupplies, err := k.GetAllUTokenSupply(ctx)
```

x/leverage/keeper/genesis.go
```diff
@@ -79,17 +79,28 @@ func (k Keeper) InitGenesis(ctx sdk.Context, genState types.GenesisState) {
...
 		k.getAllAdjustedBorrows(ctx),
 		k.getAllCollateralSettings(ctx),
 		k.getAllCollateral(ctx),
-		k.GetAllReserves(ctx),
+		reserves,
 		k.GetLastInterestTime(ctx),
 		k.getAllBadDebts(ctx),
 		k.getAllInterestScalars(ctx),
-		k.GetAllUTokenSupply(ctx),
+		usuppy,
```
toteki, line 103:
```suggestion
		usupplies,
```

toteki:
Propose `GetTokenSetting` -> `GetToken` rename. Thoughts?

x/leverage/keeper/keeper.go
```diff
@@ -63,11 +63,22 @@ func (k Keeper) ModuleBalance(ctx sdk.Context, denom string) sdk.Int {
...
+func (k Keeper) validateLendAsset(ctx sdk.Context, loan sdk.Coin) error {
+	if !loan.IsValid() {
+		return types.ErrInvalidAsset.Wrap(loan.String())
+	}
+	token, err := k.GetTokenSettings(ctx, loan.Denom)
+	if err != nil {
+		return err
+	}
+	return token.AssertLendEnabled()
+}
```
toteki, line 75: This is good - I might look for other places to do something similar

x/leverage/gov_handler_test.go
```diff
@@ -81,10 +81,10 @@ func TestUpdateRegistryProposalHandler(t *testing.T) {
 		tokens := k.GetAllRegisteredTokens(ctx)
 		require.Len(t, tokens, 3)
```
toteki, line 82:
```suggestion
		tokens, err := k.GetAllRegisteredTokens(ctx)
		require.NoError(t, err)
		require.Len(t, tokens, 3)
```

x/leverage/gov_handler_test.go
```diff
@@ -81,10 +81,10 @@ func TestUpdateRegistryProposalHandler(t *testing.T) {
...
 
-		_, err := k.GetRegisteredToken(ctx, "uumee")
+		_, err := k.GetTokenSettings(ctx, "uumee")
```
toteki, line 84:
```suggestion
		_, err = k.GetTokenSettings(ctx, "uumee")
```

toteki (approved):
Reverted some changes (notably: `error` returns on store `Get` functions, where the only possible error is unmarshal, have changed back to `panic`)

Aside from that, should be good to merge.

robert-zaremba:
thanks for updating the tests @toteki . But why did you revert back to panics? We should really avoid panics, even if it makes code more verbose. Having panics in API is a bad style, unfortunately we have too many of them in the Cosmos SDK.

toteki:
Store+Getter+Unmarshal failures in particular should be extremely rare - hopefully never occur in production - that's why the code was initially written with panics in those places.

I'm following the sdk, where simple functions like (bank) GetBalance don't have an error return, even though they _technically_ could run into badly formatted data.

robert-zaremba:
> Store+Getter+Unmarshal failures in particular should be extremely rare - hopefully never occur in production - that's why the code was initially written with panics in those places.

This is not a reason to use panics. Panic should be only used when there is no way to recover (ie, error won't make any sense).

> I'm following the sdk, where simple functions like (bank) GetBalance don't have an error return

It has been discussed various times, and we consider it a technical debt.

robert-zaremba:
In general, there is no good argument for using panic in simple functions and errors in non simple functions.

PR #1069, opened 2022-06-24
feat: add cache for chain height for 5s 

RafilxTenfen:
## Description
- Helps to decrease the number of calls from price-feeder to node chain

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/price-feeder/oracle/client/chain_height.go
+++ b/price-feeder/oracle/client/chain_height.go
@@ -1,0 +1,79 @@
+package client
+
+import (
+	"sync"
+	"time"
+
+	"github.com/cosmos/cosmos-sdk/client"
+	rpcclient "github.com/cosmos/cosmos-sdk/client/rpc"
+)
+
+const (
+	// nodeStatsCacheInterval represents the amount of seconds
+	// to update the node stats info from.
+	nodeStatsCacheInterval = time.Second * 5
+)
+
+// ChainHeight is used to cache the chain height of the
+// current node which keeps getting updated at the amount
+// of interval nodeStatsCacheInterval.
+type ChainHeight struct {
+	clientCtx         client.Context
+	mtx               sync.RWMutex
+	errGetChainHeight error
+	lastChainHeight   int64
+}
+
+var (
+	// keep the instance thread safe.
+	once sync.Once
+
+	// keeps the single instance of ChainHeight.
+	instance *ChainHeight
+)
+
+// ChainHeightInstance returns the single instance of ChainHeight.
+func ChainHeightInstance(clientCtx client.Context) *ChainHeight {
+	once.Do(func() {
+		instance = &ChainHeight{
+			clientCtx:         clientCtx,
+			errGetChainHeight: nil,
+			lastChainHeight:   0,
+		}
+		go instance.KeepUpdating()
+	})
+
+	return instance
+}
+
+// Update retrieves the most recent oracle params and
+// updates the instance.
+func (chainHeight *ChainHeight) UpdateChainHeight() {
+	chainHeight.mtx.Lock()
+	defer chainHeight.mtx.Unlock()
+
+	blockHeight, err := rpcclient.GetChainHeight(chainHeight.clientCtx)
+	if err != nil {
+		chainHeight.errGetChainHeight = err
+		return
+	}
+	chainHeight.lastChainHeight = blockHeight
+	chainHeight.errGetChainHeight = nil
+}
+
+// KeepUpdating keeps asking for the current chain
+// height from the node at each nodeStatsCacheInterval.
+func (chainHeight *ChainHeight) KeepUpdating() {
+	for {
+		chainHeight.UpdateChainHeight()
+		time.Sleep(nodeStatsCacheInterval)
+	}
+}
+
+// GetChainHeight returns the last chain height available.
+func (chainHeight *ChainHeight) GetChainHeight() (int64, error) {
+	chainHeight.mtx.RLock()
+	defer chainHeight.mtx.RUnlock()
+
+	return chainHeight.lastChainHeight, chainHeight.errGetChainHeight
+}
--- a/price-feeder/oracle/client/client.go
+++ b/price-feeder/oracle/client/client.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/cosmos/cosmos-sdk/client"
 	"github.com/cosmos/cosmos-sdk/client/flags"
-	rpcClient "github.com/cosmos/cosmos-sdk/client/rpc"
 	"github.com/cosmos/cosmos-sdk/client/tx"
 	"github.com/cosmos/cosmos-sdk/crypto/keyring"
 	sdk "github.com/cosmos/cosmos-sdk/types"
@@ -44,6 +43,7 @@ type (
 		GasAdjustment       float64
 		GRPCEndpoint        string
 		KeyringPassphrase   string
+		ChainHeight         *ChainHeight
 	}
 
 	passReader struct {
@@ -70,7 +70,7 @@ func NewOracleClient(
 		return OracleClient{}, err
 	}
 
-	return OracleClient{
+	oracleClient := OracleClient{
 		Logger:              logger.With().Str("module", "oracle_client").Logger(),
 		ChainID:             chainID,
 		KeyringBackend:      keyringBackend,
@@ -85,7 +85,15 @@ func NewOracleClient(
 		Encoding:            umeeapp.MakeEncodingConfig(),
 		GasAdjustment:       gasAdjustment,
 		GRPCEndpoint:        grpcEndpoint,
-	}, nil
+	}
+
+	clientCtx, err := oracleClient.CreateClientContext()
+	if err != nil {
+		return OracleClient{}, err
+	}
+
+	oracleClient.ChainHeight = ChainHeightInstance(clientCtx)
+	return oracleClient, nil
 }
 
 func newPassReader(pass string) io.Reader {
@@ -125,7 +133,7 @@ func (oc OracleClient) BroadcastTx(nextBlockHeight, timeoutHeight int64, msgs ..
 
 	// re-try voting until timeout
 	for lastCheckHeight < maxBlockHeight {
-		latestBlockHeight, err := rpcClient.GetChainHeight(clientCtx)
+		latestBlockHeight, err := oc.ChainHeight.GetChainHeight()
 		if err != nil {
 			return err
 		}
--- a/price-feeder/oracle/oracle.go
+++ b/price-feeder/oracle/oracle.go
@@ -11,7 +11,6 @@ import (
 	"sync"
 	"time"
 
-	rpcclient "github.com/cosmos/cosmos-sdk/client/rpc"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/rs/zerolog"
 	"golang.org/x/sync/errgroup"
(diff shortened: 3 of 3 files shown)


price-feeder/oracle/client/chain_height.go
```diff
@@ -0,0 +1,79 @@
...
+	"time"
+
+	"github.com/cosmos/cosmos-sdk/client"
+	rpcclient "github.com/cosmos/cosmos-sdk/client/rpc"
+)
+
+const (
+	// chainHeightCacheInterval represents the amount of seconds
+	// to update the chain height.
+	chainHeightCacheInterval = time.Second * 5
```
robert-zaremba, line 14: I think this is too optimistic. There are various edge cases where this won't work, eg, consider the following block height updates (in seconds from 0): 0, 6, 12, 17, ...
the cache will update at 0, 5, 10, 15, 20....

Let's imagine price-feeder will send tx every second. At second 6,7,8, 9, 12, 13,14,15, 17,18,19 the cache will have a wrong value.

robert-zaremba (changes requested):
Cache will have wrong value, even reducing it to 1s is not safe.

price-feeder/oracle/client/chain_height.go
RafilxTenfen: Is there any way to like subscribe to the node to receive a msg every time a new block is generated?

adamewozniak: I don't believe so, that's why we need to build a websocket api for osmosis

RafilxTenfen: Other question would be, what it could go wrong having the block Height at (real block height - 1) ? according to the tick function that already happens every 1 sec 
@robert-zaremba @adamewozniak

brianosaurus: Can we use this to subscribe to the new block event?

https://docs.cosmos.network/master/core/events.html#subscribing-to-events

RafilxTenfen: Good catch! I will try @brianosaurus thanks

robert-zaremba: I didn't check how the height is used, so don't know what can go wrong if we use a wrong height. I think a tx my be rejected. @adamewozniak can you confirm?

RafilxTenfen: Updated to subscribe to event `EventNewBlockHeader` 
@robert-zaremba could you review again?

adamewozniak:
Taking another pass at this shortly

price-feeder/oracle/client/chain_height.go
adamewozniak: Having the right block height at the right time is very important - could lead to missed votes / bad devx when debugging. Low latency & good voting should be prioritized over any performance.

price-feeder/oracle/client/chain_height.go
```diff
@@ -0,0 +1,113 @@
...
+	errGetChainHeight error
+	lastChainHeight   int64
+}
+
+var (
+	// keep the instance thread safe.
+	once sync.Once
+
+	// keeps the single instance of ChainHeight.
+	instance *ChainHeight
```
robert-zaremba, line 35:
```suggestion
	chainHeightCache *ChainHeight
```

robert-zaremba: this variable is in a global scope, so let's make it more meaningful.

robert-zaremba: Also, can we remove it from the global scope?

price-feeder/oracle/client/chain_height.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+
+		go instance.keepUpdating(ctx, rpcClient, newBlockHeaderSubscription)
+	})
+
+	return instance, nil
+}
+
+// UpdateChainHeight receives the data to be updated thread safe.
+func (chainHeight *ChainHeight) UpdateChainHeight(blockHeight int64, err error) {
```
robert-zaremba, line 70: Let's change it to a private function.

price-feeder/oracle/client/chain_height.go
```diff
@@ -0,0 +1,113 @@
...
+) {
+	for {
+		select {
+		case <-ctx.Done():
+			err := eventsClient.Unsubscribe(ctx, tmtypes.EventNewBlockHeader, queryEventNewBlockHeader.String())
+			if err != nil {
+				chainHeight.Logger.Err(err)
+				chainHeight.UpdateChainHeight(chainHeight.lastChainHeight, err)
+			}
+			return
```
robert-zaremba, line 93:
```suggestion
			chainHeight.Logger.Info("closing the ChainHeight subscription")
			return
```

price-feeder/oracle/client/client.go
```diff
@@ -159,6 +173,8 @@ func (oc OracleClient) BroadcastTx(nextBlockHeight, timeoutHeight int64, msgs ..
 				Str("tx_hash", hash).
 				Uint32("tx_code", code).
 				Msg("failed to broadcast tx; retrying...")
+
+			time.Sleep(time.Second * 1)
```
robert-zaremba, line 177: Since we already subscribe to the blocks, why don't we refactor and use that information here as well? Do we need to send new price every second, or it's enough to do it once per block?

price-feeder/oracle/client/chain_height.go
RafilxTenfen: we can, but the intention is to make it an singleton (single instance in the whole bin)

robert-zaremba: But is this function called more than once in the bin?

price-feeder/oracle/client/client.go
RafilxTenfen: actually, this sleep is to avoid the `for` to ask that much the chain height (even if it is cached, I don't see the point of spending process to keep getting the current height)

price-feeder/oracle/client/chain_height.go
RafilxTenfen: not right now... 
but this way, whatever other part on price-feeder needs the block height is going to get from the same instance

price-feeder/oracle/client/chain_height.go
RafilxTenfen: done at https://github.com/umee-network/umee/pull/1069/commits/1a19e3715ccf1a7641df54326edb5870d3b21a59

price-feeder/oracle/client/chain_height.go
adamewozniak: Height is not used as a part of the tx, it's used by the price feeder to time when to submit which txs.

price-feeder/oracle/client/client.go
robert-zaremba: I see, I didn't unroll the whole function body. Should we have max tries instead of max height?

price-feeder/oracle/client/chain_height.go
robert-zaremba: OK, maybe we can keep it. But in general, using global variables is a bad pattern and usually leads to technical debt. So if there is a way to avoid it then let's do it.

RafilxTenfen: we can remove it, but if in another part of the code, that is being called, we should use an single instance

price-feeder/oracle/client/client.go
RafilxTenfen: @adamewozniak what do you think?

price-feeder/oracle/client/chain_height.go
RafilxTenfen: [singleton](https://medium.com/golang-issue/how-singleton-pattern-works-with-golang-2fdd61cd5a7f) removed in https://github.com/umee-network/umee/pull/1069/commits/03a356f61401903971533335d75821a977cd2f1d

price-feeder/oracle/client/chain_height.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+var (
+	errParseEventDataNewBlockHeader = errors.New("error parsing EventDataNewBlockHeader")
+	queryEventNewBlockHeader        = tmtypes.QueryForEvent(tmtypes.EventNewBlockHeader)
+)
+
+// ChainHeight is used to cache the chain height of the
+// current node which is being updated each time the
+// node sends an event of EventNewBlockHeader.
```
robert-zaremba, line 21:
```suggestion
// node sends an event of EventNewBlockHeader.
// It starts a goroutine to subscribe to blockchain new block event and update the cached height.
```

price-feeder/oracle/client/chain_height.go
```diff
@@ -0,0 +1,105 @@
...
+type ChainHeight struct {
+	Logger zerolog.Logger
+
+	mtx               sync.RWMutex
+	errGetChainHeight error
+	lastChainHeight   int64
+}
+
+// NewChainHeight returns a new ChainHeight struct that
+// already subscribe to EventNewBlockHeader.
```
robert-zaremba, line 31:
```suggestion
// starts a new goroutine subscribed to EventNewBlockHeader.
```

robert-zaremba (approved):
lgtm, added few code-comments suggestions

price-feeder/CHANGELOG.md
```diff
@@ -50,6 +50,7 @@ Ref: https://keepachangelog.com/en/1.0.0/
 - [#1032](https://github.com/umee-network/umee/pull/1032) Update the accepted tvwap period from 3 minutes to 5 minutes.
 - [#978](https://github.com/umee-network/umee/pull/978) Cleanup the oracle package by moving deviation & conversion logic.
 - [#1050](https://github.com/umee-network/umee/pull/1050) Cache x/oracle params to decrease the number of queries to nodes.
+- [#1069](https://github.com/umee-network/umee/pull/1069) Cache chain height to subscribe to node event EventNewBlockHeader.
```
adamewozniak, line 53: Is this really a _cache_ in the normal sense anymore? I tend to associate a cache with latency - wdyt?
Let's at least make sure that the code & comments are clear that the block height is always up-to-date.

price-feeder/oracle/client/chain_height.go
```diff
@@ -0,0 +1,106 @@
...
+	chainHeight.mtx.Lock()
+	defer chainHeight.mtx.Unlock()
+
+	chainHeight.lastChainHeight = blockHeight
+	chainHeight.errGetChainHeight = err
+}
+
+// keepUpdating keeps listening to new blocks being made
+// and updates the chain height.
+func (chainHeight *ChainHeight) keepUpdating(
```
adamewozniak, line 72:
```suggestion
// subscribe listens to new blocks being made
// and updates the chain height.
func (chainHeight *ChainHeight) subscribe(
```

price-feeder/oracle/client/client.go
```diff
@@ -125,7 +139,7 @@ func (oc OracleClient) BroadcastTx(nextBlockHeight, timeoutHeight int64, msgs ..
 
 	// re-try voting until timeout
 	for lastCheckHeight < maxBlockHeight {
-		latestBlockHeight, err := rpcClient.GetChainHeight(clientCtx)
+		latestBlockHeight, err := oc.ChainHeight.GetChainHeight()
 		if err != nil {
 			return err
```
adamewozniak, line 144: What happens here? E.g., in the case of the chain stopping, chainheight returns an error, and the loop continues? 

Just want to make sure that in the event of a chain halt, validators would not have to restart all their price feeders... they'll likely be hands-off and we wouldn't want to have to worry about everyone getting jailed

adamewozniak (approved):
great job!! lgtm + a couple nits 💯

price-feeder/oracle/client/client.go
RafilxTenfen: Well, if the chain stops it will only return an error if the unsubscribe of the EventNewBlockHeader returns an error, On the old rpcClient.GetChainHeight what would need to happen to return an error?

price-feeder/CHANGELOG.md
RafilxTenfen: updated

price-feeder/oracle/client/client.go
adamewozniak: RPC would have to be unavailable - sounds like this loop wouldn't stop though 👍

robert-zaremba: that's why I proposed to use `maxTries` instead. @RafilxTenfen can you add a task for that?

PR #1084, opened 2022-07-01
fix: price-feeder height subscription cleanup

adamewozniak:
## Description

Just small cleanup around the price-feeder height subscription logic

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/price-feeder/oracle/client/chain_height.go
+++ b/price-feeder/oracle/client/chain_height.go
@@ -13,6 +13,7 @@ import (
 
 var (
 	errParseEventDataNewBlockHeader = errors.New("error parsing EventDataNewBlockHeader")
+	errNoBlockHeightEvents          = errors.New("no block height events have been received")
 	queryEventNewBlockHeader        = tmtypes.QueryForEvent(tmtypes.EventNewBlockHeader)
 )
 
@@ -101,6 +102,11 @@ func (chainHeight *ChainHeight) subscribe(
 func (chainHeight *ChainHeight) GetChainHeight() (int64, error) {
 	chainHeight.mtx.RLock()
 	defer chainHeight.mtx.RUnlock()
+
+	if chainHeight.lastChainHeight == 0 &&
+		chainHeight.errGetChainHeight == nil {
+		chainHeight.errGetChainHeight = errNoBlockHeightEvents
+	}
 
 	return chainHeight.lastChainHeight, chainHeight.errGetChainHeight
 }
--- a/price-feeder/oracle/client/chain_height.go
+++ b/price-feeder/oracle/client/chain_height.go
@@ -3,6 +3,7 @@ package client
 import (
 	"context"
 	"errors"
+	"fmt"
 	"sync"
 
 	"github.com/rs/zerolog"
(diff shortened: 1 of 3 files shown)


price-feeder/oracle/client/chain_height.go
```diff
@@ -102,5 +103,10 @@ func (chainHeight *ChainHeight) GetChainHeight() (int64, error) {
 	chainHeight.mtx.RLock()
 	defer chainHeight.mtx.RUnlock()
 
+	if chainHeight.lastChainHeight == 0 &&
+		chainHeight.errGetChainHeight == nil {
+		chainHeight.errGetChainHeight = errNoBlockHeightEvents
+	}
```
adamewozniak, line 109: I'm in between doing this (seems a bit lazy, but at least more clear), or adding an rpc query here to get the height (would be more effective, but make the code a bit messier) 

wdyt @RafilxTenfen ?

price-feeder/oracle/client/chain_height.go
```diff
@@ -102,5 +103,10 @@ func (chainHeight *ChainHeight) GetChainHeight() (int64, error) {
 	chainHeight.mtx.RLock()
 	defer chainHeight.mtx.RUnlock()
 
+	if chainHeight.lastChainHeight == 0 &&
+		chainHeight.errGetChainHeight == nil {
```
robert-zaremba, line 107: I think the condition is wrong, seams that once the err is set it will never be unset. Also why checking `lastChainHeight == 0`? since the chain is running and already above 0, then we will never set `chainHeight.errGetChainHeight`

robert-zaremba (commented):
condition is wrong

price-feeder/oracle/client/chain_height.go
adamewozniak: This condition occurs when the tick runs before a new height is received from the subscription. Alternatively we could just fill the height with a query when we initialize the object so that this does not happen.

robert-zaremba: we should initialize with the right value , rather than zero

price-feeder/oracle/client/chain_height.go
RafilxTenfen: I think we could start the ChainHeight with this error like 
```go

	chainHeight := &ChainHeight{
		Logger:            logger.With().Str("oracle_client", "chain_height").Logger(),
		errGetChainHeight: errNoBlockHeightEvents,
		lastChainHeight:   0,
	}

```
This way when a new block appears, it will throw the event, and from now on the `chainHeight` structure would be updated with the current block height and nil on the errGetChainHeight

adamewozniak: Hmm, I'm kind of leaning towards just querying for the starting height at the object initialization... that seems more clean than initializing with an error imo, and would not impact performance

price-feeder/oracle/client/client.go
```diff
@@ -93,7 +93,17 @@ func NewOracleClient(
...
 	}
 
-	chainHeight, err := NewChainHeight(context.Background(), clientCtx.Client, oracleClient.Logger)
+	blockHeight, err := rpc.GetChainHeight(clientCtx)
+	if err != nil {
+		return OracleClient{}, err
+	}
+
+	chainHeight, err := NewChainHeight(
+		context.Background(),
```
robert-zaremba, line 102: We need some top down context, rather than using Background in the middle of the flow. We can handle it later.

robert-zaremba (approved):
Let's handle the context issue here or in an new task

price-feeder/oracle/client/client.go
adamewozniak: I'll put it in a follow-up PR

PR #1087, opened 2022-07-03
docs: cleanup ADR-002

toteki:
Simplifies (and renames) ADR-002.

Does not add or change any design features.


docs/architecture/ADR-002-supply-assets.md
```diff
@@ -0,0 +1,61 @@
...
+- September 29, 2021: Update message names to reflect implementation (@toteki)
+- July 2, 2022: Cleanup and simplify (@toteki)
+
+## Status
+
+Implemented
+
+## Context
+
+One of the base functions of the `x/leverage` module is to allow users to supply assets, and earn interest on their deposits.
```
toteki, line 15:
```suggestion
The `x/leverage` module allows users to supply and withdraw assets.
```

robert-zaremba (commented):
Let's avoid file rename - it looses the track and it's hard to read the diff.

toteki:
I'll undo it here - I think if it's still needed later I can do a rename PR (without a big file changes) that keeps tracking the file.

docs/architecture/ADR-002-deposit-assets.md
```diff
@@ -1,129 +1,60 @@
...
 ## Consequences
 
 ### Positive
-
-- Banking module already provides underlying functionality
-- Asset deposit does not require us to modify state beyond what is already done by the `x/bank` module (e.g. account balances): deposit of an assets into a module account is atomic with the minting of respective uTokens to the sender's account, and uToken ownership is the sole requirement for asset withdrawal. We do not need to track individual user deposit history.
+- [x/bank module]([Cosmos Bank Keeper](https://github.com/cosmos/cosmos-sdk/blob/v0.44.0/x/bank/spec/02_keepers.md)) provides the underlying functionality.
 
 ### Negative
+- If tokens were to be sent unexpectedly to the `x/leverage` module account (see the warning on [this page](https://docs.cosmos.network/master/modules/bank/)), it would have the effect of donating said tokens to existing suppliers if the tokens were an accepted type. Otherwise, the tokens would remain inert.
```
robert-zaremba, line 54:
```suggestion
- If tokens were to be sent unexpectedly to the `x/leverage` module account (see the warning on [this page](https://docs.cosmos.network/master/modules/bank/)), it would have the effect of donating said tokens to existing suppliers if the tokens were an accepted type. Otherwise, the tokens would remain inert.
- uToken value is not the same as the underlying Token - can create some confusions
- need to involve extra calculation (exchange rates) for all uToken operations
```

robert-zaremba (approved):
Note for the future, I don't think this should be a separate ADR ;) But here we are.

docs/architecture/ADR-002-deposit-assets.md
toteki: In general, our approach to ADRs has been
- Create them one at a time, recording our reasoning as design decisions are made
- Edit later to improve the text, but not substantially alter the decisions themselves
- Create separate ADR if a major change to the design is decided much later (e.g. `Adjusted Borrow * Interest Scalar` replacing `Borrow` & `InterestEpoch`)

Contrast with the spec, which is considered a single document describing the system as it currently stands.

robert-zaremba:
ADRs should be significant. Adding a new message doesn't qualify for an ADR. Normally, ADR should document the whole process, algorithm and a structure of a solution with a significant design impact. It's not an idea nor a document explaining a new feature - and some of the ADRs we have are just capturing a spec for the functionality. This is not what an ADR is in a general sense.

> In software engineering and software architecture design, architectural decisions are design decisions that address architecturally significant requirements; they are perceived as hard to make and/or costly to change.

+ https://en.wikipedia.org/wiki/Architectural_decision
+ http://www.unified-am.com/UAM/UAM/guidances/guidelines/uam_architectural_decision_BAE7AFA2.html

Wanted to note that instead of 9 ADRs we should probably have 3-4.

toteki:
Should we add an issue to the backlog where we consolidate?

PR #940, opened 2022-05-23
docs: ADR-4: max collateral utilization

robert-zaremba:
## Description

closes: #926 

+ adding max collateral utilization spec
+ cleaning the structure and language of other ADRs (done it as I was reviewing places which are impacted by max collateral utilization)

----

### Author Checklist

*All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues.*

I have...

- [ ] included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] added appropriate labels to the PR
- [ ] added `!` to the type prefix if API or client breaking change
- [ ] targeted the correct branch (see [PR Targeting](https://github.com/umee-network/umee/blob/main/CONTRIBUTING.md#pr-targeting))
- [ ] provided a link to the relevant issue or specification
- [ ] added a changelog entry to `CHANGELOG.md`
- [ ] included comments for [documenting Go code](https://blog.golang.org/godoc)
- [ ] updated the relevant documentation or specification
- [ ] reviewed "Files changed" and left comments if necessary
- [ ] confirmed all CI checks have passed

### Reviewers Checklist

*All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items.*

I have...

- [ ] confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] confirmed `!` in the type prefix if API or client breaking change
- [ ] confirmed all author checklist items have been addressed
- [ ] reviewed state machine logic
- [ ] reviewed API design and naming
- [ ] reviewed documentation is accurate
- [ ] reviewed tests and test coverage
- [ ] manually tested (if applicable)

PR #1088, opened 2022-07-03
chore: rename borrow utilization to supply utilization

toteki:
Extracted from #940 to minimize diff

This renames `borrow utilization` to `supply utilization` in various comments, docs, and code in the leverage module.

No queries or message fields needed to be renamed, so it is not API-breaking.


robert-zaremba (changes requested):
Please see: https://github.com/umee-network/umee/pull/940#issuecomment-1173482340

If there are additional changes introduced in this PR then we can merge it after #940 is changed. So blocking until #940 will get merged. I don't want to solve additional conflicts.

Issue #898, opened 2022-05-12
x/leverage: Simplify Liquidate computation (RV #28-30)


toteki:
# Runtime Verification Audit Items 28-30

There are several simplifications that can be applied to the Liquidate function
- Maximum allowed repayment check can be improved (28)
- Liquidator minimum reward ratio check only needs price ratio and liquidation incentive (29)
- General simplification (30)

Suggestion from item 30 to be considered
```
rewardTokenPrice, err := k.TokenPrice(ctx, desiredReward.Denom)
if err != nil {
  return sdk.ZeroInt(), sdk.ZeroInt(), err
}
repaymentTokenPrice, err := k.TokenPrice(ctx, desiredRepayment.Denom)
if err != nil {
  return sdk.ZeroInt(), sdk.ZeroInt(), err
}

if rewardTokenPrice.IsZero() || repaymentTokenPrice.IsZero() {
  return sdk.ZeroInt(), sdk.ZeroInt(), types.ErrInvalidOraclePrice
}
eXRateRepayToRewardDenom := repaymentTokenPrice.Quo(rewardTokenPrice)

// get reward-specific incentive and dynamic close factor
baseRewardDenom := desiredReward.Denom
liquidationIncentive, closeFactor, err := k.LiquidationParams(ctx, baseRewardDenom, borrowValue, liquidationLimit)
if err != nil {
  return sdk.ZeroInt(), sdk.ZeroInt(), err
}

if desiredReward.Amount.IsPositive() {
  //The actual reward ratio guaranteed the liquidation event is successful
  actualRewardRatio := eXRateRepayToRewardDenom.Mul(sdk.OneDec().Add(liquidationIncentive))
  // user-controlled minimum ratio of reward to repayment, expressed in base:base assets (not uTokens)
  minimumRewardRatio := desiredReward.Amount.ToDec().Quo(desiredRepayment.Amount.ToDec())

  // confirm if the end of liquidation reward ratio meets liquidator's expectation of minimum reward ratio
  if actualRewardRatio.LT(minimumRewardRatio) {
     return sdk.ZeroInt(), sdk.ZeroInt(), types.ErrLiquidationRewardRatio
  }
}

// actual repayment starts at desiredRepayment but can be lower due to limiting factors
repayment := desiredRepayment

// get liquidator's available balance of base asset to repay
liquidatorBalance := k.bankKeeper.SpendableCoins(ctx, liquidatorAddr).AmountOf(repayment.Denom)

[...]

Issue #828, opened 2022-04-22
x/leverage: overly restrictive behavior in MsgLiquidate's ValidateBasic


toteki:
Documented behavior of `MsgLiquidate` says it should interpret zero reward amount as "no user-set minimum".

Instead, the following code rejects Amount = 0
```go
// x/leverage/types/tx.go
// func (msg *MsgLiquidate) ValidateBasic()
	if asset := msg.GetReward(); !asset.IsValid() {
		return sdkerrors.Wrap(ErrInvalidAsset, asset.String())
	}
```

The desired behavior for amount=0 can be reproduced by setting amount=1, but we should preferabkly fix the ValidateBasic.

toteki:
Misdiagnosis - it's not asset.IsValid(), because that accepts zero, but something's stopping CLI liquidation with 0 reward
```
Error: rpc error: code = InvalidArgument desc = failed to execute message; message index: 0: 0ibc/27394FB092D2ECCD56123C74F36E4C1F926001CEADA9CA97EA622B25F41E5EB2: invalid asset: invalid request
```

robert-zaremba:
`Coin.IsValid()` rejects only negative amounts. So yes, there must be other error.

robert-zaremba:
Do you think we need it for mainnet?

toteki:
Better to fix this now than have to push a state-breaking fix later, since it's a small difference.

Issue #831, opened 2022-04-24
x/leverage: unswept bad debt appearing in testnet


toteki:
## Summary of Bug

Multiple positions like the following have persisted on the testnet, despite module reserves appearing not to be exhausted.

> borrowed:
> - amount: "1546498"
>   denom: ibc/27394FB092D2ECCD56123C74F36E4C1F926001CEADA9CA97EA622B25F41E5EB2
> - amount: "209227"
>   denom: uumee

> collateral: []

## Version

`2.0.0`

toteki:
First I'll finish initial investigations using `poke` node:
- Create a test `umeed` binary which exposes queries for module balance and bad debt addresses
- Sync with test network
- Compare bad debt addresses with observed liquidation targets
- Check module balance in relation to reserves and observed bad debt
- Determine whether this is an as-expected reserve exhaustion (after the massive price change) or actual bug

toteki:
This was before liquidation was refactored and it has not reappeared, so I'll close it

PR #1118, opened 2022-07-07
feat!: liquidation directly rewards base assets

toteki:
## Description

Significant Change:
- Instead of receiving uToken rewards on `MsgLiquidate`, liquidator receives equivalent base assets.

Reasoning:

We're about to add `MaxCollateralUtilization` which restricts `MsgWithdraw` at critical times, so `AvailableSupply` can be used for liquidation.

But if liquidators receive their rewards as uTokens, then the `MsgWithdraw` restriction will apply to them just like any other user, and they will be unable to receive their base tokens.

Switching to base assets received eliminates makes `MaxCollateralUtilization` work as intended, and also simplifies the `MsgLiquidate -> MsgWithdraw -> IBC (to sell reward)` liquidation workflow to just `MsgLiquidate -> IBC`

---

API Breaking:
- Reverts `MsgLiquidate`'s reward `sdk.Coin` field to reward_denom `string`

Reasoning:

That field was for rare cases where liquidators do not trust the umee `x/oracle` and want to put their own price floors. Such cases should be handled off-chain (just query prices) rather than adding a computation and required field for all users. See #828

---

closes: #898 
closes: #828

relevant to #831 - because rounding up repayment and reward eliminates most "dust" liquidation remainders

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/leverage/keeper/keeper.go
```diff
@@ -357,215 +357,68 @@ func (k Keeper) RemoveCollateral(ctx sdk.Context, borrowerAddr sdk.AccAddress, c
...
-		for _, coin := range borrowed {
-			// Mark repayment denom as bad debt only if some debt remains after
-			// this liquidation. All other borrowed denoms were definitely not
-			// repaid in this liquidation so they are always marked as bad debt.
-			if coin.Denom != repayment.Denom || owed.IsPositive() {
-				if err := k.setBadDebtAddress(ctx, borrowerAddr, coin.Denom, true); err != nil {
-					return sdk.ZeroInt(), sdk.ZeroInt(), err
-				}
+	// detect bad debt if collateral is completely exhausted
+	if k.GetBorrowerCollateral(ctx, borrowerAddr).IsZero() {
```
RafilxTenfen, line 413: `isZero` when this could happen?

toteki: This happens when a liquidation uses up the last of a borrower's collateral. If there are any borrows left, that's Bad Debt.

x/leverage/keeper/keeper.go
```diff
@@ -357,215 +357,68 @@ func (k Keeper) RemoveCollateral(ctx sdk.Context, borrowerAddr sdk.AccAddress, c
...
-	if !k.IsAcceptedToken(ctx, desiredReward.Denom) {
-		return sdk.ZeroInt(), sdk.ZeroInt(), types.ErrInvalidAsset.Wrap(desiredReward.String())
-	}
-
-	collateral := k.GetBorrowerCollateral(ctx, borrowerAddr)
-	borrowed := k.GetBorrowerBorrows(ctx, borrowerAddr)
-
-	borrowValue, err := k.TotalTokenValue(ctx, borrowed) // total borrowed value in USD
+	ctx sdk.Context, liquidatorAddr, borrowerAddr sdk.AccAddress, desiredRepay sdk.Coin, rewardDenom string,
+) (sdk.Coin, sdk.Coin, sdk.Coin, error) {
```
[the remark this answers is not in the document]
toteki: done - should be more clear now

x/leverage/keeper/keeper.go
```diff
@@ -346,226 +346,76 @@ func (k Keeper) Decollateralize(ctx sdk.Context, borrowerAddr sdk.AccAddress, co
...
-	borrowed := k.GetBorrowerBorrows(ctx, borrowerAddr)
-
-	borrowValue, err := k.TotalTokenValue(ctx, borrowed) // total borrowed value in USD
+// Liquidate attempts to repay one of an eligible borrower's borrows (in part or in full) in exchange
+// for a the base token equivalent of selected denomination of the borrower's uToken collateral. If the
+// borrower is not over their liquidation limit, or the repayment or reward denominations are invalid,
+// an error is returned. If the attempted repayment is greater than the amount owed or the maximum that
+// can be repaid due to parameters or available balances, then a partial liquidation, equal to the maximum
+// valid amount, is performed. Because partial liquidation is possible and exchange rates vary, Liquidate
+// returns the actual amount of tokens repaid, uTokens consumed, and base tokens rewarded (in that order).
```
robert-zaremba, line 355:
```suggestion
// returns the actual amount of tokens repaid, uTokens consumed, and base tokens rewarded (uToken -> Token).
```

x/leverage/keeper/liquidate.go
```diff
@@ -0,0 +1,167 @@
...
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+
+	"github.com/umee-network/umee/v2/x/leverage/types"
+)
+
+// liquidationOutcome takes a repayment and reward denom proposed by a liquidator and calculates
+// the maximum repayment amounts a target address is eligible for, and the corresponding reward
+// amounts using current oracle, params, and available balances. Inputs must be registered tokens.
+// Outputs are base token repayment, uToken collateral cost, and base token reward from liquidation.
+func (k Keeper) liquidationOutcome(
```
robert-zaremba, line 14:
```suggestion
func (k Keeper) computeLiquidation(
```

x/leverage/keeper/math.go
```diff
@@ -27,3 +27,29 @@ func ApproxExponential(x sdk.Dec) sdk.Dec {
 	sum = sum.Add(next.QuoInt64(6)) // 3!
 	return sum                      // approximated e^x
 }
+
+// ReduceProportionally accepts two sdk.Int to be interpreted as a fraction (a/b), and
+// any number of pointers to sdk.Int which will be multiplied by (a/b) if a < b, then
```
robert-zaremba, line 32:
```suggestion
// any number of pointers to sdk.Int which will be multiplied (and updated) by (a/b) if a < b, then
```

x/leverage/keeper/math.go
```diff
@@ -27,3 +27,29 @@ func ApproxExponential(x sdk.Dec) sdk.Dec {
...
+// rounded up. If a >= b or b == 0 this is a no-op.
+func ReduceProportionally(a, b sdk.Int, nums ...*sdk.Int) {
+	if a.GTE(b) || b.IsZero() {
+		return
+	}
+	ratio := a.ToDec().Quo(b.ToDec()) // a/b
+	for _, n := range nums {
+		*n = n.ToDec().Mul(ratio).Ceil().RoundInt()
```
robert-zaremba, line 40:
```suggestion
	for _, n := range nums {
		*n = n.Mul(a).Quo(b)
```

robert-zaremba: This is an faster version. No need to use `Dec`.

x/leverage/keeper/math.go
```diff
@@ -27,3 +27,29 @@ func ApproxExponential(x sdk.Dec) sdk.Dec {
...
+	}
+}
+
+// ReduceProportionallyDec accepts two sdk.Dec to be interpreted as a fraction (a/b), and
+// any number of pointers to sdk.Int which will be multiplied by (a/b) if a < b, then
```
robert-zaremba, line 45:
```suggestion
// any number of pointers to sdk.Int which will be multiplied (and updated) by (a/b) if a < b, then
```

x/leverage/keeper/keeper.go
```diff
@@ -346,226 +346,76 @@ func (k Keeper) Decollateralize(ctx sdk.Context, borrowerAddr sdk.AccAddress, co
...
-			// Mark repayment denom as bad debt only if some debt remains after
-			// this liquidation. All other borrowed denoms were definitely not
-			// repaid in this liquidation so they are always marked as bad debt.
-			if coin.Denom != repayment.Denom || owed.IsPositive() {
-				if err := k.setBadDebtAddress(ctx, borrowerAddr, coin.Denom, true); err != nil {
-					return sdk.ZeroInt(), sdk.ZeroInt(), err
-				}
+	// detect bad debt if collateral is completely exhausted
+	if k.GetBorrowerCollateral(ctx, borrowerAddr).IsZero() {
+		// TODO: exclude blacklisted collateral
```
robert-zaremba, line 411: Is there a task for that? if not then let's add it and include in the comment

x/leverage/keeper/math.go
```diff
@@ -27,3 +27,29 @@ func ApproxExponential(x sdk.Dec) sdk.Dec {
...
+// rounded up. If a >= b or b == 0 this is a no-op.
+func ReduceProportionallyDec(a, b sdk.Dec, nums ...*sdk.Int) {
+	if a.GTE(b) || b.IsZero() {
+		return
+	}
+	ratio := a.Quo(b)
+	for _, n := range nums {
+		*n = n.ToDec().Mul(ratio).Ceil().RoundInt()
```
robert-zaremba, line 53: Shouldn't we use `Floor` here?

x/leverage/keeper/math.go
```diff
@@ -27,3 +27,29 @@ func ApproxExponential(x sdk.Dec) sdk.Dec {
 	sum = sum.Add(next.QuoInt64(6)) // 3!
 	return sum                      // approximated e^x
 }
+
+// ReduceProportionally accepts two sdk.Int to be interpreted as a fraction (a/b), and
+// any number of pointers to sdk.Int which will be multiplied by (a/b) if a < b, then
+// rounded up. If a >= b or b == 0 this is a no-op.
```
robert-zaremba, line 33:
```suggestion
// rounded down. If a >= b or b == 0 this is a no-op.
```

x/leverage/keeper/math.go
```diff
@@ -27,3 +27,29 @@ func ApproxExponential(x sdk.Dec) sdk.Dec {
...
+	}
+	ratio := a.ToDec().Quo(b.ToDec()) // a/b
+	for _, n := range nums {
+		*n = n.ToDec().Mul(ratio).Ceil().RoundInt()
+	}
+}
+
+// ReduceProportionallyDec accepts two sdk.Dec to be interpreted as a fraction (a/b), and
+// any number of pointers to sdk.Int which will be multiplied by (a/b) if a < b, then
+// rounded up. If a >= b or b == 0 this is a no-op.
```
robert-zaremba, line 46:
```suggestion
// rounded down. If a >= b or b == 0 this is a no-op.
```

robert-zaremba (changes requested):
Actually, I think we should not reduce `available_supply` by directly giving a `tokenA` (instead of `uToken`) to the liquidator. We want available supply to stay high (to optimize `collateral_utilization(tokenA)`).
So, I think we should:
* keep the liquidation logic as is (liquidator receives uTokens)
* round down when distributing rewards.

toteki:
> Keep the liquidation logic as is (liquidator receives uTokens)

Suppose we have a situation where collateral utilization has reached its limit. In this case, `MsgWithdraw` from suppliers is rejected in order to keep the remaining collateral available for liquidation. I believe our discussion was that liquidators need to be able to function in this crisis, which is why collateral utilization needed to be limited in the first place.

However, there is no way to distinguish between a _supplier_ and a _liquidator_ if they both have to use `MsgWithdraw`. If the liquidator receives uTokens, then they can't sell liquidated collateral (`Liquidate` -> `Withdraw` -> `IBC to Osmosis`) and therefore will prefer not to liquidate at all.

toteki:
> round down when distributing rewards

This change to rounding up is related to a dust problem I witnesses on chain during umeemania. Consider the following scenario

> Borrowed `7 uosmo`
> Collateral `1 u/uatom` worth `1.1 uatom`
> Assume due to exchange rates that `1 ATOM = 8.1 OSMO`

Note the microatom and microosmo, which cannot be subdivided further.

The transaction `Liquidate(3uosmo -> uatom)` will have the following behavior
| Rounding | Repaid | Collateral | Reward | Problem |
| - | - | - | - | - |
| Down for all | 3 | 0 | 0 | Collateral dust remains no matter what, liquidation is not incentivized, next liquidate will repeat and repay 3 more uosmo for 0 reward, and bad debt repayment is not reached because collateral never reaches zero |
| Up for all | 3 | 1 | 1 | Base token reward is slightly (1 micro) higher than if collateral had been input into `MsgWithdraw` |
| Up for collateral, Down for reward | 3 | 1 | 0 | Liquidation is not incentivized |

The priority for rounding during liquidation in my opinion should be to not leave dust. This is because
- Bad debt repayment requires collateral to reach exactly zero
- Collateral dust in a denomination makes that collateral denomination appear to be an eligible `RewardDenom` during the _next_ liquidation if multiple liquidations happen over time. This could lead to loops where bots cannot eliminate the last `1` collateral no matter how many times they liquidate.

That said, I'll look into the `Collateral -> Reward` (uToken -> Token) rounding to make sure it can't be expoited (e.g. repeated `MsgLiquidate` on a controlled borrower, each time adding `1` collateral, to drain funds from the module by rounding up uToken exchange rate.

Since that looks like it might be the case, I'll change the code to round up for collateral consumed, and down for base tokens rewarded. I'll await further discussion on the rest.

x/leverage/keeper/math.go
toteki: See rounding discussion

RafilxTenfen:
https://github.com/umee-network/umee/pull/1118#issuecomment-1183318250
This is great context, whatever the decision, this should be in the repo `.md` file as a decision log kind of thing or even as a comment on the function

proto/umee/leverage/v1/tx.proto
```diff
@@ -116,6 +115,7 @@ message MsgRepayResponse {
 
 // MsgLiquidateResponse defines the Msg/Liquidate response type.
 message MsgLiquidateResponse {
-  cosmos.base.v1beta1.Coin repaid = 1 [(gogoproto.nullable) = false];
-  cosmos.base.v1beta1.Coin reward = 2 [(gogoproto.nullable) = false];
+  cosmos.base.v1beta1.Coin repaid     = 1 [(gogoproto.nullable) = false];
+  cosmos.base.v1beta1.Coin collateral = 2 [(gogoproto.nullable) = false];
+  cosmos.base.v1beta1.Coin reward     = 3 [(gogoproto.nullable) = false];
```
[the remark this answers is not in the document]
toteki: Added clarifying comments to proto

proto/umee/leverage/v1/tx.proto
```diff
@@ -83,15 +83,14 @@ message MsgRepay {
...
   // Liquidator is the account address performing a liquidation and the signer
   // of the message.
-  string                   liquidator = 1;
-  string                   borrower   = 2;
-  cosmos.base.v1beta1.Coin repayment  = 3 [(gogoproto.nullable) = false];
-  cosmos.base.v1beta1.Coin reward     = 4 [(gogoproto.nullable) = false];
+  string                   liquidator   = 1;
+  string                   borrower     = 2;
+  cosmos.base.v1beta1.Coin repayment    = 3 [(gogoproto.nullable) = false];
+  string                   reward_denom = 4;
```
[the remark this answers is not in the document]
robert-zaremba: In that case the complexity is moved to liquidator... he will need to calculate which denom he can liquidate and how much. 

We are already adjusting the repayment. Instead we can go one by one of the collateral list and sum app to the value related to the repayment. We could use denom list to make it simpler.

robert-zaremba:
> The priority for rounding during liquidation in my opinion should be to not leave dust

The problem is not in dust, but in negative balance. System must not have any negative balance if possible. If there is a diff, then it should go to a module rewards pool. In your example (7uosmo borrowed, 1uatom collateral) we can't do up for all because it will leave the position uncollaterized and create a negative balance (liquidator will receive 1uatom for paying 3uosmo, so the system has -4uosmo balance). That's why my suggestion to round down the reward. Liquidator will receive a bit less, but we are talking about the lowest denominator.

toteki:
See the [Bad Debt Repayment ADR](https://github.com/umee-network/umee/blob/main/docs/architecture/ADR-007-bad-debt.md) - the role of liquidators in that case is to take care of some of the borrower's debt using any remaining collateral. After that we need the module to pay off the remainder using reserves so it does not continue to accrue interest.

If collateral dust is left over, bad debt repayment does not trigger in its current implementation.

proto/umee/leverage/v1/tx.proto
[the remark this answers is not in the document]
toteki: Yes, the liquidator can do that - I had a script doing as much during the testnet, and what it does is proceed in order of preference for repay and reward denoms by offering the maximum amount it is willing to repay.

For most borrowers, even with multiple borrowed and collateral types, the first `MsgLiquidate` brings them back to health, so the liquidator's initial preference for reward and repay denoms is enough.

For borrowers where multiple transactions are needed, the liquidator only needs to follow their ranked order of preferences, because the first transaction will have either exhausted a repay denom or a reward denom by succeeding.

Note: Exhausted a denom can mean:
- Repaid a borrow completely in one token denom
- Reduced liquidator's token balance to zero (or consumed maximum offered amount) for a token denom
- Reduced borrower's collateral position in a uToken denom to zero
- Exhausted available supply of a reward token

RafilxTenfen (approved):

x/leverage/keeper/keeper.go
toteki: Added this in a commit this morning

robert-zaremba:
Let's consider the following case:
* Gold token is worth 1000 USD and Stone token is worth 1 USD
* Liquidation threshold is 80% and liquidation premium is 10%
* Alice puts 1gold as a collateral and borrows 800stone
* stone drops and Alice position has to be liquidated.
Let's say we liquidate all borrowed position. With 10% premium, liquidator should get 880USD. But there is only one token to liquidate. So liquidator should repay 800 stones and get the whole 1gold token.

Could you add this test?

toteki:
Yeah - I'll add the test, and also ensure that if `1 u/GOLD` is worth `1.01 GOLD` that the liquidator receives `1 GOLD` and not `2` or `0`.

I might divide the function up a little too, so it's easier to write test cases for without mocking oracle prices.

x/leverage/keeper/liquidate.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/umee-network/umee/v2/x/leverage/types"
+)
+
+// computeLiquidation takes a repayment and reward denom proposed by a liquidator and calculates
+// the actual repayment amount a target address is eligible for, and the corresponding collateral
+// to burn and rewards to return to the liquidator.
+func (k Keeper) computeLiquidation(
```
toteki, line 12: Refactor note: This un-exported keeper function handles the parts of liquidation that require a the leverage `Keeper` - mostly fetching balances and parameters. It then calls the purely mathematical function in `types` package.

x/leverage/keeper/oracle.go
```diff
@@ -13,17 +13,15 @@ import (
...
+// and the exponent. When error is nil, price is guaranteed to be positive.
 func (k Keeper) TokenPrice(ctx sdk.Context, denom string) (sdk.Dec, error) {
 	t, err := k.GetTokenSettings(ctx, denom)
 	if err != nil {
 		return sdk.ZeroDec(), err
 	}
 
 	if t.Blacklist {
-		return sdk.ZeroDec(), nil
+		return sdk.ZeroDec(), types.ErrBlacklisted
```
toteki, line 24: I was able to restore the strict nonzero behavior of TokenPrice as a result of other functions filtering blacklisted tokens before calling TokenPrice.

x/leverage/client/tests/tests.go
```diff
@@ -386,7 +377,7 @@ func (s *IntegrationTestSuite) TestLeverageScenario() {
 		cli.GetCmdRepay(),
 		[]string{
 			val.Address.String(),
-			"51uumee",
+			"50uumee",
```
toteki, line 380: Changes in rounding behavior have created a number of simpler decimals in this test file.

x/leverage/types/leverage.go
```diff
@@ -0,0 +1,137 @@
...
+	availableRepay sdk.Int,
+	availableCollateral sdk.Int,
+	availableReward sdk.Int,
+	repayTokenPrice sdk.Dec,
+	rewardTokenPrice sdk.Dec,
+	uTokenExchangeRate sdk.Dec,
+	liquidationIncentive sdk.Dec,
+	closeFactor sdk.Dec,
+	borrowedValue sdk.Dec,
+) (tokenRepay sdk.Int, collateralBurn sdk.Int, tokenReward sdk.Int) {
```
toteki, line 30: Refactor note: This function, which does not require a reference to the leverage keeper as it uses its inputs in isolation, has been move to leverage types for more efficient testing. (It doesn't require a test keeper or suite).

I plan to take a similar approach for other computations (#1168) in other PRs

x/leverage/types/leverage.go
```diff
@@ -0,0 +1,137 @@
...
+	// to burn. This will round down identically to MsgWithdraw, favoring the module
+	// over the liquidator.
+	tokenReward = collateralBurn.ToDec().Mul(uTokenExchangeRate).TruncateInt()
+
+	return tokenRepay, collateralBurn, tokenReward
+}
+
+// ComputeCloseFactor uses a borrower's borrowed value and liquidation threshold and
+// some leverage module parameters to derive a dynamic close factor for liquidation.
+func ComputeCloseFactor(
```
toteki, line 98: Same as above - this function (formerly called `LiquidationParams`) was moved from `keeper` to `types`.

Also, it now computes only `CloseFactor`, instead of also passing through `LiquidationIncentive` unmodified.

x/leverage/types/leverage.go
```diff
@@ -0,0 +1,137 @@
...
+	// can be considered in two scenarios:
+	// 1) If collateral was the limiting factor above, then it will have already been
+	// an integer amount and truncating is a no-op.
+	// 2) If collateral was not the limiting factor, then there will be a non-dust
+	// quantity left over anyway.
+
+	// Finally, the base token reward amount is derived directly from the collateral
+	// to burn. This will round down identically to MsgWithdraw, favoring the module
+	// over the liquidator.
+	tokenReward = collateralBurn.ToDec().Mul(uTokenExchangeRate).TruncateInt()
```
toteki, line 91: Detailed reasoning has been included for why each quantity is rounded up or down and in what order. See test cases for confirmation.

toteki (commented):
Finally ready again - a large amount of test cases were added after the refactoring out of math stuff.

x/leverage/types/leverage.go
toteki: Moved back to original location - functionality improvement is retained.

x/leverage/types/leverage.go
toteki: Moved back to original location - functionality improvement is retained.

toteki:
@robert-zaremba Can you re-review here? I don't see anything else in it that requires changes.

robert-zaremba (changes requested):
Part 1. Will finish tomorrow.

- [ ] need to validate (in `ValidateBasic`) that `MsgLiquidate.RewardDenom` is not a uToken.... On the other hand, we should support `uToken` in case liquidator will like to keep `uTokens` - but that's probably a case for another task. Please confirm.

toteki:
> need to validate (in ValidateBasic) that MsgLiquidate.RewardDenom is not a uToken

In the current code, this is handled in `keeper.go`

```
if err := k.validateAcceptedDenom(ctx, rewardDenom); err != nil {
		return sdk.Coin{}, sdk.Coin{}, sdk.Coin{}, err
	}
```

After #1202 both options will be valid (but lead to different outcomes)

x/leverage/keeper/borrows.go
```diff
@@ -120,18 +122,22 @@ func (k Keeper) CalculateLiquidationThreshold(ctx sdk.Context, collateral sdk.Co
...
+		// ignore blacklisted tokens
+		if !ts.Blacklist {
+			// get USD value of base assets
+			v, err := k.TokenValue(ctx, baseAsset)
+			if err != nil {
+				return sdk.ZeroDec(), err
+			}
+
+			// add each collateral coin's weighted value to liquidation threshold
+			totalThreshold = totalThreshold.Add(v.Mul(ts.LiquidationThreshold))
```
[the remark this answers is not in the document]
toteki: `BorrowLimit` (which uses token's collateral weight) and `LiquidationThreshold` (which uses token's liquidation threshold) have a slight safety buffer between them.

Liquidation threshold is always higher - see Token.Validate

x/leverage/client/cli/tx.go
```diff
@@ -243,7 +243,7 @@ func GetCmdRepay() *cobra.Command {
 // transaction with a MsgLiquidate message.
 func GetCmdLiquidate() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "liquidate [liquidator] [borrower] [amount] [reward]",
+		Use:   "liquidate [liquidator] [borrower] [amount] [reward-denom]",
```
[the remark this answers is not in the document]
toteki: If we do that here, we should consider doing it for the other leverage transactions, in which case I'd prefer a separate issue / PR.

x/leverage/keeper/filter.go
```diff
@@ -0,0 +1,27 @@
+package keeper
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// filterCoins returns the subset of an sdk.Coins that meet a given condition
+func (k Keeper) filterCoins(ctx sdk.Context, coins sdk.Coins, accept func(sdk.Coin) bool) sdk.Coins {
```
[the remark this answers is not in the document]
toteki: relic of an initial implementation - removed now

proto/umee/leverage/v1/tx.proto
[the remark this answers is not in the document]
toteki: Liquidation logic is independent of final state (except for the collateral == 0 check for bad debt flagging) so that isn't a problem.

In particular, the test case labeled `repayLimited` covers this scenario in `liquidate_test.go`

x/leverage/client/cli/tx.go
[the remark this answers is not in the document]
robert-zaremba: OK. I will create an issue for that.
FYI: at Regen we were discussing it and in general we came to a conclusions that having one way to provide the sender is more clear and less confusing.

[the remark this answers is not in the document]
robert-zaremba: https://github.com/umee-network/umee/issues/1205

x/leverage/client/cli/tx.go
```diff
@@ -243,7 +243,7 @@ func GetCmdRepay() *cobra.Command {
...
 		Args:  cobra.ExactArgs(4),
 		Short: "Liquidate a specified amount of a borrower's debt for a chosen reward denomination",
```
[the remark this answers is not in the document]
toteki: Let's do this in a different PR - potentially after the `--from` changes. I need this to merge so I can start other PRs that would modify the same files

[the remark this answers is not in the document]
robert-zaremba: Why? If you are modifying it, then let's add a simple example in the long description. It's 2  lines of string.

x/leverage/keeper/math_test.go
```diff
@@ -1,10 +1,12 @@
-package keeper
+package keeper_test
```
robert-zaremba, line 1: These are unit tests, so no need to change the test package type.

x/leverage/keeper/liquidate.go
```diff
@@ -0,0 +1,216 @@
...
+) (tokenRepay sdk.Coin, collateralBurn sdk.Coin, tokenReward sdk.Coin, err error) {
+	repayDenom := maxRepay.Denom
+	collateralDenom := k.FromTokenToUTokenDenom(ctx, rewardDenom)
+
+	// get relevant liquidator, borrower, and module balances
+	borrowerCollateral := k.GetBorrowerCollateral(ctx, targetAddr)
+	totalBorrowed := k.GetBorrowerBorrows(ctx, targetAddr)
+	availableRepay := k.bankKeeper.SpendableCoins(ctx, liquidatorAddr).AmountOf(repayDenom)
+
+	// calculate borrower health in USD values
```
robert-zaremba, line 27:
```suggestion
	// calculate total borrowed value in USD
```

x/leverage/keeper/liquidate.go
```diff
@@ -0,0 +1,216 @@
...
+	// to burn. This will round down identically to MsgWithdraw, favoring the module
+	// over the liquidator.
+	tokenReward = collateralBurn.ToDec().Mul(uTokenExchangeRate).TruncateInt()
+
+	return tokenRepay, collateralBurn, tokenReward
+}
+
+// ComputeCloseFactor uses a borrower's borrowed value and liquidation threshold and
+// some leverage module parameters to derive a dynamic close factor for liquidation.
+func ComputeCloseFactor(
```
robert-zaremba, line 177: let's explain in the doc string what is a close factor in few words.

x/leverage/client/cli/tx.go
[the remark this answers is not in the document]
toteki: It's added now - but expending CLI hints was outside the scope of this PR and `Long` will need to be edited a bit when the `--from` changes take effect.

x/leverage/keeper/liquidate.go
```diff
@@ -0,0 +1,216 @@
...
+func ComputeLiquidation(
+	availableRepay sdk.Int,
+	availableCollateral sdk.Int,
+	availableReward sdk.Int,
+	repayTokenPrice sdk.Dec,
+	rewardTokenPrice sdk.Dec,
+	uTokenExchangeRate sdk.Dec,
+	liquidationIncentive sdk.Dec,
+	closeFactor sdk.Dec,
+	borrowedValue sdk.Dec,
```
robert-zaremba, line 108:
```suggestion
	availableRepay,
	availableCollateral,
	availableReward sdk.Int,
	repayTokenPrice,
	rewardTokenPrice,
	uTokenExchangeRate,
	liquidationIncentive,
	closeFactor,
	borrowedValue sdk.Dec,
```

robert-zaremba (approved):
pre-approving. Left few comments.
Importantly, let's verify if our interest rate logic is applied correctly. In the `setBorrow` and `getBorrow` I don't see how time ingredient is applied. Interest rates are defined over year, so need to do few checks:
* to not apply the interest rate in the same block twice
* add time component to set/get Borrow.

x/leverage/client/cli/tx.go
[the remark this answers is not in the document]
robert-zaremba: Sure, thanks. I've created already issue to consolidate `--from`.

x/leverage/keeper/keeper.go
```diff
@@ -346,226 +346,83 @@ func (k Keeper) Decollateralize(ctx sdk.Context, borrowerAddr sdk.AccAddress, co
...
-			// repaid in this liquidation so they are always marked as bad debt.
-			if coin.Denom != repayment.Denom || owed.IsPositive() {
-				if err := k.setBadDebtAddress(ctx, borrowerAddr, coin.Denom, true); err != nil {
-					return sdk.ZeroInt(), sdk.ZeroInt(), err
-				}
+	// detect bad debt if collateral is completely exhausted
+	if remainingCollateral.IsZero() {
+		for _, coin := range k.GetBorrowerBorrows(ctx, borrowerAddr) {
+			// set a bad debt flag for each borrowed denom
+			if err := k.setBadDebtAddress(ctx, borrowerAddr, coin.Denom, true); err != nil {
```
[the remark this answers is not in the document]
toteki: Bad debt is currently just something to iterate over, so storing `0x01` is sufficient.

We could store the `amount` - but then it would be redundant with `adjusted borrow` which also stores amount. Adjusted borrow would also continue to accrue interest while this does not. Additionally, any operation (e.g. partial repay) that updated one of the two would have to keep the other in sync, while the bad debt existed.

Overall, it's simpler to avoid storing amount here.

toteki:
> verify if our interest rate logic is applied correctly.

Overall, try to keep discussions within the bounds of the PR.

The interest rate logic you're looking for is in `EndBlock`. This only happens once per block, and increases `InterestScalar` for each denom. The math and functionality has been audited during the Runtime Verification audit.

x/leverage/keeper/math_test.go
toteki: Noted, I'll revert it

x/leverage/keeper/liquidate.go
toteki: keeping as-is (the comment describes the next three steps: `borrowed value`, `liquidation threshold`, and check liquidation eligibility - which together comprise the borrower's health in this context)

x/leverage/keeper/liquidate.go
toteki: updated

robert-zaremba:
> Overall, try to keep discussions within the bounds of the PR.

This PR uses get/set Borrow. So it make sense to look there and eventually could be fixed in other PR if needed.

x/leverage/keeper/math_test.go
robert-zaremba: thanks

x/leverage/keeper/liquidate.go
robert-zaremba: Personally I found it a bit confusing, because we don't have a clear health metric (eg, I would imagine health as a number between 0-1, 1=perfect). What we really calculate is the total borrowed value in USD.

Issue #1107, opened 2022-07-05
Use uint32 instead of sdk.Dec


robert-zaremba:
## Summary

For % values, we should use `uint32` rather than `sdk.Dec`:
* it's faster for serialization
* it's smaller
* it's faster to cmpute
* users tend to think 15% rather than 0.15

Ref: https://github.com/umee-network/umee/pull/1096#discussion_r913277290

---

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

toteki:
Seems any storage benefits we get from using `uint32` per token will be too small to notice - tokens aren't that numerous.

Also we'd need to write `uint32->sdk.Dec` getter functions as well, i.e. if `Token.CollateralWeight` is `uint32` then we need to write `Token.GetCollateralWeight() sdk.Dec` so we can multiply collateral prices in computations.

Overall, I don't think the storage changes are worth the extra boilerplate. Closing for now.

robert-zaremba:
Let's don't close the issue before discussing it. 
1. In places we use sdk.Int we don't need `uint32->sdk.Dec` (in fact that one is super simple)
2. there are computational benefits.

robert-zaremba:
I've created https://github.com/umee-network/umee/pull/1149 to demonstrate how this can be done.
This doesn't add more complexity - it's hidden in helper math basis point math functions. For `LiquidationIncentive` we don't even need to convert to `sdk.Dec` and we can stay in `sdk.Int` - so operations will be faster (no need to multiply by big number: 1e18).

Exactly same thing can be applied for `CollateralWeight`.

Convertion to Dec is easy and fast (as fast as converting sdk.Int) - I've created a helper method to that.

robert-zaremba:
> MaxBorrowAPY will likely be above 100% in many cases.

We can lift that limitation, and change to 100'000'000% (100k times) which is more than enough.

robert-zaremba:
We can easily convert all params and token params from x/leverage to uint. We will need to continue using sdk.Dec for some internal operations, before a bigger Dec refactoring, without changing the state.

I think the bigger question is if we want to use per-cent, per-mil or basis point.

toteki:
How about for a sequence of events:
- First merge new params (proto) as `sdk.Dec` in #1159
- Then a PR with all the relevant `sdk.Dec` -> `uint32` on proto.

[...]

PR #1149, opened 2022-07-21
feat!: convert LiquidationIncentive from Dec to uint32

robert-zaremba:
## Description

Following discussion about using fixed point integers to represent small decimal values (percentage, basis points), this PR demonstrates how we can do it without adding more complexity.

See https://github.com/umee-network/umee/issues/1107 for discussion and motivation.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


robert-zaremba:
Another thing to consider is the size of small decimal numbers. Is basis points too precise or per-mile will need good enough?

toteki:
Percent should be okay in my opinion, but it might help to have the extra precision in case it's needed. Make sure we're capable of handling values like `MaxBorrowRate: 120%` for APY though if we're including those fields.

robert-zaremba:
I have added a new type for package: `BP` to represent operations beyond "1" (10'000bp)

robert-zaremba:
I will add MaxBorrowRate to this PR to cover two examples and have better taste of how this will work.

robert-zaremba:
I've added `MaxBorrowRate` to this PR. In the keeper code there is almost no change.

RafilxTenfen:
Changing everything from `sdk.Dec` to `uint32` would be API breaking right? 
Would also affect the structures on the rust side

robert-zaremba:
> Changing everything from sdk.Dec to uint32 would be API breaking right?

These are not parameters to the messages. Except the general gov param update, for which we don't have any additional interface yet. `Token` is part of the `QueryRegisteredTokensResponse`  - so that one will indeed be the only change seeing externally.

adamewozniak (approved):
utack 👍 just has to pass e2e

toteki (changes requested):
Let's do these params all at once, and without the custom math types at first (so uint32 only in proto)

robert-zaremba:
> Let's do these params all at once, and without the custom math types at first (so uint32 only in proto)

Why? I don't see the reason why not using the custom type. Custom type gives us more control in fact and provides the essence of the solution. 
Otherwise it's like using string without using `sdk.Dec`. Doesn't make much sense.

toteki:
The change allows us to:
- Get `uint32` proto change into v3
- Save all math conversions and custom typing for v4+

This is acting under the assumption that we are not removing `sdk.Dec` from internal calculations in v3

robert-zaremba:
I see only disadvantages of using uint32 directly (`uint32` is ambiguous) rather than adding a proper type for that. The work with types is done and it's simple. If you want to move it to v4 then please provide a good argument.

robert-zaremba:
BTW, the helper functions are short and provide more convenience, to deal with new type which adds more reasoning to the code.

toteki:
If we reduce the complexity slightly - specifically by eliminating `FixedBP`, `Mul`, and `Quo` but keeping `BP` and its `ToDec()`, would that be acceptable for current purposes?

In that case we'd keep the class and proto clarity and the most common helper method, but not introduce the 4-precision computations.

robert-zaremba:
+ `Mul` and `Quo` is a helper function, and we are using it.
+ `FixedBP` I don't have a strong affiliation to it. I think it's nice to have, but if you really don't want it then I can remove it.

proto/umee/leverage/v1/leverage.proto
```diff
@@ -111,10 +110,10 @@ message Token {
 
   // Liquidation Incentive determines the portion of bonus collateral of
   // a token type liquidators receive as a liquidation reward.
-  string liquidation_incentive = 9 [
-    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
-    (gogoproto.nullable)   = false,
-    (gogoproto.moretags)   = "yaml:\"liquidation_incentive\""
+  // Value is in basis points; allowed values are in [0, 10000] range.
+  uint32 liquidation_incentive = 9 [
```
toteki, line 114: We don't generally comment range information in our proto files (even though `Token.Validate` range restricts most fields.)

x/leverage/types/token.go
```diff
@@ -6,11 +6,13 @@ import (
...
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/umee-network/umee/v2/math/bpmath"
 )
 
 const (
 	// UTokenPrefix defines the uToken denomination prefix for all uToken types.
-	UTokenPrefix = "u/"
+	UTokenPrefix       = "u/"
+	maxBorrowRateLimit = bpmath.ONE * 10_000
```
toteki, line 15: This limit isn't necessary to hard code

x/leverage/keeper/keeper.go
```diff
@@ -432,7 +433,7 @@ func (k Keeper) LiquidateBorrow(
 	}
 
 	// apply liquidation incentive
-	reward.Amount = reward.Amount.ToDec().Mul(sdk.OneDec().Add(liquidationIncentive)).TruncateInt()
+	reward.Amount = bpmath.Mul(reward.Amount, liquidationIncentive).Add(reward.Amount)
```
toteki, line 436: This is the only use of `Mul` (or `Quo`) I could find in the current diff

Interestingly enough, the original line had to be removed in my liquidation PR, because even the small `sdk.Dec` early rounding was contributing to dust issues. Now liquidation incentive is multiplied to a Dec without rounding back to Int.

I think it is possible to avoid 4-precision `Mul` and `Quo` altogether

toteki:
Preferences:
- Convert `FixedBP` -> `BP`
- See if we can avoid `Mul` and `Quo` on `BP`

I did the merge conflict cleanup on this, so it should be ready after that. We can merge it before #1118 if you're worried about merge conflicts.

toteki:
I can also handle conflicts here if #1118 merges first - as far as I know it's being blocked not because it has problems but just to keep this one first.

x/leverage/types/token.go
robert-zaremba: It's `1`. So, should be renamed to ONE.

app/test_helpers.go
```diff
@@ -115,9 +115,9 @@ func IntegrationTestNetworkConfig() network.Config {
 		LiquidationThreshold:   sdk.MustNewDecFromStr("0.05"),
 		BaseBorrowRate:         sdk.MustNewDecFromStr("0.02"),
 		KinkBorrowRate:         sdk.MustNewDecFromStr("0.2"),
-		MaxBorrowRate:          sdk.MustNewDecFromStr("1.5"),
+		MaxBorrowRate:          1_5000,
```
robert-zaremba, line 118: for bigger numbers we can use `_` to separate "decimal". Note: we can't write `0_100` (for 0.1), because go would interpret it in octal notation.

Issue #1167, opened 2022-07-25
price-feeder: naming strings from map keys


RafilxTenfen:
## Problem Definition

- Facilitate the code readability 

## Proposal
- Create types for strings used as keys in map like `type ProviderName string` to be used everywhere an `providerName` is used as the key of a map 
`providerPairs map[string][]types.CurrencyPai` -> `providerPairs map[ProviderName][]types.CurrencyPair`

---

## For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

RafilxTenfen:
Provider Name is at #1175 

but I also think we should have those type for 
`TokenSymbol`, `BaseSymbol`, `QuoteSymbol` == "BTC", "UMEE"
or call it all `Symbol` and at the providers move the `Symbol` property to `PairSymbol` == "BTC_USDT" 

@adamewozniak Suggestions?

PR #1175, opened 2022-07-27
chore: Add ProviderName type

RafilxTenfen:
## Description

- Type `ProviderName` created to help identify structures instead of using `map[string]` it changed to `map[ProviderName]`

helps to close: #1167 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


price-feeder/config/config.go
```diff
@@ -19,15 +20,6 @@ const (
...
 	defaultProviderTimeout = 100 * time.Millisecond
-
-	ProviderKraken   = "kraken"
-	ProviderBinance  = "binance"
-	ProviderOsmosis  = "osmosis"
-	ProviderHuobi    = "huobi"
-	ProviderOkx      = "okx"
-	ProviderGate     = "gate"
-	ProviderCoinbase = "coinbase"
-	ProviderMock     = "mock"
```
robert-zaremba, line 30: this is definitely a config, shouldn't be moved to `types`.

price-feeder/oracle/provider/binance.go
```diff
@@ -110,7 +110,7 @@ func NewBinanceProvider(
 	provider := &BinanceProvider{
 		wsURL:           wsURL,
 		wsClient:        wsConn,
-		logger:          logger.With().Str("provider", "binance").Logger(),
+		logger:          logger.With().Str("provider", types.ProviderBinance.String()).Logger(),
```
robert-zaremba, line 113:
```suggestion
		logger:          logger.With().Str("provider", string(types.ProviderBinance)).Logger(),
```

robert-zaremba: this should be more efficient.

robert-zaremba (commented):
I don't see the provider list nor `Provider` type to be stored in `types` package. 
- [ ] Let's move it back to `config` or we can create a new package, `provider` to store both the type and list of providers.
- [ ] `Provider` type should be stored next to the list of providers, not in the `types` package. In the future we will refactor the types package.

RafilxTenfen:
> * [ ]  Let's move it back to `config` or we can create a new package, `provider` to store both the type and list of providers.
- Putting the structure `ProviderName` inside the config package would generate an import cycle we could create a new package `provider` in `price-feeder/provider`, but this way we would have 2 `provider` packages at 2 different locations, which doesn't sound good to me

> * [ ]  `Provider` type should be stored next to the list of providers, not in the `types` package. In the future we will refactor the types package.
- The `Provider` types is already inside `price-feeder/oracle/provider` 

Any suggestions @robert-zaremba ?

price-feeder/oracle/provider/binance.go
RafilxTenfen: done at https://github.com/umee-network/umee/pull/1175/commits/5dfae8bf1207f167bb90a5755bc234c41c243264

price-feeder/oracle/convert_test.go
```diff
@@ -111,7 +110,7 @@ func TestConvertCandlesToUSD(t *testing.T) {
 	require.Equal(
 		t,
 		atomPrice.Mul(usdtPrice),
-		convertedCandles["binance"]["ATOM"][0].Price,
+		convertedCandles[types.ProviderBinance]["ATOM"][0].Price,
```
adamewozniak, line 113: Ty 💯

price-feeder/oracle/types/provider.go
```diff
@@ -0,0 +1,17 @@
...
+	ProviderOkx      ProviderName = "okx"
+	ProviderGate     ProviderName = "gate"
+	ProviderCoinbase ProviderName = "coinbase"
+	ProviderMock     ProviderName = "mock"
+)
+
+// ProviderName name of an oracle provider. Usually it is an exchange
+// but this can be any provider name that can give token prices
+// examples.: "binance", "osmosis", "kraken".
+type ProviderName string
```
adamewozniak, line 17: Might be a bit cleaner to add a .String() function and call it instead of using `string(p)` everywhere - wdyt?
```
func (p ProviderName) String() string {
	return string(p)
}
```

adamewozniak (commented):
just a small suggestion 😄

price-feeder/oracle/types/provider.go
RafilxTenfen: I had done that way before - https://github.com/umee-network/umee/pull/1175#discussion_r931842536 
Robert commented that it would be more efficient `string(p)` 

In fact, it is, but it doesn't change that much

RafilxTenfen: In my opinion using `func (p ProviderName) String()` would be better for keeping methods in the same place the type is defined

adamewozniak: If we ever decide to change the `String()` method (make it all caps, etc) it would be cleaner to do so as well. Imo cleaner code > efficiency here, but either way doesn't change much atm.

adamewozniak (approved):

price-feeder/oracle/types/provider.go
robert-zaremba: `string(p)` is a type conversion. `String` is a method. Two different things.

robert-zaremba:
>  The Provider types is already inside price-feeder/oracle/provider

@RafilxTenfen then let's keep both the `Provider` type and the list in the existing `provider` package.

price-feeder/oracle/types/provider.go
RafilxTenfen: > @RafilxTenfen then let's keep both the `Provider` type and the list in the existing `provider` package.
import cycle again... 
the good thing about being in the `types` package is that the `types` itself doesn't import any other package
@robert-zaremba

RafilxTenfen:
> @RafilxTenfen then let's keep both the `Provider` type and the list in the existing `provider` package.

import cycle again... 
the good thing about being in the `types` package is that the `types` itself doesn't import any other package
@robert-zaremba

robert-zaremba:
Something must be wrong. Why  moveing the list to the provider package creates the import cycle? Can you write down the cycle here?

RafilxTenfen:
> Something must be wrong. Why moveing the list to the provider package creates the import cycle? Can you write down the cycle here?

- The list of providers and the new type are inside `price-feeder/oracle/provider` 

--> Installing...
CGO_ENABLED=0 go install -mod=readonly -ldflags '-X github.com/umee-network/umee/price-feeder/cmd.Version=-3cc05765e399057b3b6ff8b27a9f2d8fdeff468f -X github.com/umee-network/umee/price-feeder/cmd.Commit=3cc05765e399057b3b6ff8b27a9f2d8fdeff468f' ./...
package github.com/umee-network/umee/price-feeder
	imports github.com/umee-network/umee/price-feeder/cmd
	imports github.com/umee-network/umee/price-feeder/config
	imports github.com/umee-network/umee/price-feeder/oracle/provider
	imports github.com/umee-network/umee/price-feeder/config: import cycle not allowed
package github.com/umee-network/umee/price-feeder
	imports github.com/umee-network/umee/price-feeder/cmd
	imports github.com/umee-network/umee/price-feeder/config
	imports github.com/umee-network/umee/price-feeder/oracle/provider
	imports github.com/umee-network/umee/price-feeder/config: import cycle not allowed

RafilxTenfen:
@robert-zaremba can you check this again?

robert-zaremba (commented):
@RafilxTenfen it seems that the import cycle was caused by the importing the `config.ProviderEndpoint` in provider package. If we move that type to the `provider` package then we solve the import cycle, and we will store all provider types in the `provider` package.
Note: when moving `ProviderEndpoint` to provider package, we should remove `Provider` prefix (to avoid name clash: `provider.ProviderEndpoint`).

RafilxTenfen:
> @RafilxTenfen it seems that the import cycle was caused by the importing the `config.ProviderEndpoint` in provider package. If we move that type to the `provider` package then we solve the import cycle, and we will store all provider types in the `provider` package. Note: when moving `ProviderEndpoint` to provider package, we should remove `Provider` prefix (to avoid name clash: `provider.ProviderEndpoint`).

The struct `CurrencyPair` is also imported, do you think that should be sent to the `price-feeder/oracle` package as well?

RafilxTenfen:
This reorganization of price-feeder structs could be done in another PR after we discuss and plan an issue specific to that 
Since @adamewozniak already approve this and if you agree @robert-zaremba  I will merge it

robert-zaremba:
> The struct CurrencyPair is also imported, do you think that should be sent to the price-feeder/oracle package as well?

Yes. `CurrencyPair` is in the domain of a provider. So let's put it there.

Issue #957, opened 2022-05-25
Add Swagger to Umeed


brianosaurus:
## Summary

Turn on and customize swagger endpoint for Umeed

## Problem Definition

Swagger is a self-documenting system that is used to document cosmos nodes' protopuf APIs. Turn this on so that developers have a clearer understanding how to develop against Umeed.

## Proposal

Examine the Regen and Axelar Makefile systems to copy best-in-class swagger implementations.

----

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

robert-zaremba:
Cosmos SDK has swagger integration in a docker file.

dreamcodez:
any updates on this? @brianosaurus @robert-zaremba

robert-zaremba:
I didn't work on this task.

PR #1203, opened 2022-08-02
feat: enable swagger

RafilxTenfen:
## Description

closes: #957 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


RafilxTenfen:
@toteki @adamewozniak could you guys check if the module specifics seem correct?

app/app.go
```diff
@@ -110,6 +113,9 @@ import (
 	"github.com/umee-network/umee/v2/x/oracle"
 	oraclekeeper "github.com/umee-network/umee/v2/x/oracle/keeper"
 	oracletypes "github.com/umee-network/umee/v2/x/oracle/types"
+
+	// unnamed import of statik for swagger UI support
+	_ "github.com/umee-network/umee/v2/client/docs/statik"
```
adamewozniak, line 118: Should this be moved to tools/tools.go?

Makefile
```diff
@@ -218,3 +224,61 @@ proto-lint:
...
+UNAME_M ?= $(shell uname -m)
+
+GOPATH ?= $(shell $(GO) env GOPATH)
+GITHUBDIR := $(GOPATH)$(FS)src$(FS)github.com
+
+BUF_VERSION ?= 0.11.0
+
+TOOLS_DESTDIR  ?= $(GOPATH)/bin
+STATIK         = $(TOOLS_DESTDIR)/statik
+RUNSIM         = $(TOOLS_DESTDIR)/runsim
```
adamewozniak, line 244: What is the runsim part of this?

app/app.go
```diff
@@ -844,6 +850,22 @@ func (app *UmeeApp) RegisterAPIRoutes(apiSvr *api.Server, apiConfig config.APICo
 	// register legacy and grpc-gateway routes for all modules
 	ModuleBasics.RegisterRESTRoutes(clientCtx, apiSvr.Router)
 	ModuleBasics.RegisterGRPCGatewayRoutes(clientCtx, apiSvr.GRPCGatewayRouter)
+
+	// register swagger API from root so that other applications can override easily
```
adamewozniak, line 854:
```suggestion
```

client/docs/config.json
```diff
@@ -0,0 +1,26 @@
+{
```
adamewozniak, line 1: Could we move this into a folder `swagger` instead of `client/docs` maybe?

adamewozniak (commented):
Just a few questions - also @RafilxTenfen should we ignore the generated files here and maybe add it to `make build`, or something similar? 

These files will likely become expired if they're not generated each time queries are updated

RafilxTenfen:
> Just a few questions - also @RafilxTenfen should we ignore the generated files here and maybe add it to `make build`, or something similar?
> 
> These files will likely become expired if they're not generated each time queries are updated

Make sense, maybe we could put it together with proto-gen? since it will be modified it time the proto files change

Makefile
RafilxTenfen: that is my bad, forgot to take it off

Makefile
```diff
@@ -218,3 +224,61 @@ proto-lint:
...
+
+###############################################################################
+###                              Documentation                              ###
+###############################################################################
+
+update-swagger-docs: statik
+	$(STATIK) -src=client/docs/swagger-ui -dest=client/docs -f -m
+	@if [ -n "$(git status --porcelain)" ]; then \
+        echo "\033[91mSwagger docs are out of sync!!!\033[0m";\
+        exit 1;\
```
toteki, line 279: Are we taking the approach where the swagger docs are always gitignored, or will they be committed?

And if they're being committed - this is most of the ingredients to a CI check that we could make, which would check it they're out of sync on any PR

toteki (changes requested):
`client/docs/statik/statik.go` is so large that my `git clone` time for the repository seems to have doubled from this commit. 

Because it's all in 13 lines, I also fear that it cannot `diff` efficiently, so our repo size would increase by the full 10+ MB every time it's generated.

Thus, we need to go the `.gitignore` route

Makefile
```diff
@@ -218,3 +224,61 @@ proto-lint:
...
 	@$(DOCKER_BUF) breaking --against $(HTTPS_GIT)#branch=main
+
+###############################################################################
+###                                 Tools                                   ###
+###############################################################################
+
+PREFIX ?= /usr/local
+BIN ?= $(PREFIX)/bin
+UNAME_S ?= $(shell uname -s)
+UNAME_M ?= $(shell uname -m)
```
robert-zaremba, line 235: do we really need this? binaries should be set by the user, we should not hardcode the path here (eg /usr/local/bin...)

Makefile
```diff
@@ -218,3 +224,61 @@ proto-lint:
...
+statik: $(STATIK)
+$(STATIK):
+	@echo "Installing statik..."
+	@go install github.com/rakyll/statik@v0.1.6
+
+# Install the runsim binary
+runsim: $(RUNSIM)
+$(RUNSIM):
+	@echo "Installing runsim..."
+	@go install github.com/cosmos/tools/cmd/runsim@v1.0.0
```
robert-zaremba, line 263: I don't think runsim nor statik is needed for swagger. Statik can be used to pack static files in a go library. Not sure if this is the best solution though. I know SDK and Regen are using Statik to serve doc files directly from the app. Ideally I would serve documentation from other app, rather than bundling it into the chain. What do you think?

app/app.go
RafilxTenfen: If we move it, than the swagger will use the cosmos-sdk statik file which will display the swagger for cosmos-sdk API

client/docs/config.json
robert-zaremba: this is what other projects are using, but I agree - `docs/swagger` would be probably better place.

Makefile
```diff
@@ -218,3 +224,61 @@ proto-lint:
...
+RUNSIM         = $(TOOLS_DESTDIR)/runsim
+
+tools: tools-stamp
+tools-stamp: statik runsim
+	# Create dummy file to satisfy dependency and avoid
+	# rebuilding when this Makefile target is hit twice
+	# in a row.
+	touch $@
+
+# Install the runsim binary
```
robert-zaremba, line 253: comment is wrong

Makefile
```diff
@@ -218,3 +224,61 @@ proto-lint:
...
+statik: $(STATIK)
+$(STATIK):
+	@echo "Installing statik..."
+	@go install github.com/rakyll/statik@v0.1.6
```
robert-zaremba, line 257: why do we have this in makefile?
BTW, some of this can be moved to contrib/scripts/update-swagger-ui-statik.sh

robert-zaremba (commented):
Maybe we can check with the ecosystem if someone else has better setup rather than bundling the swagger docs into the chain app.

RafilxTenfen:
@adamewozniak @toteki @robert-zaremba 
Okay, forget the first implementation, can you guys take a look again? 
I used how Juno did to make ours

app/app.go
```diff
@@ -844,6 +848,17 @@ func (app *UmeeApp) RegisterAPIRoutes(apiSvr *api.Server, apiConfig config.APICo
...
+	if apiConfig.Swagger {
+		RegisterSwaggerAPI(apiSvr.ClientCtx, apiSvr.Router)
+	}
+}
+
+// RegisterSwaggerAPI registers swagger route with API Server
+func RegisterSwaggerAPI(_ client.Context, rtr *mux.Router) {
+	// register app's OpenAPI routes.
+	rtr.Handle("/swagger.yaml", http.FileServer(http.FS(swagger.Docs)))
+	rtr.HandleFunc("/swagger/", openapiconsole.Handler(Name, "/swagger.yaml"))
```
adamewozniak, line 861: Move this into `swagger.go` please 😄

swagger/swagger.yaml
```diff
@@ -0,0 +1,2624 @@
+swagger: '2.0'
```
adamewozniak, line 1: Does this get auto generated whenever the chain starts with proto? Or do devs have to run `proto-swagger-gen`?

adamewozniak (approved):
Couple of questions but this looks much cleaner 🥳

toteki (commented):
Doc file size (swagger.yaml) is much more manageable now (80 kB) and its format would probably allow for diff.

Remaining question is whether we should
- .gitignore it anyway (so it's only generated locally when needed, and can't be out of sync)
- commit it (and either manually or through CI, make sure it doesn't desync)

swagger/swagger.yaml
RafilxTenfen: right now, if we change the proto files we need to run `proto-swagger-gen` to update the `swagger.yaml`

app/app.go
RafilxTenfen: done at https://github.com/umee-network/umee/pull/1203/commits/d1f7c30ca273b1a4ce490a76a394a7e812df959a

RafilxTenfen:
> Doc file size (swagger.yaml) is much more manageable now (80 kB) and its format would allow for efficient diff.
> 
> Remaining question is whether we should
> 
> * .gitignore it anyway (so it's only generated locally when needed, and can't be out of sync)
> * commit it (and either manually or through CI, make sure it doesn't desync)

In my opinion, we could commit that, in the same way we commit the generated structs from proto files

adamewozniak:
> > Doc file size (swagger.yaml) is much more manageable now (80 kB) and its format would allow for efficient diff.
> > Remaining question is whether we should
> > 
> > * .gitignore it anyway (so it's only generated locally when needed, and can't be out of sync)
> > * commit it (and either manually or through CI, make sure it doesn't desync)
> 
> In my opinion, we could commit that, in the same way we commit the generated structs from proto files

Maybe we can add a CI task where we check to see if swagger is out of date?

robert-zaremba:
>  We could hash the existing swagger file & then compare that hash to a newly generated file.

Or check if proto was modified: whenever proto was modified, swagger should be checked if it should be modified as well.

swagger/proto-config-gen.json
```diff
@@ -0,0 +1,26 @@
+{
+  "swagger": "2.0",
+  "info": {
+    "title": "Umee SDK - gRPC Gateway docs",
+    "description": "A REST interface for state queries",
+    "version": "1.0.0"
+  },
+  "apis": [
+    {
+      "url": "./tmp-swagger-gen/umee/leverage/v1/query.swagger.json",
```
robert-zaremba, line 10: why `tmp` prefix? It looks weird.

RafilxTenfen: it is just the name of the temp dir `/tmp-swagger-gen` used in `contrib/scripts/protoc-swagger-gen.sh` to generate the `./swagger/swagger.yaml`

robert-zaremba (approved):
long term we should consider serving swagger docs not from the chain app.

Issue #1227, opened 2022-08-09
Limit token supply (token.MaxSupply)


robert-zaremba:
## Summary

We want to apply a slow run for the leverage module release and apply additional protection rules for supplying tokens.

## Problem Definition

A common bridge exploit is to be able to mint unlimited amount of bridged tokens to deposit into a lending protocol and then drain the pools by lending other tokens.

## Proposal

We will add a token supply limit to token registry. This will be a required parameter.
Umee governance will be responsible to adjust it during the runtime. Initial parameters will be proposed by Umee Ops Team.

- [x] Proto update @robert-zaremba https://github.com/umee-network/umee/pull/1228
- [ ] Implementation + tests @toteki 
- [ ] Define initial parameters in default genesis @ Ming
- [ ] Set initial parameters for mainnet in the v1->v3 migration logic.

toteki:
I believe we have MaxTokenSupply specced out to be a flat amount rather than a % share, so it wouldn't be an issue here. 

But with `MaxCollateralShare` that bootstrap issue will appear. I think for certain safe assets (e.g. `ATOM`) the max token supply and max collateral share can both be unlimited (stored as `0`).

robert-zaremba:
`MaxTokenSupply` is speced as a share, so 0-100%, or more precisely `0-1`:

  // Max Supply Utilization specifies the maximum supply utilization a token is
  // allowed to reach as a direct result of user borrowing. New borrows are not allowed when
  // the supply utilization is above `max_supply_utilization`.
  // Valid values: 0-1.

robert-zaremba:
I think we should change `MaxTokenSupply` to a share as well (0-100%).... Otherwise we will need to more often adjust when the total supply will grow / shrink.
cc: Ming

robert-zaremba:
Of course the % share value makes only sense when the value is expressed in a common denominator, eg USD.

PR #1228, opened 2022-08-09
feat(leverage): token supply limit

robert-zaremba:
## Description

Ref: #1227

NOTE: I decided to use `Uint` instead of `Int`. However it seams it has a problem with yaml serialization (one test failing - when using `Int` it was not failing). Maybe it's fixed in 0.46 . Will check it out.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


robert-zaremba:
OK, I see the reason why `Uint` doesn't work. Basically the type doesn't implement `MarshalYAML`. Falling back to `Int`.

robert-zaremba:
I'm adding YAML support in SDK, we can change to Uint once this will be merged and backported to 0.46

x/leverage/types/token.go
```diff
@@ -96,6 +96,10 @@ func (t Token) Validate() error {
 		return sdkerrors.ErrInvalidRequest.Wrap("Token.MinCollateralLiquidity be between 0 and 1")
 	}
 
+	if !t.MaxSupply.IsPositive() {
+		return sdkerrors.ErrInvalidRequest.Wrap("Token.MaxSupply must be positive")
+	}
```
toteki, line 101: This rule would make for a more clear token registry.

```suggestion
	if !t.EnableMsgSupply && MaxSupply.IsPositive() {
		return sdkerrors.ErrInvalidRequest.Wrap("MaxSupply must be zero when EnableMsgSupply is false")
	}
	if t.MaxSupply.IsNegative() {
		return sdkerrors.ErrInvalidRequest.Wrap("Token.MaxSupply must not be negative")
	}
```

That way, any token with supply disabled has an accurate `MaxSupply` of `0` (I fear that a nonzero max supply, when supply is disabled, would be misleading).

robert-zaremba: OK, however we should use `MaxSupply.IsZero()` instead of `MaxSupply.IsPositive()`

robert-zaremba: Also it will change the proto comment - currently I set in proto that zero is not a valid number. Will update both.

robert-zaremba:
After making a relationship with `EnableMsgSupply` tests starts to fail... so a bit more work :/

x/leverage/fixtures/token.go
```diff
@@ -0,0 +1,31 @@
+package fixtures
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/umee-network/umee/v2/x/leverage/types"
+)
+
+// Token returns a valid token
+func Token(base, symbol, reserveFactor string) types.Token {
```
toteki, line 10: How about `types.ExampleToken()` instead of `fixtures.Token(base,symbol,reserveFactor)`? And have it return `uumee`/`UMEE` with sensible values.

That way
- We don't need a new folder/package
- We cover most of the repeated code in the test cases
- Reserve factor doesn't get special treatment (since any of the parameters can be customized after initializing)

toteki:
Since it looks like we're going to want the option for unlimited supply, how about we interpret `MaxTokenSupply = 0` + `EnableMsgSupply = true` as unlimited supply? Meanwhile, we can remove the requirement of "if supply disabled, max supply is 0"

x/leverage/fixtures/token.go
robert-zaremba: 1. We should not mix test domain with API domain. This data is part of the test domain, hence should be in a separate package
2. the package is called `fixtures`, so we don't need `Example` prefix.
3. I used `base, symbol, reserveFactor` because it was used like that in other test files (see tests).

robert-zaremba: I can check if we can remove `reserveFactor` argument.

robert-zaremba:
> how about we interpret MaxTokenSupply = 0 + EnableMsgSupply = true as unlimited supply?

Yes, I had the same idea. Now once Ming confirmed I will update this.

robert-zaremba:
I've updated the meaning of `max_supply=0`

proto/umee/leverage/v1/leverage.proto
```diff
@@ -126,7 +126,8 @@ message Token {
   uint32 exponent = 11 [(gogoproto.moretags) = "yaml:\"exponent\""];
 
   // Enable Msg Supply allows supplying for lending or collateral using this
-  // token. Note that withdrawing is always enabled. Disabling supplying would
+  // token. `false` means that a token is not valid for any new supply.
```
toteki, line 129:
```suggestion
  // token. `false` means that a token can no longer be supplied.
```

starport.ci.yml
```diff
@@ -26,6 +26,8 @@ genesis:
           max_collateral_share: "1.00000000000000000"
           max_supply_utilization: "1.00000000000000000"
           min_collateral_liquidity: "0.000000000000000000"
+          min_collateral_liquidity: "0.000000000000000000"
+          max_supply: "100000000000000000000"
```
toteki, line 30:
```suggestion
          min_collateral_liquidity: "0.000000000000000000"
          max_supply: "100000000000000000000"
```

x/leverage/gov_handler_test.go
```diff
@@ -89,6 +67,6 @@ func TestUpdateRegistryProposalHandler(t *testing.T) {
 
 		token, err := k.GetTokenSettings(ctx, "uosmo")
 		require.NoError(t, err)
-		require.Equal(t, "0.300000000000000000", token.ReserveFactor.String())
+		require.Equal(t, "0.200000000000000000", token.ReserveFactor.String())
```
toteki, line 70: This test is for a proposal modifying the reserve factor of an existing token from 0.2 to 0.3. Making the expected value to 0.2 should not be done, because a failure to update the registry would be misinterpreted as a success.

x/leverage/types/token_test.go
```diff
@@ -204,6 +191,18 @@ func TestToken_Validate(t *testing.T) {
...
 		},
+		"invalid max supply (negative)": {
+			input:     invalidMaxSupply1,
+			expectErr: true,
+		},
+		"valid max supply (msgsend=false)": {
+			input:     validMaxSupply1,
+			expectErr: false,
+		},
+		"valid max supply (msgsend=true)": {
```
toteki, line 202:
```suggestion
		"valid max supply (enable_msg_supply=false)": {
			input:     validMaxSupply1,
			expectErr: false,
		},
		"valid max supply (enable_msg_supply=true)": {
```

x/leverage/gov_handler_test.go
robert-zaremba: thanks for spotting

robert-zaremba: updated

toteki (approved):

PR #1239, opened 2022-08-12
feat: add release tagging scheme for experimental releases

dreamcodez:
## Description

add release tagging scheme for experimental releases


UmeeAndrey (approved):

adamewozniak:
imo we should discuss with the team whether or not we'd like to release alpha binaries before merging this  💯

robert-zaremba:
I second @adamewozniak . I don't see the reason why we should create a noise on github release page. `-alpa` / `-experimental` releases are not official. Moreover with go.mode links, this won't work.

adamewozniak (changes requested):
gonna block this since we've agreed to only have RCs / official RCs in our meeting today

robert-zaremba:
Then let's close this PR. Good that we found a consensus.

Issue #1195, opened 2022-08-01
Price-feeder: (HAL-07) HTML ESCAPING NOT IMPLEMENTED


RafilxTenfen:
## Summary
It was found that Write is being used to generate HTTP responses, instead of using the html/template package that handles HTML and other encodings more safely.

## Proposal
- It is recommended to use the html/template package when returning data
back during an HTTP response.
---

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1243, opened 2022-08-16
chore: WIP HAL-07 use the html/template package #1195

khoerling:
- when returning data back during an HTTP response
- since we can trust `json.Marshal`, let's just check the error case

## Description

closes: #XXXX

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/price-feeder/router/v1/response.go
+++ b/price-feeder/router/v1/response.go
@@ -2,6 +2,7 @@ package v1
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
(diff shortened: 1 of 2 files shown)


khoerling:
Is there reason why we shouldn't trust `json.Marshal` to produce safe output?  Easy to add html/template there although it seems unnecessary after reading docs.

adamewozniak:
> Is there reason why we shouldn't trust `json.Marshal` to produce safe output? Easy to add html/template there although it seems unnecessary after reading docs.

I think @RafilxTenfen might be able to give more context around here, iirc he was involved with the audit

RafilxTenfen:
I think that Halborn was giving an suggestion to use [html/template](https://pkg.go.dev/html/template) and having templates for 
```go
                t := template.New("jsonMarshalError")
		t.Parse("<h1>Error {{.Code}}</h1> <a>{{.Error}}</a>")
		_ = t.Execute(w, bz)
```

also we could add HTML to the templates so they can look better... 

The "parse" part is not needed, but the idea is to avoid using `w.Write(bytes)`

RafilxTenfen (approved):
@adamewozniak Keith made a good point, that even that Halborn recommended to use `html/template` usually who is going to consume this API is like another API. So JSON is better in my opinion, what do you think?

khoerling:
We could also inspect `Accept:` header and offer html or json, for human or computer

adamewozniak:
> @adamewozniak Keith made a good point, that even that Halborn recommended to use `html/template` usually who is going to consume this API is like another API. So JSON is better in my opinion, what do you think?

I agree JSON works best here 👍

adamewozniak (approved):

Issue #1245, opened 2022-08-17
Update x/oracle & gravity bridge default params to jail after 1 day


adamewozniak:
## Summary

We should temporarily set the default jailing periods for the oracle and gravity bridge parameters to 1 day in order for our devnets to fail earlier 

1. Switch to 1 day (#1246)
2. Switch back to 1 week :^) 

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1246, opened 2022-08-17
chore: update default oracle slashing window to 1 day

adamewozniak:
## Description

Strictly for devnet purposes

addresses: #1245, plan on reverting before RC

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/x/oracle/types/params.go
+++ b/x/oracle/types/params.go
@@ -23,7 +23,7 @@ var (
 // Default parameter values
 const (
 	DefaultVotePeriod               = BlocksPerMinute / 2 // 30 seconds
-	DefaultSlashWindow              = BlocksPerWeek       // window for a week
+	DefaultSlashWindow              = BlocksPerDay        // window for a day
 	DefaultRewardDistributionWindow = BlocksPerYear       // window for a year
 )
 


toteki:
Might want to create an issue or other todo item to make sure we don't forget to set them back

adamewozniak:
> Might want to create an issue or other todo item to make sure we don't forget to set them back

I'm going to keep the original issue here open to remind myself 💯

x/oracle/types/params.go
```diff
@@ -23,7 +23,7 @@ var (
 // Default parameter values
 const (
 	DefaultVotePeriod               = BlocksPerMinute / 2 // 30 seconds
-	DefaultSlashWindow              = BlocksPerWeek       // window for a week
+	DefaultSlashWindow              = BlocksPerDay        // window for a day
```
robert-zaremba, line 26: Maybe instead of changing it here, we should use non-default genesis? The idea would be to modify genesis either:
* before running `collect-gentx` using a script (so that would require one more step when setting up a testnet)
* or adding a switch / genesis modification in migration code (`app/upgrades.go`)
WDYT?

robert-zaremba: Editing someone else comment to add response looks odd :P

adamewozniak: Wtf? I did this on the app, I think that's why 😓

robert-zaremba:
Why closing this PR, we can update it rather than closing.
If you really want to close it then you should remove the branch as well.

PR #1288, opened 2022-08-26
fix: safeguard LastInterestTime against hard forks

toteki:
## Description

This will prevent the excessive (~10^20 factor) interest accrued during our hard fork of asherah testnet, where we modified LastInterestTime in genesis.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


CHANGELOG.md
```diff
@@ -97,6 +97,7 @@ Ref: https://keepachangelog.com/en/1.0.0/
 - [987](https://github.com/umee-network/umee/pull/987) Streamline x/leverage CLI tests
 - [1012](https://github.com/umee-network/umee/pull/1012) Improve negative time elapsed error message
 - [1236](https://github.com/umee-network/umee/pull/1236) Improve leverage event fields.
+- [XXXX](https://github.com/umee-network/umee/pull/XXXX) Safeguards LastInterestTime against time reversals and unintended interest from hard forks.
```
toteki, line 100:
```suggestion
- [1288](https://github.com/umee-network/umee/pull/1288) Safeguards LastInterestTime against time reversals and unintended interest from hard forks.
```

x/leverage/keeper/interest.go
```diff
@@ -80,9 +84,19 @@ func (k Keeper) AccrueAllInterest(ctx sdk.Context) error {
...
+		// if LastInterestTime appears to tbe in the future, do nothing (besides logging) and leave
+		// LastInterestTime at its stored value. This will repeat every block until BlockTime exceeds
+		// LastInterestTime.
 		return nil
 	}
+
 	yearsElapsed := sdk.NewDec(currentTime - prevInterestTime).QuoInt64(types.SecondsPerYear)
+	if yearsElapsed.GTE(sdk.OneDec()) {
+		// this safeguards primarily against misbehaving block time or incorrectly modified genesis states
+		// which would accrue significant interest on borrows instantly. Chain will halt.
```
RafilxTenfen, line 96: Where the chain will halt?

toteki: Error in `AccrueAllInterest`, which occurs during `EndBlock`, becomes a panic (similar to when we see negative time elapsed)

RafilxTenfen:
can we backport this to 2.x.x ?

RafilxTenfen (approved):

x/leverage/keeper/interest.go
robert-zaremba: How it is possible that we can each such a long diff? The interest is accuring regularly.

x/leverage/keeper/interest.go
```diff
@@ -178,6 +192,14 @@ func (k *Keeper) SetLastInterestTime(ctx sdk.Context, interestTime int64) error
 	store := ctx.KVStore(k.storeKey)
 	timeKey := types.CreateLastInterestTimeKey()
 
+	prevTime := k.GetLastInterestTime(ctx)
+
+	if interestTime < prevTime {
```
robert-zaremba, line 197: when this is possible?

x/leverage/keeper/keeper_test.go
```diff
@@ -661,6 +661,36 @@ func (s *IntegrationTestSuite) TestAccrueZeroInterest() {
...
+	// already has 1k u/umee for collateral.
+	addr, _ := s.initBorrowScenario()
+
+	// user borrows 40 umee
+	err := s.app.LeverageKeeper.Borrow(s.ctx, addr, sdk.NewInt64Coin(umeeapp.BondDenom, 40000000))
+	s.Require().NoError(err)
+
+	// Setting last interest time to a negative value is not allowed
+	err = s.tk.SetLastInterestTime(s.ctx, -1)
+	s.Require().Error(err)
```
robert-zaremba, line 675: let's check errors using `ErrorContains`

robert-zaremba (commented):
Could you add a bit more comments? You are investigating cases which potentially are not possible. It's OK to make that check, but let's describe it for completness when these things can happen. Otherwise it's a bit confusing.... it's like checking `if 2+2 > 0`.

x/leverage/keeper/interest.go
toteki: It shouldn't be possible - but there's still that bug mentioned in the comments of AccrueAllInterest, where block time moves backwards. Protecting against it at the `Set` level seems to be a good redundant defense.

`// TODO fix this when tendermint solves https://github.com/tendermint/tendermint/issues/8773`

toteki:
> How it is possible that we can each such a long diff? The interest is accruing regularly.

Scenarios where time difference of a year would be possible
- If a chain halt was so long, the next block would be a year later. (This is not what I expect)
- If `BlockTime` behaves incorrectly (Shouldn't happen, but bugs are always possible and this one would be _lethal_ to our users.)
- If `LastInterestTime` in genesis file was modified during a hard fork, and the next block would see massive interest accrue on all borrows to catch up (we suspect this just happened on Asherah-2).

Curiously, the `if lastInterestTime == 0` condition in `AccrueAllInterest` should have protected us, but it didn't.

toteki:
> can we backport this to 2.x.x ?

Yeah (might take a few manual adjustments on the backport PR).

Also I need to add a gov proposal types to adjust interest scalar and set reserve amounts - those will be a separate PR but if we backport them rather than modifying genesis file, we could fix Asherah-2 with a binary swap + proposals instead of directly editing genesis. Might be faster to just edit the file though.

x/leverage/keeper/keeper_test.go
toteki: Used `ErrIs(err,types.ErrNegativeTimeElapsed)` at `67fd679`

x/leverage/keeper/interest.go
```diff
@@ -80,9 +84,19 @@ func (k Keeper) AccrueAllInterest(ctx sdk.Context) error {
 			"prev", prevInterestTime,
 		)
 
+		// if LastInterestTime appears to tbe in the future, do nothing (besides logging) and leave
```
toteki, line 87:
```suggestion
		// if LastInterestTime appears to be in the future, do nothing (besides logging) and leave
```

x/leverage/keeper/interest.go
robert-zaremba: would be good to add a comment to clarify this here. Otherwise it looks strange.

Issue #1281, opened 2022-08-25
price-feeder version


RafilxTenfen:
## Summary of Bug
```shell
$~ price_feeder version
Error: open go.mod: no such file or directory
Usage:
  price-feeder version [flags]

Flags:
      --format string   Print the version in the given format (text|json) (default "text")
  -h, --help            help for version

Global Flags:
      --log-format string   logging format; must be either json or text (default "text")
      --log-level string    logging level (default "info")

open go.mod: no such file or directory

```

A clear and concise description of what the 🐞 is.

## Version

Please provide the output of the following commands:

- `$ umeed version` main-18aaf2669e63e70297a6beefee76b1344bd7ca0d
- `$ go version` go version go1.19 linux/amd64
- `$ uname -a` Linux rafilx-pc 5.15.0-27-generic #28-Ubuntu SMP Thu Apr 14 04:55:28 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

## Steps to Reproduce

1. build price-feeder
2. move the bin to a non-go.mod location 
3. go to that location in the shell
4. run `price-feeder version`

----

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1291, opened 2022-08-26
fix: price-feeder version

rbajollari:
…on command

## Description

closes: #1281

This bug is happening because the price-feeder version command is trying to open the go.mod file of the price-feeder but if you're outside of that directory it doesn't know where that go.mod file is. My proposed solution is to add an argument to the version command for the directory of the price-feeder in your environment. Let me know if you have a better way to approach this.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


RafilxTenfen:
Hey @rbajollari, So the idea is that the `price-feeder version` doesn't need any go.mod to display it's current version 
instead when we are building the binary with `make build`, it should somewhat be able to save the current git commit / tag version in which the binary was build 

An idea how to do that it to look how umeed is doing `umeed version`

price-feeder/cmd/version.go
```diff
@@ -36,10 +36,13 @@ type versionInfo struct {
...
-		Use:   "version",
+		Use:   "version [path to price-feeder directory]",
+		Args:  cobra.ExactArgs(1),
 		Short: "Print binary version information",
-		RunE: func(cmd *cobra.Command, args []string) error {
-			modBz, err := os.ReadFile("go.mod")
+		RunE: func(cmd *cobra.Command, args []string) error {			
+			priceFeederDir := args[0]
+
+			modBz, err := os.ReadFile(priceFeederDir + "/go.mod")
```
robert-zaremba, line 45: We should not read go.mod file.

robert-zaremba (changes requested):
Please use the pattern of setting a version during the build time:
* https://github.com/umee-network/umee/blob/main/Makefile#L62
* https://github.com/umee-network/umee/blob/main/Makefile#L72

You need to adopt it in price-feeder by adding `Version` variable to `price-feeder/main.go` and setting it through ldflags as outlined above.

price-feeder/cmd/version.go
adamewozniak: Agreed - @rbajollari we should be able to use this command with just a binary. So let’s make it not error out when there’s no go.mod

rbajollari: Got it!

price-feeder/Makefile
```diff
@@ -22,7 +23,8 @@ endif
 ###############################################################################
 
 ldflags = -X github.com/umee-network/umee/price-feeder/cmd.Version=$(VERSION) \
```
adamewozniak, line 25: Looks like above we have this line: 
```
  ifeq (,$(VERSION))
    VERSION := $(BRANCH)-$(COMMIT)
  endif
```

Which resolves to `-hash` when built using this makefile, because `BRANCH` is not filled out here.

Could you add the branch to the top of this file much how `umee/makefile` does it? that way it'll look like `branch-hash` 😄

rbajollari: done!

adamewozniak (approved):
Nice! Looks good to me!

adamewozniak:
@robert-zaremba please check this out when you can

RafilxTenfen (approved):
Tested ACK, just a question on umeed we have `umeed version` which just returns like `main-1f36d964a9449cb307806bef68eb90197b845274` and the `--long` flag, which it prints all the build deps, build flags used and cosmos SDK version

What do you think about following that pattern?

robert-zaremba (approved):
utACK

adamewozniak:
> Tested ACK, just a question on umeed we have `umeed version` which just returns like `main-1f36d964a9449cb307806bef68eb90197b845274` and the `--long` flag, which it prints all the build deps, build flags used and cosmos SDK version
> 
> What do you think about following that pattern?

Hmm, we could add a task to add the "--long" flag, but atm I think the contents of this PR are enough (we mostly care about commit # and cosmos version with this tool) - wdyt @RafilxTenfen ?

RafilxTenfen:
> > Tested ACK, just a question on umeed we have `umeed version` which just returns like `main-1f36d964a9449cb307806bef68eb90197b845274` and the `--long` flag, which it prints all the build deps, build flags used and cosmos SDK version
> > What do you think about following that pattern?
> 
> Hmm, we could add a task to add the "--long" flag, but atm I think the contents of this PR are enough (we mostly care about commit # and cosmos version with this tool) - wdyt @RafilxTenfen / @rbajollari ?

Agree

Issue #1242, opened 2022-08-15
x/leverage: Increase testing of msg response return values


toteki:
Currently `MsgSupply`, `MsgWithdraw`, `MsgRepay`, and `MsgLiquidate` have tests in `keeper_test.go` which check their return values against expected values.

These tests could be improved by making them case-based (and increasing the number of situations covered) as well as checking the consistency of the returned values (e.g. collateral received) against actual changes in balances, to catch any scenarios where the effects of a transaction do not match its reported returns.

PR #1300, opened 2022-08-30
refactor: improve leverage keeper tests and errors

toteki:
## Description

Except small behavior changes, this PR is limited to `x/leverage/keeper/*_test.go`

Behavior changes:
- Fixed MsgCollateralize ValidateBasic allowing nil coin
- Added uToken exchange rates >= 1 invariant
- Change validation order to return more accurate error messages for a few transactions

Main improvements:
- Switch keeper tests for each message type to table based, and greatly expanded testing of secondary effects on balances.
- Reorganize errors and test for specific error types
- Remove the cascading failures of `x/leverage/client/tests` which would make it hard to identify the root cause of a failure
- Remove `s.initBorrowScenario` in favor of more specific test setup functions (e.g. `s.borrow`)
- Check invariants after `keeper_test.go` functions
- For all expected errors, expect a typed error, e.g. `require.ErrorIs(err, types.ErrNotRegisteredToken)` 

Cosmetic improvements:
- Use underscores to separate 6 digits for coins in test files. e.g, `400_000000` when creating `400 UMEE`
- Ensure `expected` value is always on the left in `require.Equal(expected,actual)` for accurate test fail messages

closes: #1310
closes: #1242 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/leverage/client/tests/suite.go
```diff
@@ -17,8 +17,6 @@ import (
 type IntegrationTestSuite struct {
 	suite.Suite
 
-	abort bool // stop interdependent tests on the first error for clarity
-
```
toteki, line 21: This wasn't working due to a misuse of `s.Run` - now it's not needed after the fix.

x/leverage/keeper/reserves_test.go
```diff
@@ -0,0 +1,76 @@
+package keeper_test
+
```
toteki, line 2: This file (and the other completely new files, except `suite_test.go`) contains only functions which have been relocated from `keeper_test.go`

x/leverage/keeper/suite_test.go
```diff
@@ -0,0 +1,183 @@
+package keeper_test
+
```
toteki, line 2: This file contains test suite setup and helpers (some new, some relocated from `keeper_test.go`).

Now `keeper_test.go` is strictly to test the functions in `keeper.go`.

x/leverage/keeper/validate.go
```diff
@@ -9,6 +9,9 @@ import (
 // validateAcceptedDenom validates an sdk.Coin and ensures it is a registered Token
 // with Blacklisted == false
 func (k Keeper) validateAcceptedDenom(ctx sdk.Context, denom string) error {
+	if types.HasUTokenPrefix(denom) {
+		return types.ErrUToken.Wrap(denom)
+	}
 	token, err := k.GetTokenSettings(ctx, denom)
 	if err != nil {
```
toteki, line 16: Adding this check before `GetTokenSettings`, which would fail anyway, changes the error from `not a registered denom: u/uumee` to `denom should not be a uToken: u/uumee`, which is more accurate.

x/leverage/types/tx.go
```diff
@@ -66,7 +65,7 @@ func (msg MsgCollateralize) Route() string { return ModuleName }
 func (msg MsgCollateralize) Type() string  { return EventTypeCollateralize }
 
 func (msg *MsgCollateralize) ValidateBasic() error {
-	return validateSenderAndAsset(msg.Borrower, nil)
+	return validateSenderAndAsset(msg.Borrower, &msg.Asset)
```
toteki, line 68: This was a bug

x/leverage/types/tx.go
```diff
@@ -189,8 +188,11 @@ func validateSenderAndAsset(sender string, asset *sdk.Coin) error {
 	if err != nil {
 		return err
 	}
-	if asset != nil && !asset.IsValid() {
-		return sdkerrors.Wrap(ErrInvalidAsset, asset.String())
+	if asset == nil {
+		return ErrNilAsset
+	}
+	if err := asset.Validate(); err != nil {
+		return err
```
toteki, line 195: Related to bug - we were letting `nil` through unintentionally

x/leverage/keeper/suite_test.go
```diff
@@ -0,0 +1,211 @@
...
+func (s *IntegrationTestSuite) requireEqualCoins(coinsA, coinsB sdk.Coins, msgAndArgs ...interface{}) {
+	s.Require().Equal(
+		sdk.NewCoins(coinsA...),
+		sdk.NewCoins(coinsB...),
+		msgAndArgs...,
+	)
+}
+
+// newToken creates a test token with reasonable initial parameters
+func newToken(base, symbol string) types.Token {
```
RafilxTenfen, line 91: :100:

x/leverage/keeper/token.go
```diff
@@ -20,7 +18,7 @@ func (k Keeper) SetTokenSettings(ctx sdk.Context, token types.Token) error {
 
 	bz, err := k.cdc.Marshal(&token)
 	if err != nil {
-		panic(fmt.Errorf("failed to encode token settings: %w", err))
```
RafilxTenfen, line 23: Why remove the panic?

toteki: Since the function has an error return, we can just pass the error out that way. (It it's in `EndBlock` after a gov proposal the error will turn into a panic anyway, but will have the original error message)

x/leverage/keeper/validate.go
```diff
@@ -44,32 +47,60 @@ func (k Keeper) validateAcceptedUToken(ctx sdk.Context, coin sdk.Coin) error {
 }
 
 // validateSupply validates an sdk.Coin and ensures its Denom is a Token with EnableMsgSupply
-func (k Keeper) validateSupply(ctx sdk.Context, loan sdk.Coin) error {
-	if err := loan.Validate(); err != nil {
+func (k Keeper) validateSupply(ctx sdk.Context, coin sdk.Coin) error {
+	if err := coin.Validate(); err != nil {
```
RafilxTenfen, line 51: use `validateUToken`

x/leverage/keeper/validate.go
```diff
@@ -44,32 +47,60 @@ func (k Keeper) validateAcceptedUToken(ctx sdk.Context, coin sdk.Coin) error {
...
 	if err != nil {
 		return err
 	}
 	return token.AssertBorrowEnabled()
 }
 
-// validateCollateralAsset validates an sdk.Coin and ensures it is a uToken of an accepted
+// validateRepay validates an sdk.Coin and ensures its Denom is not a uToken
+func (k Keeper) validateRepay(coin sdk.Coin) error {
+	if err := coin.Validate(); err != nil {
```
RafilxTenfen, line 92: We do this sometimes I am not against duplicating code, but we don't even need the keeper to do this '-'

x/leverage/types/errors.go
```diff
@@ -7,38 +7,44 @@ import (
...
-	ErrMinCollateralLiquidity  = sdkerrors.Register(ModuleName, 1125, "market would fall below MinCollateralLiquidity")
-	ErrMaxCollateralShare      = sdkerrors.Register(
-		ModuleName,
-		1126,
-		"market total collateral would exceed MaxCollateralShare",
-	)
-	ErrNotUToken            = sdkerrors.Register(ModuleName, 1127, "denom should be a uToken")
-	ErrUToken               = sdkerrors.Register(ModuleName, 1128, "denom should not be a uToken")
-	ErrExcessiveTimeElapsed = sdkerrors.Register(ModuleName, 1130, "excessive time elapsed since last interest time")
+	// 1XX = General Validation
```
RafilxTenfen, line 10: This is nice! 
But, any reason why we start from 1100 before?

x/leverage/keeper/oracle_test.go
```diff
@@ -43,69 +44,77 @@ func (m *mockOracleKeeper) GetExchangeRateBase(ctx sdk.Context, denom string) (s
...
-		s.ctx,
+	app, ctx, require := s.app, s.ctx, s.Require()
+
+	// (2.4 UMEE * $4.21) + (4.7 ATOM * $39.38)
+	v, err := app.LeverageKeeper.TotalTokenValue(
+		ctx,
 		sdk.NewCoins(
-			sdk.NewInt64Coin(umeeapp.BondDenom, 2400000),
-			sdk.NewInt64Coin(atomIBCDenom, 4700000),
+			coin(umeeapp.BondDenom, 2_400000),
```
RafilxTenfen, line 87: :tada: much nicer

x/leverage/keeper/validate.go
toteki: True, we could make this a regular function rather than a `(k Keeper)` method. However, some of the functions in the same family like `validateSupply` and `validateBorrow` require keeper (since they use the token registry). We would be splitting the functions up between namespaces if we removed keeper from validateRepay.

Alternately, they could all be turned into functions by moving keeper to the args, like `validateRepay(sdk.Coin)` and `validateSupply(sdk.Context, Keeper, sdk.Coin)`. That would also keep them together.

I did this originally by copy+paste magic, so I'm not too attached to either option.

x/leverage/types/errors.go
toteki: I couldn't find a reason - we start from 0 in oracle module (and external cosmos projects I looked at do the same), but leverage had been using 11XX since the start.

toteki:
Note: markdown link test fail is http `429: Too many requests` - we'll have to look at that in the future.

x/leverage/keeper/collateral_test.go
```diff
@@ -7,88 +7,111 @@ import (
...
 func (s *IntegrationTestSuite) TestGetCollateralAmount() {
+	ctx, require := s.ctx, s.Require()
 	uDenom := types.ToUTokenDenom(umeeDenom)
 
-	// get u/umee collateral amount of empty account address (zero)
-	collateral := s.tk.GetCollateralAmount(s.ctx, sdk.AccAddress{}, uDenom)
-	s.Require().Equal(sdk.NewInt64Coin(uDenom, 0), collateral)
+	// get u/umee collateral amount of empty account address
+	collateral := s.tk.GetCollateralAmount(ctx, sdk.AccAddress{}, uDenom)
+	require.Equal(coin(uDenom, 0), collateral)
```
robert-zaremba, line 15: `coin` could be a keyword :D

x/leverage/keeper/invariants.go
```diff
@@ -297,3 +304,39 @@ func InterestScalarsInvariant(k Keeper) sdk.Invariant {
...
+		}
+
+		broken := count != 0
+
+		return sdk.FormatInvariant(
+			types.ModuleName, routeExchangeRates,
+			fmt.Sprintf("amount of uToken exchange rates lower than one %d\n%s", count, msg),
+		), broken
+	}
+}
```
robert-zaremba, line 342: do we really want to check it after each block by every validator? I think we should have that function only invalidate genesis.

x/leverage/keeper/invariants_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (s *IntegrationTestSuite) TestReserveAmountInvariant() {
+	app, ctx, require := s.app, s.ctx, s.Require()
+
+	// artificially set reserves
+	s.setReserves(coin(umeeapp.BondDenom, 300_000000))
+
+	// check invariants
+	_, broken := keeper.ReserveAmountInvariant(app.LeverageKeeper)(ctx)
+	require.False(broken)
```
robert-zaremba, line 17: we should also test negative scenarios

x/leverage/keeper/iter_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	// user borrows 4 atom (max current allowed - 1) user amount enabled as collateral * CollateralWeight
+	// = (50 * 0.1) - 1
+	// = 4app.
+	s.borrow(addr, coin(atomDenom, 4_000000))
+
+	// Note: Setting umee liquidation threshold to 0.05 to make the user eligible for liquidation
+	umeeToken := newToken("uumee", "UMEE")
+	umeeToken.CollateralWeight = sdk.MustNewDecFromStr("0.05")
+	umeeToken.LiquidationThreshold = sdk.MustNewDecFromStr("0.05")
```
robert-zaremba, line 63: this 3 lines are repeated often. Could be better to create a function: `newUmeeToken(collateraW, liquidationT)`

x/leverage/keeper/suite_test.go
```diff
@@ -0,0 +1,211 @@
...
+	return fixtures.Token(base, symbol)
+}
+
+// coin creates a coin with a given base denom and amount
+func coin(denom string, amount int64) sdk.Coin {
+	return sdk.NewInt64Coin(denom, amount)
+}
+
+// newAccount creates a new account for testing, and funds it with any input tokens.
+func (s *IntegrationTestSuite) newAccount(funds ...sdk.Coin) sdk.AccAddress {
```
robert-zaremba, line 101: :+1:

robert-zaremba (commented):
late review. Good job a nice improvements. Left few small things / suggestions.

x/leverage/keeper/invariants.go
toteki: Some of these would be expensive computationally to do each block, but having checked other modules (e.g. [x/bank invariants](https://github.com/cosmos/cosmos-sdk/blob/main/x/bank/keeper/invariants.go) I see that iteration over balances is done there as well.

Is there some setting that validators are using to run invariants less often as they operate other chains?

robert-zaremba: No, there is no such setting. But we can code it. The x/bank invariants are often disabled, because they take lot of time.

robert-zaremba: I've made a task for this issue: https://github.com/umee-network/umee/issues/1324

robert-zaremba: BTW, the bank invariants are not even registered in the simapp.

PR #1412, opened 2022-09-18
chore: bump Gravity-Bridge

robert-zaremba:
## Description

In this update, Gravity Bridge `MsgValsetConfirm` and `MsgValsetUpdatedClaim` is active, and validators must update the validator set in Ethereum smart contract to fully enable Gravity Bridge and not get slashed.
This is the first step to fully enable Gravity Bridge, after we disabled it in `v1.1.x`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/cosmos/cosmos-proto v1.0.0-alpha7
 	github.com/cosmos/cosmos-sdk v0.46.1
 	github.com/cosmos/go-bip39 v1.0.0
-	github.com/cosmos/ibc-go/v5 v5.0.0-rc1.0.20220831174724-8045fcccbd30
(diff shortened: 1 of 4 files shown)


robert-zaremba:
@adamewozniak this update allows validators to have a "slow start" with peggo. They won't get slashed if they are late etc...
We need this update because:
* we need to enable GB sooner than later
* this will allow us to test the status of valset in GB contract.

RafilxTenfen:
I will take a look on e2e

facundomedica:
Just saw this one. We also need to enable the Batch Confirms to be 100% safe.
BTW MsgValsetUpdatedClaim is not necessary to avoid slashes (the Umee chain doesn't care if the ETH valset is up to date)

robert-zaremba:
We don't enable batch confirms, only Valset related messages.

PR #1435, opened 2022-09-20
chore: update gravity bridge

robert-zaremba:
## Description

backporting the "disable slashing"


robert-zaremba:
sim tests has again missing cashe issue.

PR #1479, opened 2022-10-06
feat: add MsgSupplyCollateral

toteki:
This might or might not be okay to merge in a minor version (due to adding a message type). Supports a common frontend use case.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


toteki:
This PR should be evaluated for inclusion in `v3.1.0`

robert-zaremba (commented):
I don't think we need it. API, especially on proto level, should not have redundancy when possible.
The same effect can be achieved with one transaction with multiple messages.

x/leverage/keeper/msg_server.go
```diff
@@ -108,6 +108,51 @@ func (s msgServer) Collateralize(
...
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	supplierAddr, err := sdk.AccAddressFromBech32(msg.Supplier)
+	if err != nil {
+		return nil, err
+	}
+	uToken, err := s.keeper.Supply(ctx, supplierAddr, msg.Asset)
+	if err != nil {
+		return nil, err
+	}
```
robert-zaremba, line 124: how about calling `s.Supply` and `s.Collaterize` directly here (by creating the appropriate message)?

robert-zaremba:
> this could have unintended consequences if we feed them into MsgServer (I'm not 100% sure).

You don't go through MsgServer, and no need to check signatures. You directly call the handler.
In fact even if you would go through MsgServer (which is not clearly supported today), it won't check signatures nor call ValidateBasic (which is happening at different stage)

robert-zaremba (approved):
pre-approving. See https://github.com/umee-network/umee/pull/1479#issuecomment-1270837149 . That would be my preference, but I'm not against using the current way. I just prefer to avoid possible duplications.

Issue #1470, opened 2022-10-04
Epic: Price feeder v2


adamewozniak:
## Summary

Epic to track Price Feeder V2 tickets, emphasizing asset support and debugging.

Closes #1470

- [x] #1305
- [x] #792
- [x] https://github.com/umee-network/osmosis-api/issues/43
- [x] https://github.com/umee-network/osmosis-api/issues/42
- [x] https://github.com/umee-network/osmosis-api/issues/41
- [x] #1263 
- [x] #1380
- [x] #1467
- [x] #1507
- [x] #1584

---

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

adamewozniak:
I'm keeping this open for now even though all these tasks are through, since releasing price feeder v2.0.0 is dependent on https://github.com/umee-network/umee/issues/1521

adamewozniak:
Closing since we released this anyways as a part of the last asset update :^)

PR #1496, opened 2022-10-13
feat: Query coin gecko for available providers for a given currency

rbajollari:
## Description

closes: #1470 (addresses "three-provider-rule" limitation in issue)
---

### Author Checklist

I have...

### Reviewers Checklist

I have...


adamewozniak:
Reviewing this today! Awesome work @rbajollari 🚀

price-feeder/config/config.go
```diff
@@ -216,9 +218,17 @@ func ParseConfig(configPath string) (Config, error) {
 		}
 	}
 
+	// Use currencyProviderTracker to check available exchanges for each currency pair
+	// and determine minimum amount of providers required for each pair
+	currencyProviderTracker, err := NewCurrencyProviderTracker(ctx, logger, cfg.CurrencyPairs...)
+	if err != nil {
+		return cfg, fmt.Errorf("failed to start currency provider tracker: %w", err)
+	}
+
```
adamewozniak, line 227: Imo the `ParseConfig` function should be limited to parsing the config file itself and checking against hard-coded rules. Maybe we move this into another function, `CheckProviderMins` or somthing?

price-feeder/config/config_test.go
```diff
@@ -199,7 +201,8 @@ global-labels = [["chain-id", "umee-local-testnet"]]
 	_, err = tmpFile.Write(content)
 	require.NoError(t, err)
 
-	cfg, err := config.ParseConfig(tmpFile.Name())
+	logger := zerolog.New(zerolog.ConsoleWriter{Out: os.Stderr}).Level(zerolog.InfoLevel).With().Timestamp().Logger()
```
adamewozniak, line 204: let's move this logger to a variable at the top of the file so we don't have to keep making new ones

price-feeder/config/currency_provider_tracker.go
```diff
@@ -0,0 +1,175 @@
...
+	// support the currency pairs set in the price feeder config. It will poll the API
+	// every 24 hours to log any new exchanges that were added for a given currency.
+	//
+	// REF: https://www.coingecko.com/en/api/documentation
+	CurrencyProviderTracker struct {
+		logger              zerolog.Logger
+		pairs               []CurrencyPair
+		coinIDSymbolMap     map[string]string   // ex: map["ATOM"] = "cosmos"
+		CurrencyProviders   map[string][]string // map of price feeder currencies and what exchanges support them
+		CurrencyProviderMin map[string]int      // map of price feeder currencies and min required providers for them
```
adamewozniak, line 33: Awesome docs here!!

price-feeder/config/currency_provider_tracker.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	// List of assets on CoinGecko and their corresponding id and symbol.
+	CoinGeckoCoinList struct {
```
adamewozniak, line 37: Unless this struct is used outside this package, let's keep this internal. That way we can name it locally rather than appending "coinGecko" in front:
```suggestion
	coinList struct {
```

price-feeder/config/currency_provider_tracker.go
```diff
@@ -0,0 +1,175 @@
...
+		ID     string `json:"id"`     // ex: "cosmos"
+		Symbol string `json:"symbol"` // ex: "ATOM"
+	}
+
+	// CoinGecko ticker shows market data for a given currency pair including what
+	// exchanges they're on.
+	CoinGeckoCoinTickerResponse struct {
```
adamewozniak, line 44: Same here & for the others:
```suggestion
	coinTickerResponse struct {
```

price-feeder/config/currency_provider_tracker.go
```diff
@@ -0,0 +1,175 @@
+package config
```
adamewozniak, line 1: I think `config` might be the wrong package here, maybe we create a new folder for this file? maybe we put it in a utils package? wdyt?

price-feeder/config/currency_provider_tracker.go
```diff
@@ -0,0 +1,175 @@
...
+	return currencyProviderTracker, nil
+}
+
+func (t *CurrencyProviderTracker) getCurrencyProviders() map[string][]string {
+	return t.CurrencyProviders
+}
+
+func (t *CurrencyProviderTracker) getCurrencyProviderMin(base string) int {
+	return t.CurrencyProviderMin[base]
+}
```
adamewozniak, line 91: There's not really a point for having these getter functions without a mutex, would adding one be helpful for making sure we're threadsafe here?

Also, if we want to control how other packages are getting `CurrencyProviderMin` we should make it private

price-feeder/config/currency_provider_tracker.go
```diff
@@ -0,0 +1,175 @@
...
+
+func (t *CurrencyProviderTracker) logCurrencyProviders() {
+	for currency, providers := range t.CurrencyProviders {
+		t.logger.Info().Msg(fmt.Sprintf("Providers supporting %s: %v", currency, providers))
```
adamewozniak, line 95: nit:
```suggestion
		t.logger.Info().Msg(fmt.Sprintf("providers supporting %s: %v", currency, providers))
```

price-feeder/config/currency_provider_tracker.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+}
+
+// setCoinIDSymbolMap gets list of assets on coingecko to cross reference coin symbol to id.
+func (t *CurrencyProviderTracker) setCoinIDSymbolMap() error {
+	resp, err := http.Get(coinGeckoRestHost + coinGeckoRestPath + coinGeckoListEndpoint)
```
adamewozniak, line 101: let's use interpolation here so that we don't have to have slashes in `coinGeckoRestHost` & `coinGeckoRestPath`, and it looks a bit cleaner

price-feeder/config/currency_provider_tracker.go
```diff
@@ -0,0 +1,175 @@
...
+
+	return nil
+}
+
+// setCurrencyProviders queries CoinGecko's tickers endpoint to get all the exchanges that
+// support each price feeder currency pair and store it in the CurrencyProviders map
+func (t *CurrencyProviderTracker) setCurrencyProviders() error {
+	for _, pair := range t.pairs {
+		pairBaseID := t.coinIDSymbolMap[strings.ToLower(pair.Base)]
+		resp, err := http.Get(coinGeckoRestHost + coinGeckoRestPath + pairBaseID + "/" + coinGeckoTickersEndpoint)
```
adamewozniak, line 124: Same, interpolation here please :^)

price-feeder/config/currency_provider_tracker.go
```diff
@@ -0,0 +1,175 @@
...
+			}
+		}
+	}
+
+	return nil
+}
+
+// setCurrencyProviderMin will set the minimum amount of providers for each currency
+// to the amount of exchanges that support them if its less than 3. Otherwise it is
+// set to 3 providers.
```
adamewozniak, line 147: Perfect 🚀 🚀

price-feeder/config/currency_provider_tracker.go
```diff
@@ -0,0 +1,175 @@
...
+		if len(exchanges) < 3 {
+			t.CurrencyProviderMin[base] = len(exchanges)
+		} else {
+			t.CurrencyProviderMin[base] = 3
+		}
+	}
+}
+
+// trackCurrencyProviders resets CurrencyProviders map and logs out supported
+// exchanges for each currency every 24 hours
```
adamewozniak, line 159: As a general rule if it's a comment as a complete sentence, add a period - just a nitpick 
```suggestion
// trackCurrencyProviders resets CurrencyProviders map and logs supported
// exchanges for each currency every 24 hours.
```

price-feeder/config/currency_provider_tracker.go
```diff
@@ -0,0 +1,175 @@
...
+func (t *CurrencyProviderTracker) trackCurrencyProviders(ctx context.Context) {
+	t.logCurrencyProviders()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(trackingPeriod):
+			if err := t.setCurrencyProviders(); err != nil {
+				t.logger.Error().Err(err).Msg("failed to set available providers for currencies")
```
adamewozniak, line 169: This is great! Question, what happens if on startup the validator is unable to reach the coingecko API, but they have 3 providers per asset? Will it not start up?

adamewozniak (commented):
Great work @rbajollari !! Mostly just nitpicks.
Question, what happens if:
1. A new market comes online in the middle of the price feeder running? in this case it should log a warning message, but it shouldn't stop
2. A new market comes online for a provider we haven't implemented, and then a validator starts to run the price feeder? In this case we also shouldn't stop. We'd be super vulnerable if new markets started to pop up for some of these less-popular assets.

price-feeder/config/currency_provider_tracker.go
rbajollari: That's a good point. Unless you prefer an alternate approach, I'm gonna make it so that if the request fails it will just set the minimum providers to 3 by default and just log that the request failed for that currency.

adamewozniak: @rbajollari what if coingecko is down for an extended period of time, say a week? We've then got a bottleneck for validators' votes. Maybe we need an override in the config?

Wdyt @zarazan ?

zarazan: If coingecko is down I think you should be able to start the price-feeder and everything should work normally except a warning that it was unable to do currency checks. That means if we allow a single provider when there is only one available that should still be allowed if coingecko is down.

zarazan: A config override is a decent option. It should probably just bypass the validation completely if you use the override.

price-feeder/config/currency_provider_tracker.go
zarazan: I'm a little torn on this one, it technically is a validation for your config...this file has a good structure and it wouldn't bother me if it lived here until it becomes larger and we move it to its own package.

price-feeder/config/config.go
zarazan: This code is probably fine in `cmdHandler`

price-feeder/config/config.go
```diff
@@ -216,9 +218,17 @@ func ParseConfig(configPath string) (Config, error) {
...
 	for base, providers := range pairs {
-		if _, ok := pairs[base][provider.ProviderMock]; !ok && len(providers) < 3 {
-			return cfg, fmt.Errorf("must have at least three providers for %s", base)
+		minProviders := currencyProviderTracker.getCurrencyProviderMin(base)
```
zarazan, line 229: Instead of having to call functions on currencyProviderTracker here you could return the entire `map[string]int` initially from the currencyProviderTracker. It would turn currencyProviderTracker into more of a service architecture where you pass in currencyPairs and get back a map of the min providers for each one.

This check could probably be moved outside of ParseConfig to reduce how intertwined the responsibilities are.

price-feeder/config/currency_provider_tracker.go
adamewozniak: Override > defaulting to allowing any amount of providers imo, at least at startup - just because we might never hear about it if validators don't have to set something manually

adamewozniak:
Oh also, changelog entry pls 💙

price-feeder/config/currency_provider_tracker.go
rbajollari: So if the override config just doesn't enforce the provider minimum when set, does it take precedence over the coin gecko check or only take effect when the coin gecko request fails? @adamewozniak

adamewozniak: Good question, I think it should only take effect when the coingecko request fails @rbajollari

price-feeder/config/config.go
rbajollari: Created CheckProviderMins method and I'm calling it in priceFeederCmdHandler now so ParseConfig doesn't need to take the logger and context anymore.

price-feeder/config/currency_provider_tracker.go
rbajollari: I agree with @zarazan on this since it seems like this isn't big enough to warrant its own package. That being said, if we plan on expanding it to use more than CoinGecko as a data source, that may justify moving it. Do you think its fine under config for now @adamewozniak?

price-feeder/config/currency_provider_tracker.go
rbajollari: Using interpolation and put coinGeckoRestHost & coinGeckoRestPath together into coinGeckoRestUrl. Lmk how it looks!

price-feeder/config/config.go
rbajollari: I've removed the getter methods now and the check into another method.

price-feeder/config/currency_provider_tracker.go
rbajollari: @adamewozniak Implemented that as `provider_min_override` config. Lmk what you think about it.

rbajollari:
Regarding the 2 questions at the top of the review. The currency provider tracker is just logging what providers are available for a given currency every 24 hours; it won't stop the price-feeder. The only time it will stop the price-feeder is at startup if the config isn't set with the right amount of minimum providers. 


>

price-feeder/config/currency_provider_tracker.go
adamewozniak: Perfect!

price-feeder/price-feeder.example.toml
```diff
@@ -93,6 +93,8 @@ global-labels = [["chain_id", "umee-local-testnet"]]
 service-name = "price-feeder"
 prometheus-retention-time = 100
 
+provider_min_override = false
```
adamewozniak, line 96: Is this value required? If not let's default to false pls

price-feeder/config/currency_provider_tracker.go
```diff
@@ -0,0 +1,166 @@
...
+				t.CurrencyProviders[pair.Base] = append(t.CurrencyProviders[pair.Base], ticker.Market.Name)
+			}
+		}
+	}
+
+	return nil
+}
+
+// setCurrencyProviderMin will set the minimum amount of providers for each currency
+// to the amount of exchanges that support them if its less than 3. Otherwise it is
```
adamewozniak, line 137:
```suggestion
// to the amount of exchanges that support them if it's less than 3. Otherwise it is
```

price-feeder/config/currency_provider_tracker.go
adamewozniak: Agree, no need to put it in a separate package until we add more APIs 🔥

adamewozniak (approved):
lgtm! just a couple nits

zarazan (approved):
🎉

PR #1510, opened 2022-10-19
feat: Integrate osmosis-api into price-feeder

rbajollari:
## Description

closes: https://github.com/umee-network/osmosis-api/issues/42

- currency_provider_tracker.go was edited to account for OSMO/ATOM not being a supported market when querying the CoinGecko API and only supported by our osmosis-api (a better approach can be investigated in another PR perhaps)
- ATOM was added as a supported quote currency in config.go to support the OSMO/ATOM pair

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


adamewozniak:
🔥 🔥 🔥 🔥 🔥

price-feeder/config/config.go
```diff
@@ -268,6 +270,8 @@ func CheckProviderMins(ctx context.Context, logger zerolog.Logger, cfg Config) e
 		}
 	}
 
+	fmt.Println("pairs", pairs)
+
```
adamewozniak, line 274:
```suggestion
```

price-feeder/config/currency_provider_tracker.go
```diff
@@ -111,6 +111,11 @@ func (t *CurrencyProviderTracker) setCoinIDSymbolMap() error {
 // support each price feeder currency pair and store it in the CurrencyProviders map.
 func (t *CurrencyProviderTracker) setCurrencyProviders() error {
 	for _, pair := range t.pairs {
+		// OSMO/ATOM is supported by UMEE's osmosis-api.
+		if pair.Base == "OSMO" && pair.Quote == "ATOM" {
+			t.CurrencyProviders[pair.Base] = append(t.CurrencyProviders[pair.Base], "osmosisv2")
+		}
+
```
adamewozniak, line 118: So will we have to do this every time we add another asset to the osmo API?

price-feeder/oracle/provider/osmosisv2.go
```diff
@@ -0,0 +1,415 @@
...
+	volume, err := sdk.NewDecFromStr(candlePair.Volume)
+	if err != nil {
+		p.logger.Warn().Err(err).Msg("osmosisv2: failed to parse candle volume")
+		return
+	}
+	candle := types.CandlePrice{
+		Price:  close,
+		Volume: volume,
+		// convert seconds -> milli
+		TimeStamp: SecondsToMilli(candlePair.EndTime),
```
adamewozniak, line 308: iirc the endTime that the osmo api spits out is already in milliseconds
```suggestion
		TimeStamp: candlePair.EndTime,
```

adamewozniak (changes requested):
Just a couple things
Also, when I run this locally I get these prices:

```json
{"prices":{"ATOM":"11.708378187296907403","OSMO":"298476702737041502812695741623261092264129340012544696292460811.171082010520520593","UMEE":"0.011180009912590074","USDC":"0.999901023107751526","USDT":"1.000000886635933422"}}
```

And then eventually an int overflow panic. Somehow osmosis's price is increasing more as the PF runs

price-feeder/oracle/provider/osmosisv2.go
```diff
@@ -0,0 +1,415 @@
...
+	for _, c := range p.candles[symbol] {
+		if staleTime < c.TimeStamp {
+			candleList = append(candleList, c)
+		}
+	}
+
+	p.candles[symbol] = candleList
+}
+
+func (p *OsmosisV2Provider) handleWebSocketMsgs(ctx context.Context, pairs ...types.CurrencyPair) {
```
zarazan, line 324: You could use the WebsocketController instead of duplicating this code. (https://github.com/umee-network/umee/blob/main/price-feeder/oracle/provider/mexc.go#L114). There is already an issue to convert the rest of the providers over so we can also wait until then.

adamewozniak: Let's definitely use the controller as we add new providers, less cleanup for later

price-feeder/config/currency_provider_tracker.go
rbajollari: Only if that market does not exist on coin gecko

price-feeder/oracle/provider/osmosisv2.go
```diff
@@ -0,0 +1,415 @@
...
+	}
+
+	p.logger.Error().
+		Int("length", len(bz)).
+		AnErr("message", messageErr).
+		Msg("Error on receive message")
+}
+
+// ping to check websocket connection
+func (p *OsmosisV2Provider) ping() error {
```
zarazan, line 265: I don't think the osmosis-api responds to the standard ping message (maybe it should???). https://github.com/umee-network/osmosis-api/blob/main/server/client.go#L32. However, this might still work - because it will successfully send a message - the server will receive it and ignore it. It probably also fails after the connection has been closed.

adamewozniak: Ping is typically used as a way to say "please keep me alive!", here we can just keep it as a "are you there?" message. We should add a test to make sure the ping fails if there is no active connection.

One thing to note @zarazan is that the websocket API currently doesn't disconnect anything after x amount of time. Can you add a 24 hour disconnect, with a limit of 250 current clients? (we've currently got 111 validators in the inactive set)

price-feeder/config/currency_provider_tracker.go
adamewozniak: @rbajollari why does this market not exist on coingecko? Does it only have "ATOM/OSMO"?

We can just allow the flip of these too for the websocket API (I.e. if ATOM/OSMO exists, then OSMO/ATOM also exists)

price-feeder/oracle/provider/osmosisv2.go
rbajollari: ping method is no longer there after using the WebsocketController

price-feeder/oracle/provider/osmosisv2.go
rbajollari: Done 👍

rbajollari:
Figured out that the increasing OSMO price over time was caused by the provider candles map pointer being passed and written to by the convert to usd logic in the price-feeder (since the quote for OSMO is ATOM this resulted in the OSMO candle price being multiplied by the price of ATOM every tick). I've made this fix for the MEXC and Crypto provider as well.

price-feeder/config/currency_provider_tracker.go
rbajollari: Should I address this in another PR since we talked about exposing what assets are supported on our osmo API through an endpoint on it?

adamewozniak: @rbajollari yeah let's do this in another PR after the websocket adds an endpoint for it

adamewozniak (approved):
this looks good! we'll test it a bit more with the osmo websocket API after I update it with kyle's most recent update 🚀 

would love a review from @zarazan here before merging in

zarazan (approved):
Good job with that bug 🔥

PR #1514, opened 2022-10-23
DNM: incentive module proto

toteki:
**Marking as DO NOT MERGE so I can break it up into smaller PRs**

This PR contains the first step in the incentive module

Progress:
- Proto files
- Proto gen
- Implement Msg interfaces

---

Design of gov proposals:

I have created two options for funding incentive programs:
- one that uses a community fund or known address to fund a program the moment it passes governance
- another that creates the program and sets its total rewards via governance, but requires a basic account to fund it after passing or it won't start. 

### Two step process (basic account)

If doing external incentives, we may want to let the source of the funding use a basic account. In that case, `MsgGovCreateProgram` (and governance vote) is used to set the program's `max_rewards` but NOT fund it.

Then, a `MsgSponsorProgram` can be used by the basic account, which must fund the exact amount specified in `max_rewards` or the program will not start / activate.

### One step process (community fund)

For ordinary programs that use the community fund, or instances where external incentives are already waiting at a known address, a `MsgCreateAndSponsorProgram` is used to do it all in one step.

---

### Author Checklist


proto/umee/incentive/v1/tx.proto
```diff
@@ -0,0 +1,114 @@
...
+  rpc BeginUnbonding(MsgBeginUnbonding) returns (MsgBeginUnbondingResponse);
+
+  // Sponsor defines a permissionless method for sponsoring an incentive program.
+  rpc Sponsor(MsgSponsor) returns (MsgSponsorResponse);
+
+  // CreateProgram is used by governance proposals to create (but not fund) an incentive program.
+  rpc CreateProgram(MsgCreateProgram) returns (MsgCreateProgramResponse);
+
+  // CreateAndSponsorProgram is used by governance proposals to create and fund an incentive program.
+  rpc CreateAndSponsorProgram(MsgCreateAndSponsorProgram) returns (MsgCreateAndSponsorProgramResponse);
```
toteki, line 30: These separate flows for creating incentive programs through governance give us multiple options for how to create external incentives when the time comes - either funding them on creation from a predetermined address (probably community fund) or creating the incentive program but letting any regular account use `MsgSponsor` to fund it after it passes.

x/incentive/types/tx.go
```diff
@@ -0,0 +1,317 @@
...
+	if len(description) == 0 {
+		return sdkerrors.Wrap(types.ErrInvalidProposalContent, "proposal description cannot be blank")
+	}
+	if len(description) > gov1b1.MaxDescriptionLength {
+		return sdkerrors.Wrapf(types.ErrInvalidProposalContent, "proposal description is longer than max length of %d",
+			gov1b1.MaxDescriptionLength)
+	}
+
+	return nil
+}
```
toteki, line 282:
```suggestion
func (msg *MsgCreateAndSponsorProgram) ValidateAbstract() error {
	title := msg.GetTitle()
	if len(strings.TrimSpace(title)) == 0 {
		return sdkerrors.Wrap(types.ErrInvalidProposalContent, "proposal title cannot be blank")
	}
	if len(title) > gov1b1.MaxTitleLength {
		return sdkerrors.Wrapf(types.ErrInvalidProposalContent, "proposal title is longer than max length of %d",
			gov1b1.MaxTitleLength)
	}

	description := msg.GetDescription()
	if len(description) == 0 {
		return sdkerrors.Wrap(types.ErrInvalidProposalContent, "proposal description cannot be blank")
	}
	if len(description) > gov1b1.MaxDescriptionLength {
		return sdkerrors.Wrapf(types.ErrInvalidProposalContent, "proposal description is longer than max length of %d",
			gov1b1.MaxDescriptionLength)
	}

	return nil
}
```

x/incentive/types/tx.go
```diff
@@ -0,0 +1,317 @@
...
```
toteki, line 196:
```suggestion
func (msg *MsgCreateProgram) ValidateAbstract() error {
	title := msg.GetTitle()
	if len(strings.TrimSpace(title)) == 0 {
		return sdkerrors.Wrap(types.ErrInvalidProposalContent, "proposal title cannot be blank")
	}
	if len(title) > gov1b1.MaxTitleLength {
		return sdkerrors.Wrapf(types.ErrInvalidProposalContent, "proposal title is longer than max length of %d",
			gov1b1.MaxTitleLength)
	}

	description := msg.GetDescription()
	if len(description) == 0 {
		return sdkerrors.Wrap(types.ErrInvalidProposalContent, "proposal description cannot be blank")
	}
	if len(description) > gov1b1.MaxDescriptionLength {
		return sdkerrors.Wrapf(types.ErrInvalidProposalContent, "proposal description is longer than max length of %d",
			gov1b1.MaxDescriptionLength)
	}

	return nil
}
```

proto/umee/incentive/v1/genesis.proto
```diff
@@ -0,0 +1,74 @@
...
+    (gogoproto.nullable)     = false,
+    (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.DecCoins"
+  ];
+}
+
+// RewardAccumulator is a global reward tracking struct that indicates the amount
+// of rewards that a single unit of lock_denom would have acucmulated if it was
+// locked at a given tier since genesis.
+message RewardAccumulator {
+  uint32   tier                                     = 1;
```
robert-zaremba, line 60: shouldn't this be part of a key?

toteki: In store, the key is `denom | tier` and the value is a `DecCoins`, but in genesis they must all be in a struct

(similar to how `x/bank` balance structs contain both `address` and `Coins` in their genesis.proto)

x/incentive/types/params.go
```diff
@@ -0,0 +1,166 @@
...
+	defaultTierWeightMiddle        = sdk.MustNewDecFromStr("0.8")
+)
+
+func NewParams() Params {
+	return Params{}
+}
+
+// ParamSetPairs implements the ParamSet interface and returns all the key/value
+// pairs pairs of x/incentive module's parameters.
+func (p *Params) ParamSetPairs() paramtypes.ParamSetPairs {
```
robert-zaremba, line 39: This file should be removed, and module params should be used (managed directly by module keeper and the new x/gov authority) rather than the paramspace.

https://github.com/cosmos/cosmos-sdk/blob/main/docs/architecture/adr-046-module-params.md#backwards-compatibility

x/incentive/types/tx.go
```diff
@@ -0,0 +1,317 @@
...
+)
+
+func NewMsgClaim(supplier sdk.AccAddress) *MsgClaim {
+	return &MsgClaim{
+		Supplier: supplier.String(),
+	}
+}
+
+func (msg MsgClaim) Route() string { return sdk.MsgTypeURL(&msg) }
+func (msg MsgClaim) Type() string  { return sdk.MsgTypeURL(&msg) }
```
robert-zaremba, line 30:
```suggestion
```

robert-zaremba: not needed.

x/incentive/types/tx.go
```diff
@@ -0,0 +1,317 @@
...
+	return &MsgLock{
+		Supplier: supplier.String(),
+		Tier:     tier,
+		Asset:    asset,
+	}
+}
+
+func (msg MsgLock) Route() string { return sdk.MsgTypeURL(&msg) }
+func (msg MsgLock) Type() string  { return sdk.MsgTypeURL(&msg) }
+
```
robert-zaremba, line 57:
```suggestion
```

x/incentive/types/tx.go
```diff
@@ -0,0 +1,317 @@
...
+	return &MsgBeginUnbonding{
+		Supplier: supplier.String(),
+		Tier:     tier,
+		Asset:    asset,
+	}
+}
+
+func (msg MsgBeginUnbonding) Route() string { return sdk.MsgTypeURL(&msg) }
+func (msg MsgBeginUnbonding) Type() string  { return sdk.MsgTypeURL(&msg) }
+
```
robert-zaremba, line 82:
```suggestion
```

x/incentive/types/tx.go
```diff
@@ -0,0 +1,317 @@
...
+	return &MsgSponsor{
+		Sponsor: sponsor.String(),
+		Program: programID,
+		Asset:   asset,
+	}
+}
+
+func (msg MsgSponsor) Route() string { return sdk.MsgTypeURL(&msg) }
+func (msg MsgSponsor) Type() string  { return sdk.MsgTypeURL(&msg) }
+
```
robert-zaremba, line 107:
```suggestion
```

x/incentive/types/tx.go
```diff
@@ -0,0 +1,317 @@
...
+		Title:       title,
+		Description: description,
+		Program:     program,
+		Authority:   authority,
+	}
+}
+
+func (msg MsgCreateProgram) Route() string { return sdk.MsgTypeURL(&msg) }
+func (msg MsgCreateProgram) Type() string  { return sdk.MsgTypeURL(&msg) }
+
```
robert-zaremba, line 136:
```suggestion
```

x/incentive/types/tx.go
```diff
@@ -0,0 +1,317 @@
...
+	if len(description) == 0 {
+		return sdkerrors.Wrap(types.ErrInvalidProposalContent, "proposal description cannot be blank")
+	}
+	if len(description) > gov1b1.MaxDescriptionLength {
+		return sdkerrors.Wrapf(types.ErrInvalidProposalContent, "proposal description is longer than max length of %d",
+			gov1b1.MaxDescriptionLength)
+	}
+
+	return nil
+}
```
robert-zaremba, line 196: why not merging this into the `ValidateBasic`

x/incentive/types/tx.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+	if !msg.Program.FundedRewards.IsZero() {
+		return ErrNonzeroFundedRewards.Wrap(msg.Program.FundedRewards.String())
+	}
+
+	return nil
+}
+
+func (msg *MsgCreateProgram) ValidateAbstract() error {
+	title := msg.GetTitle()
```
robert-zaremba, line 177: let's remove getters (see example in x/leverage proto files) to make the code smaller.

x/incentive/types/tx.go
```diff
@@ -0,0 +1,317 @@
...
+	if len(strings.TrimSpace(title)) == 0 {
```
robert-zaremba, line 178:
```suggestion
	if len(strings.TrimSpace(msg.Title)) == 0 {
```

x/incentive/types/tx.go
```diff
@@ -0,0 +1,317 @@
...
+		Description: description,
+		Program:     program,
+		Authority:   authority,
+		Sponsor:     sponsor,
+	}
+}
+
+func (msg MsgCreateAndSponsorProgram) Route() string { return sdk.MsgTypeURL(&msg) }
+func (msg MsgCreateAndSponsorProgram) Type() string  { return sdk.MsgTypeURL(&msg) }
+
```
robert-zaremba, line 222:
```suggestion
```

x/incentive/types/tx.go
```diff
@@ -0,0 +1,317 @@
...
+func (msg MsgCreateAndSponsorProgram) String() string {
+	out, _ := yaml.Marshal(msg)
+	return string(out)
+}
```
robert-zaremba, line 226: Don't need to overwrite `String` method - proto already does that.

x/incentive/types/tx.go
```diff
@@ -0,0 +1,317 @@
...
+	if len(description) == 0 {
+		return sdkerrors.Wrap(types.ErrInvalidProposalContent, "proposal description cannot be blank")
+	}
+	if len(description) > gov1b1.MaxDescriptionLength {
+		return sdkerrors.Wrapf(types.ErrInvalidProposalContent, "proposal description is longer than max length of %d",
+			gov1b1.MaxDescriptionLength)
+	}
+
+	return nil
+}
```
robert-zaremba, line 282: ditto, should go into `ValidateBasic`

robert-zaremba (commented):
- [ ] please follow the 0.46 module organization: types should be generated directly into the root package of the module, together with codec, errors and keys.
- [x] don't use yaml tags - they are deprecated
- [ ] don't use `paramtypes.ParamSet` - new gov mechanism should be used instead when possible.
- [x] tx.go with sdk.Msg implementations should be renamed to `msg.go`

x/incentive/types/tx.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+
+	description := msg.GetDescription()
+	if len(description) == 0 {
+		return sdkerrors.Wrap(types.ErrInvalidProposalContent, "proposal description cannot be blank")
+	}
+	if len(description) > gov1b1.MaxDescriptionLength {
+		return sdkerrors.Wrapf(types.ErrInvalidProposalContent, "proposal description is longer than max length of %d",
+			gov1b1.MaxDescriptionLength)
+	}
```
robert-zaremba, line 193: this code is repeated, let's make a function for this

x/incentive/types/tx.go
```diff
@@ -0,0 +1,317 @@
...
+
+func (msg *MsgCreateProgram) ValidateAbstract() error {
+	title := msg.GetTitle()
+	if len(strings.TrimSpace(title)) == 0 {
+		return sdkerrors.Wrap(types.ErrInvalidProposalContent, "proposal title cannot be blank")
+	}
+	if len(title) > gov1b1.MaxTitleLength {
+		return sdkerrors.Wrapf(types.ErrInvalidProposalContent, "proposal title is longer than max length of %d",
+			gov1b1.MaxTitleLength)
+	}
```
robert-zaremba, line 184: this code is repeated, let's make a function for this

x/incentive/types/codec.go
```diff
@@ -0,0 +1,48 @@
...
+	RegisterLegacyAminoCodec(amino)
+	cryptocodec.RegisterCrypto(amino)
+	amino.Seal()
+}
+
+// RegisterLegacyAminoCodec registers the necessary x/leverage interfaces and
+// concrete types on the provided LegacyAmino codec. These types are used for
+// Amino JSON serialization.
+func RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {
+	cdc.RegisterConcrete(&MsgClaim{}, "umee/incentive/MsgClaim", nil)
```
robert-zaremba, line 30:
```suggestion
	cdc.RegisterConcrete(&MsgClaim{}, proto.MessageName(&MsgClaim{}), nil)
```

robert-zaremba: same for the msgs below

x/incentive/types/codec.go
```diff
@@ -0,0 +1,48 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/msgservice"
+)
+
+var (
+	amino = codec.NewLegacyAmino()
+
+	// ModuleCdc references the global x/incentive module codec. Note, the codec
+	// should ONLY be used in certain instances of tests and for JSON encoding as
+	// Amino is still used for that purpose.
```
robert-zaremba, line 16:
```suggestion
```

robert-zaremba: this will be also used by all dependencies, we removed all that legacy comments.

proto/umee/incentive/v1/tx.proto
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// MsgSponsorResponse defines the Msg/Sponsor response type.
+message MsgSponsorResponse {}
+
+// MsgCreateProgram is used by governance to create an incentive program.
+message MsgCreateProgram {
+  option (gogoproto.equal)            = true;
+  option (gogoproto.goproto_getters)  = false;
+  option (gogoproto.goproto_stringer) = false;
```
robert-zaremba, line 80:
```suggestion
  option (gogoproto.goproto_stringer) = true;
```

robert-zaremba: should keep proto stringer

robert-zaremba: Same in other places below.

proto/umee/incentive/v1/incentive.proto
```diff
@@ -0,0 +1,89 @@
...
+  // unbonding_duration_short defines the unbonding duration (in blocks) of the short tier.
+  uint64 unbonding_duration_short = 4 [(gogoproto.moretags) = "yaml:\"lock_duration_short\""];
+
+  // The tier_weight_short defines how the proportion of rewards assets locked
+  // in the short unbonding duration receive compared to what the same amount
+  // would receive on the long tier.
+  string tier_weight_short = 5 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
+    (gogoproto.nullable)   = false,
+    (gogoproto.moretags)   = "yaml:\"tier_weight_short\""
```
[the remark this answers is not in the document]
toteki: removed

proto/umee/incentive/v1/genesis.proto
```diff
@@ -0,0 +1,74 @@
...
+    (gogoproto.nullable)     = false,
+    (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.DecCoins"
+  ];
+}
+
+// Unbonding is a structure that tracks an in-progress token unbonding.
+message Unbonding {
+  string                   supplier = 1;
+  uint32                   tier     = 2;
+  uint64                   end      = 3;
```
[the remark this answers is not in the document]
toteki: Incentive programs start and end at block heights, so I have unbondings working the same way. This will prevent bursts of activity instantly from happening (potentially out of order) after any chain halts.

x/incentive/types/keys.go
```diff
@@ -0,0 +1,15 @@
...
+	ModuleName = "incentive"
+
+	// StoreKey defines the primary module store key
+	StoreKey = ModuleName
+
+	// RouterKey is the message route for slashing
+	RouterKey = ModuleName
+
+	// QuerierRoute defines the module's query routing key
+	QuerierRoute = ModuleName
```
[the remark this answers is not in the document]
toteki: From leverage module - should we review that for legacy code too?

x/incentive/types/tx.go
toteki: done

proto/umee/incentive/v1/tx.proto
toteki: done, and removed the stringers in tx.go

proto/umee/incentive/v1/tx.proto
```diff
@@ -0,0 +1,114 @@
...
+  string                   sponsor = 1;
+  uint64                   program = 2;
+  cosmos.base.v1beta1.Coin asset   = 3 [(gogoproto.nullable) = false];
+}
+
+// MsgSponsorResponse defines the Msg/Sponsor response type.
+message MsgSponsorResponse {}
+
+// MsgCreateProgram is used by governance to create an incentive program.
+message MsgCreateProgram {
```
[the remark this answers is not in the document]
toteki: Could you comment this (and any other notes) on Sai's `MsgUpdateRegistry` PR as well? Some of them will apply there probably.

proto/umee/incentive/v1/tx.proto
```diff
@@ -0,0 +1,114 @@
...
+  option (cosmos.msg.v1.signer)       = "authority";
+
+  // authority is the address of the governance account.
+  string           authority   = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+  string           title       = 2;
+  string           description = 3;
+  IncentiveProgram program     = 4 [(gogoproto.nullable) = false];
+
+  // sponsor is the address of the community fund or an address from which program rewards will be taken.
+  string sponsor = 5;
```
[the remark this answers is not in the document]
toteki: See outlined approaches in PR description

x/incentive/types/keys.go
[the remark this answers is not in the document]
robert-zaremba: yes, but in other task.

proto/umee/incentive/v1/tx.proto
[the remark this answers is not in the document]
robert-zaremba: OK, good point. will do.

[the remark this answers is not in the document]
robert-zaremba: I see you already commented.

proto/umee/incentive/v1/tx.proto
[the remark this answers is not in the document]
robert-zaremba: Let's describe it here - in the proto comment as well please.

proto/umee/incentive/v1/genesis.proto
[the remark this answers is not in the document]
robert-zaremba: To be honest, I'm not sure what's better: in case of a 3 day halt (for example) - do we want to commit and compute farming using blocks or seconds? 
I think seconds is more logical - because the time passed and stakers didn't have access to their founds. Seconds is the "real" time.

proto/umee/incentive/v1/tx.proto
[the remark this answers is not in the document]
robert-zaremba: Let's add it here - in the proto description as well, please.

proto/umee/incentive/v1/genesis.proto
```diff
@@ -0,0 +1,74 @@
...
+  string   lock_denom                               = 2;
+  repeated cosmos.base.v1beta1.DecCoin reward_basis = 4 [
+    (gogoproto.nullable)     = false,
+    (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.DecCoins"
+  ];
+}
+
+// Unbonding is a structure that tracks an in-progress token unbonding.
+message Unbonding {
+  string                   supplier = 1;
```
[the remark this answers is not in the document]
toteki: that works

proto/umee/incentive/v1/incentive.proto
```diff
@@ -0,0 +1,89 @@
...
+  // It is zero when the program is being proposed by governance.
+  uint32 id = 1;
+
+  // start_height is the block height at which the incentives begin.
+  uint64 start_height = 2 [(gogoproto.moretags) = "yaml:\"start_height\""];
+
+  // duration is the length of the incentive program in blocks.
+  uint64 duration = 3;
+
+  // lock_denom defines the denomination of the incentivized collateral uToken.
```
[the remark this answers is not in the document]
toteki: switched to `denom`

proto/umee/incentive/v1/tx.proto
```diff
@@ -0,0 +1,114 @@
...
+// MsgClaimResponse defines the Msg/Claim response type.
+message MsgClaimResponse {
+  repeated cosmos.base.v1beta1.Coin amount = 1 [
+    (gogoproto.nullable)     = false,
+    (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"
+  ];
+}
+
+// MsgLock represents a supplier's request to lock uToken collateral.
+message MsgLock {
```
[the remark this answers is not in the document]
toteki: changed to bond/unbond everywhere

x/incentive/types/params.go
toteki: Added `MsgGovSetParams`

proto/umee/incentive/v1/query.proto
```diff
@@ -0,0 +1,62 @@
...
+
+// QueryParamsResponse defines the response structure for the Params gRPC
+// service handler.
+message QueryParamsResponse {
+  Params params = 1 [(gogoproto.nullable) = false];
+}
+
+// QueryIncentiveProgramsRequest defines the request structure for the IncentivePrograms
+// gRPC service handler.
+message QueryIncentiveProgramsRequest {}
```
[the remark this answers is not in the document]
toteki: pagination added in proto - implementation will be in a separate PR.

(reference used: [x/bank's DenomOwner query and response](https://github.com/cosmos/cosmos-sdk/blob/eb1e3ebf6fcfd56e1f1d8f4c0a6192740bd8a6c1/proto/cosmos/bank/v1beta1/query.proto#L214))

proto/umee/incentive/v1/tx.proto
```diff
@@ -0,0 +1,114 @@
...
+  // Claim defines a method for claiming any pending incentive rewards.
+  rpc Claim(MsgClaim) returns (MsgClaimResponse);
+
+  // Lock defines a method for locking uToken collateral into a locking tier.
+  rpc Lock(MsgLock) returns (MsgLockResponse);
+
+  // BeginUnbonding defines a method for starting to unbond uToken collateral.
+  rpc BeginUnbonding(MsgBeginUnbonding) returns (MsgBeginUnbondingResponse);
+
+  // Sponsor defines a permissionless method for sponsoring an incentive program.
```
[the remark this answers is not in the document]
robert-zaremba: Let's add a short description here, when `MsgSponsor` is a valid request.

x/incentive/types/msg.go
```diff
@@ -0,0 +1,270 @@
...
+		Program:     program,
+		Authority:   authority,
+	}
+}
+
+func (msg MsgGovCreateProgram) ValidateBasic() error {
+	if _, err := sdk.AccAddressFromBech32(msg.Authority); err != nil {
+		return sdkerrors.Wrap(err, "invalid authority address")
+	}
+	if err := validateProposal(msg.Title, msg.Description); err != nil {
```
robert-zaremba, line 157: let's add `msg.Authority` to the arguments of `validateProposal`

robert-zaremba (commented):
We still need to decide if we want to use block time or "real" time (seconds).

x/incentive/types/msg.go
toteki: note: will switch from this function to Sai's `util/checkers` implementation

proto/umee/incentive/v1/genesis.proto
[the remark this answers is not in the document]
robert-zaremba: The motivation reduces to the same reason we have _real time_ (seconds) in staking. I don't think we should re-invent this thing.

proto/umee/incentive/v1/tx.proto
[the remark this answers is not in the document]
toteki: comments added

proto/umee/incentive/v1/genesis.proto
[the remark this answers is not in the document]
toteki: Switched to unix time for all

toteki:
Switched all block height (unbondings, durations, program ends) to unix time.

Also split up where the incentive programs are stored in state and queried, so it's easier to validate them (different values should be zero/nonzero at different points in the program's lifecycle (upcoming, ongoing, completed).

Additionally, only the completed programs query has pagination - that's where programs will accumulate (the other queries will always have a limited amount of programs in them)

proto/umee/oracle/v1/genesis.proto
```diff
@@ -22,7 +22,7 @@ message GenesisState {
       [(gogoproto.nullable) = false];
   repeated AggregateExchangeRateVote aggregate_exchange_rate_votes = 6
       [(gogoproto.nullable) = false];
-  repeated HistoricPrice               historic_prices  = 8 [(gogoproto.nullable) = false];
+  repeated HistoricPrice     historic_prices  = 8 [(gogoproto.nullable) = false];
```
toteki, line 25: Effect of `make proto-all` (the fmt step specifically)

x/leverage/types/msgs.go
```diff
@@ -44,11 +38,7 @@ func (msg MsgGovUpdateRegistry) String() string {
 
 // ValidateBasic implements Msg
 func (msg MsgGovUpdateRegistry) ValidateBasic() error {
-	if _, err := sdk.AccAddressFromBech32(msg.Authority); err != nil {
-		return sdkerrors.Wrap(err, "invalid authority address")
-	}
-
-	if err := validateProposal(msg.Title, msg.Description); err != nil {
+	if err := checkers.ValidateProposal(msg.Title, msg.Description, msg.Authority); err != nil {
```
robert-zaremba, line 41: :+1:

x/incentive/params.go
```diff
@@ -0,0 +1,100 @@
...
+func DefaultParams() Params {
+	return Params{
+		MaxUnbondings:           defaultMaxUnbondings,
+		UnbondingDurationLong:   defaultUnbondingDurationLong,
+		UnbondingDurationMiddle: defaultUnbondingDurationMiddle,
+		UnbondingDurationShort:  defaultUnbondingDurationShort,
+		TierWeightMiddle:        defaultTierWeightMiddle,
+		TierWeightShort:         defaultTierWeightShort,
+	}
+}
```
robert-zaremba, line 30: `Params` are deprecated, please use `Msg` based params system. You can seem my draft PR. 
I think, for the sake of time, you can keep this PR without params implementaiton, and we can do that separately. So it will only include params proto.

x/incentive/msgs.go
```diff
@@ -0,0 +1,257 @@
...
+	return nil
+}
+
+// Validate performs validation on an IncentiveProgram type returning an error
+// if the program is invalid.
+func (ip IncentiveProgram) Validate() error {
+	if err := sdk.ValidateDenom(ip.Denom); err != nil {
+		return err
+	}
+	if !strings.HasPrefix(ip.Denom, "u/") {
```
robert-zaremba, line 249: Don't we have a function for that in `x/leverage`? If not, let's create a common function, eg `assertIsUToken`

proto/umee/incentive/v1/genesis.proto
```diff
@@ -0,0 +1,77 @@
...
+  ];
+  repeated BondAmount        bond_amounts        = 8 [(gogoproto.nullable) = false];
+  repeated PendingReward     pending_rewards     = 9 [(gogoproto.nullable) = false];
+  repeated RewardBasis       reward_bases        = 10 [(gogoproto.nullable) = false];
+  repeated RewardAccumulator reward_accumulators = 11 [(gogoproto.nullable) = false];
+  repeated Unbonding         unbondings          = 12 [(gogoproto.nullable) = false];
+}
+
+// BondAmount tracks the amount of coins of one uToken denomination bonded to a
+// given reward tier by a single account.
```
robert-zaremba, line 30:
```suggestion
// Bond tracks the user rewards per tire and uToken coin.
```

robert-zaremba: I think we can simplify the name, otherwise we would need to call it `UserTireBond` or `UserTireCoinBond`

proto/umee/incentive/v1/genesis.proto
```diff
@@ -0,0 +1,77 @@
...
+  ];
+}
+
+// RewardBasis tracks the value of a given tier and lock denom's RewardAccumulator
+// at the last time a specific account calculated pending rewards for it.
+message RewardBasis {
+  string   account                                  = 1;
+  uint32   tier                                     = 2;
+  string   denom                                    = 3;
+  repeated cosmos.base.v1beta1.DecCoin reward_basis = 4 [
```
robert-zaremba, line 53:
```suggestion
  repeated cosmos.base.v1beta1.DecCoin reward_acc = 4 [
```

proto/umee/incentive/v1/genesis.proto
```diff
@@ -0,0 +1,77 @@
...
+    (gogoproto.nullable)     = false,
+    (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.DecCoins"
+  ];
+}
```
robert-zaremba, line 57: should we add a timestamp here as well?

proto/umee/incentive/v1/genesis.proto
```diff
@@ -0,0 +1,77 @@
...
+  string   account                                 = 1;
+  repeated cosmos.base.v1beta1.Coin pending_reward = 2 [
+    (gogoproto.nullable)     = false,
+    (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"
+  ];
+}
+
+// RewardBasis tracks the value of a given tier and lock denom's RewardAccumulator
+// at the last time a specific account calculated pending rewards for it.
+message RewardBasis {
```
robert-zaremba, line 49: what's the difference between `RewardBasis` and `RewardAccumulator`? Can we rename `RewardBasis` to something more clear and update the comment?

proto/umee/incentive/v1/incentive.proto
```diff
@@ -0,0 +1,84 @@
...
+
+  // unbonding_duration_middle defines the unbonding duration (in seconds) of the middle tier.
+  uint64 unbonding_duration_middle = 3;
+
+  // unbonding_duration_short defines the unbonding duration (in seconds) of the short tier.
+  uint64 unbonding_duration_short = 4;
+
+  // The tier_weight_short defines how the proportion of rewards assets locked
+  // in the short unbonding duration receive compared to what the same amount
+  // would receive on the long tier.
```
robert-zaremba, line 27:
```suggestion
  // would receive on the long tier.
  // valid values: [0;1]
```

proto/umee/incentive/v1/incentive.proto
```diff
@@ -0,0 +1,84 @@
...
+  string tier_weight_short = 5 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
+    (gogoproto.nullable)   = false
+  ];
+
+  // The tier_weight_middle defines how the proportion of rewards assets locked
+  // in the middle unbonding duration receive compared to what the same amount
+  // would receive on the long tier.
```
robert-zaremba, line 35:
```suggestion
  // would receive on the long tier.
  // valid values: [0;1]
```

proto/umee/incentive/v1/incentive.proto
```diff
@@ -0,0 +1,84 @@
...
+  uint32 max_unbondings = 1;
+
+  // unbonding_duration_long defines the unbonding duration (in seconds) of the long tier.
+  uint64 unbonding_duration_long = 2;
+
+  // unbonding_duration_middle defines the unbonding duration (in seconds) of the middle tier.
+  uint64 unbonding_duration_middle = 3;
+
+  // unbonding_duration_short defines the unbonding duration (in seconds) of the short tier.
+  uint64 unbonding_duration_short = 4;
```
robert-zaremba, line 23: I think the values above should be immutable, yes? Let's clarify it in the comment and later assert it when we will implement the keeper.

proto/umee/incentive/v1/incentive.proto
```diff
@@ -0,0 +1,84 @@
...
+// IncentiveProgram defines a liquidity mining incentive program on a single
+// locked uToken denom that will run for a set amount of time.
+message IncentiveProgram {
+  option (gogoproto.equal) = true;
+
+  // ID uniquely identifies the incentive program after it has been created.
+  // It is zero when the program is being proposed by governance.
+  uint32 id = 1;
+
+  // start_time is the block time at which the incentives begin.
```
robert-zaremba, line 51:
```suggestion
  // start_time is a timestamp (in seconds) at which the incentives begin.
```

robert-zaremba: We don't know the block time in advance. So it should be either a timestamp or a block height.

proto/umee/incentive/v1/incentive.proto
```diff
@@ -0,0 +1,84 @@
...
+  uint64 start_time = 2;
+
+  // duration is the length of the incentive program in seconds.
+  uint64 duration = 3;
+
+  // denom defines the denomination of the incentivized collateral uToken.
```
robert-zaremba, line 57:
```suggestion
  // denom is a uToken incentivized collateral.
```

proto/umee/incentive/v1/genesis.proto
```diff
@@ -0,0 +1,77 @@
...
+option go_package = "github.com/umee-network/umee/v3/x/incentive";
+
+import "cosmos/base/v1beta1/coin.proto";
+import "gogoproto/gogo.proto";
+import "umee/incentive/v1/incentive.proto";
+
+// GenesisState defines the x/incentive module's genesis state.
+message GenesisState {
+  Params                    params               = 1 [(gogoproto.nullable) = false];
+  repeated IncentiveProgram completed_programs   = 2 [(gogoproto.nullable) = false];
```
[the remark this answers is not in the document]
robert-zaremba: hmm, maybe we can add it to v2, once we will have proper indexer?

proto/umee/incentive/v1/tx.proto
```diff
@@ -0,0 +1,134 @@
...
+  rpc Claim(MsgClaim) returns (MsgClaimResponse);
+
+  // Bond defines a method for bonding uToken collateral into reward tier.
+  rpc Bond(MsgBond) returns (MsgBondResponse);
+
+  // BeginUnbonding defines a method for starting to unbond uToken collateral.
+  rpc BeginUnbonding(MsgBeginUnbonding) returns (MsgBeginUnbondingResponse);
+
+  // Sponsor defines a permissionless method for sponsoring an upcoming, not yet funded incentive program.
+  rpc Sponsor(MsgSponsor) returns (MsgSponsorResponse);
```
robert-zaremba, line 24: I think w should enable partial sponsoring, when 2 separate sponsors will like to sponsor a program. Or do it in v2.

proto/umee/incentive/v1/tx.proto
```diff
@@ -0,0 +1,134 @@
...
+
+  // GovSetParams is used by governance proposals to update parameters.
+  rpc GovSetParams(MsgGovSetParams) returns (MsgGovSetParamsResponse);
+
+  // GovCreateProgram is used by governance proposals to create (but not fund) an incentive program.
+  rpc GovCreateProgram(MsgGovCreateProgram) returns (MsgGovCreateProgramResponse);
+
+  // GovCreateAndSponsorProgram is used by governance proposals to create and fund an incentive program.
```
robert-zaremba, line 32:
```suggestion
  // GovCreateAndSponsorProgram is used by governance proposals to create an incentive program 
  // and sponsor it from the Community Fund.
```

proto/umee/incentive/v1/tx.proto
```diff
@@ -0,0 +1,134 @@
...
+  IncentiveProgram program     = 4 [(gogoproto.nullable) = false];
+}
+
+// MsgGovCreateProgramResponse defines the Msg/CreateProgram response type.
+message MsgGovCreateProgramResponse {}
+
+// MsgGovCreateAndSponsorProgram is used by governance to create an incentive program
+// and fund it in the same transaction. The sponsor address should be the community
+// fund or an agreed upon account - program reward funds will be taken from that address
+// when the incentive program passes governance.
```
robert-zaremba, line 116: there is no sponsor parameter in this message. Let's just write that "when passed, the program is fully sponsored by the Community Fund".

robert-zaremba (commented):
- Usually, the best is to keep the first PR only proto oriented + generated files. This way we firstly deal with proto, without messing with validation etc... But the validation is already here, so we can go with that. So let's make the review in mulitple stage, here I firstly review the proto.

1. Let's don't use `params` - it's officially deprecated. See my comment how to move forward with that.
2. Let's add a `create_program_fee` param to avoid a spam (maybe we can discuss it in notion). This will be paid in Umee.
3. Please create a notion page to specify functionality (not a design doc). Specifically:
    * how the programs will work with uTokens? 
    * Can there be multiple programs for the same uToken with overlapping period? 
    * Do we support auto compounding (now, or v2?)? 
    * Can programs be extended or resumed?
    * Can we partially fund a program?
    * Maybe we should support a scenario when community fund will only partially sponsor a program? In that case we would need to add `ratio` parameter to `MsgGovCreateAndSponsorProgram`

proto/umee/incentive/v1/query.proto
```diff
@@ -0,0 +1,104 @@
...
+      returns (QueryOngoingIncentiveProgramsResponse) {
+    option (google.api.http).get = "/umee/incentive/v1/incentive_programs/ongoing";
+  }
+
+  // UpcomingIncentivePrograms queries for all incentives programs that have been passed
+  // by governance, but not yet started. They may or may not have been funded.
+  rpc UpcomingIncentivePrograms(QueryUpcomingIncentivePrograms)
+      returns (QueryUpcomingIncentiveProgramsResponse) {
+    option (google.api.http).get = "/umee/incentive/v1/incentive_programs/upcoming";
+  }
```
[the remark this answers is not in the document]
toteki: Frontend will probably just need current, and it's easier to have them stored separately in state since it reduces iterations when looking for just one type.

proto/umee/incentive/v1/tx.proto
toteki: If so, I'll need to hardcode (or add to `Params`) a community fund address.

proto/umee/incentive/v1/genesis.proto
toteki: `RewardAccumulator` is global:
> `1u/ATOM` bonded to tier `3` at genesis would have accumulated `0.04ATOM and 0.01UMEE` incentive rewards. This increases every block when programs are ongoing.

`RewardBasis` is local:
> The last time account `umee1234...` bonded, began unbonding, was liquidated, or claimed rewards for `u/ATOM` at tier 3, the value of the `u/ATOM RewardAccumulator` was `0.03ATOM and 0.01UMEE`. Therefore it the user were to claim rewards now, their tier 3 bonded `u/ATOM` will give additional rewards (0.03 -> 0.04).
This is only modified when the user does one of the transactions listed above, which is required to keep reward calculations mathematically correct.

proto/umee/incentive/v1/incentive.proto
toteki: I see no problems with keeping it as a parameter as originally intended

proto/umee/incentive/v1/genesis.proto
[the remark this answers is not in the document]
toteki: Since each one must be passed by governance and the `IncentiveProgram` structure is small, I don't think storing them in state is too threatening. It's better to have a way to audit what programs have existed in the past.

x/incentive/params.go
toteki: Here I define a `Params` structure and a `MsgGovSetParams`.

No deprecated subspaces are used, so is there any problem with grouping them into a structure called `Params`?

x/incentive/msgs.go
toteki: We have one - originally, I was worried about an import cycle between leverage an incentive types and did not want to use `leveragetypes.HasUTokenPrefix`. I'll use it for now, since a cycle doesn't seem likely yet.

proto/umee/incentive/v1/genesis.proto
toteki: no - the previous value of the accumulator is sufficient

proto/umee/incentive/v1/incentive.proto
```diff
@@ -0,0 +1,84 @@
...
+  string denom = 4;
+
+  // total_rewards are total amount of rewards which can be distributed to
+  // suppliers by this program. This is set to its final value when the program
+  // is proposed by governance.
+  cosmos.base.v1beta1.Coin total_rewards = 5 [
```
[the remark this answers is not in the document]
toteki: per the design: single incentive denom per IncentiveProgram

proto/umee/incentive/v1/incentive.proto
```diff
@@ -0,0 +1,84 @@
...
+    (gogoproto.nullable)     = false,
+    (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"
+  ];
+
+  // funded_rewards are total amount of rewards which have been funded by a
+  // sponsor to this program. This is zero until the program is both passed
+  // by governance and funded.
```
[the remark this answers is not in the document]
toteki: This is disallowed - such a MsgSponsor would be rejected

proto/umee/incentive/v1/incentive.proto
```diff
@@ -0,0 +1,84 @@
...
+  cosmos.base.v1beta1.Coin funded_rewards = 6 [
+    (gogoproto.nullable)     = false,
+    (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"
+  ];
+
+  // remaining_rewards are total amount of this program's funded rewards
+  // which have not yet been allocated to suppliers. This is zero until the
```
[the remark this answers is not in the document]
toteki: I don't think this wording increases clarity

proto/umee/incentive/v1/tx.proto
toteki: Note: sponsor parameter is currently below the highlighted comment - it is the last field

proto/umee/incentive/v1/genesis.proto
```diff
@@ -0,0 +1,77 @@
...
+  uint64                    last_rewards_time    = 6;
+  repeated cosmos.base.v1beta1.Coin total_bonded = 7 [
+    (gogoproto.nullable)     = false,
+    (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"
+  ];
+  repeated BondAmount        bond_amounts        = 8 [(gogoproto.nullable) = false];
+  repeated PendingReward     pending_rewards     = 9 [(gogoproto.nullable) = false];
+  repeated RewardBasis       reward_bases        = 10 [(gogoproto.nullable) = false];
+  repeated RewardAccumulator reward_accumulators = 11 [(gogoproto.nullable) = false];
+  repeated Unbonding         unbondings          = 12 [(gogoproto.nullable) = false];
```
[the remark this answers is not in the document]
robert-zaremba: yes, I was suppose to remove this comment. Initially I thought that anyone can just create a new proposal, but this is a gov propsal

proto/umee/incentive/v1/query.proto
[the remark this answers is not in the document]
robert-zaremba: 1. So you are moving the program between storage prefixes?
2. My question was more related to the query, why having 3 queries if we can only have 1 or 2? That would reduce and simplify the code. 
3. Let's verify with the frontend. My gut feeling is that they will like to show both present and future.

proto/umee/incentive/v1/tx.proto
robert-zaremba: Yes, but maybe we need to support a scenario when, for example we have 2 sponsors?

proto/umee/incentive/v1/tx.proto
robert-zaremba: You check the Community Fund address during the execution. It's not a param, because it doesn't change.

x/incentive/params.go
robert-zaremba: It's good. I made the comment at the beginning of the review, and forgot to remove. Thanks for responding though.

proto/umee/incentive/v1/genesis.proto
robert-zaremba: So how about renaming `Basis` to `Local`?

proto/umee/incentive/v1/tx.proto
brandoncomer: Something that could be useful in the future, but probably not near-term. Possible use case would be ecosystems with multiple tokens like Comdex. Maybe they want to incentivize borrowing their CMST stablecoin with CMDX, HARBOR, etc.

proto/umee/incentive/v1/incentive.proto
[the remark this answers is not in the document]
robert-zaremba: Let's clarify this in a comment (or maybe firstly confirm in Slack if we are OK with this behavior)

proto/umee/incentive/v1/tx.proto
robert-zaremba: yes, I was thinking about similar scenario. Other would be if there are two independent accounts which would sponsor only if the proposal passes (Eg: Umee DAO and Osmosis DAO).

PR #1535, opened 2022-10-30
fix: enable revive linter in golangci

toteki:
## Description

- Enabled the recommended list of `revive` rules in `.golangci.yml` - it looks like it was running with an empty rule list.
- Fixed `make lint` after revive linter was enabled to not require a `go mod tidy`.
- Pinned tool versions
- Disabled repetitive warnings from `exported` and `package-comments` rules.
- Ensured revive `errors` fail, but `warnings` pass, the CI job
- Commented out currently failing rules. Follow-up issue: #1536

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


.github/workflows/lint.yml
```diff
@@ -25,6 +25,8 @@ jobs:
       - name: revive lint
         uses: morphy2k/revive-action@v2
         with:
+          config: .revive.toml
+          exclude: "*.pb.go;*.pb.gw.go;*.pulsar.go;swagger/..."
           path: "./..."
```
toteki, line 30: This step generates annotations on files in pull requests, and mirrors the local output of `make lint`.

Neither revive `warnings` nor `errors` will appear as a CI fail though, so any rules we want to require should appear in the `golangci.yml` config.

.golangci.yml
```diff
@@ -60,14 +60,40 @@ issues:
...
+      - name: range
+      # - name: receiver-naming
+      # - name: redefines-builtin-id
+      - name: superfluous-else
+      - name: time-naming
+      - name: unexported-return
+      - name: unreachable-code
+      # - name: unused-parameter
+      - name: var-declaration
+      # - name: var-naming
```
toteki, line 99: These need to be fixed and commented in - CI will fail if they generate errors (but not warnings).

.revive.toml
```diff
@@ -0,0 +1,66 @@
...
+[rule.empty-block]
+[rule.error-return]
+[rule.error-strings]
+[rule.error-naming]
+[rule.errorf]
+#[rule.exported]
+[rule.if-return]
+[rule.increment-decrement]
+[rule.indent-error-flow]
+#[rule.package-comments]
```
toteki, line 24: There are hundreds of "exported var needs comment" or "package needs comment" annotations. I'm disabling them to reduce noise.

.revive.toml
```diff
@@ -0,0 +1,66 @@
...
+[rule.struct-tag]
+[rule.time-equal]
+[rule.unconditional-recursion]
+#[rule.unexported-naming]
+#[rule.unhandled-error]
+[rule.unnecessary-stmt]
+#[rule.unused-receiver]
+#[rule.use-any]
+#[rule.useless-break]
+[rule.waitgroup-by-value]
```
toteki, line 60: The rules commented out in this section have active issues, but look like they would be reasonable to fix.

.revive.toml
```diff
@@ -0,0 +1,66 @@
...
+
+# Rules that we might want to enable
+#[rule.bare-return]
+#[rule.deep-exit]
+#[rule.defer]
+#[rule.import-shadowing]
```
toteki, line 66: These rules might not be worth enabling, but feel free to check out the `make lint` output they produce when turned on.

Makefile
```diff
@@ -150,37 +150,33 @@ else
...
 	@go install github.com/mgechev/revive
-	@revive ./...
+	@revive -config .revive.toml -formatter friendly ./...
 
 lint-fix:
 	@echo "--> Running linter to fix the lint issues"
-	@go install mvdan.cc/gofumpt@latest
-	@go install github.com/golangci/golangci-lint/cmd/golangci-lint@$(golangci_version)
+	@go install github.com/mvdan/gofumpt
+	@go install github.com/golangci/golangci-lint/cmd/golangci-lint
```
toteki, line 173: Note on `go install`:

A version suffix like `@latest` or `@v1.49.0` will ignore the version of a tool tracked in `go.mod`, but I observed that there were still some `go.sum` changes after `make lint` in practice.

Now the tool versions are pinned in `tools.go` and no `go.sum` instability occurs.

RafilxTenfen (approved):

.golangci.yml
robert-zaremba: Why do we have it here an in `.revive.toml`? Can't CI use the revive.toml?

toteki: In my initial searches I didn't see a way to have them use the same config - also, I'm enabling many more optional linters in `.revive.toml` for annotations that don't need to fail `golangci-lint`, so it might be helpful to keep the lists separate.

Issue #1540, opened 2022-10-31
Add keeper getters and setters for historacle objects


adamewozniak:
For historic pricing we will have to store: 

* The running medians of prices of all assets ( map[asset] -> value )
* Historic prices ( map [asset][block] -> value )
* New parameters ( StampPeriod, PrunePeriod )

We need to create getters and setters in the oracle keeper so that our endblockers and queries can interface with this

PR #1548, opened 2022-11-04
feat: Add keeper methods for historacle prices and medians

rbajollari:
## Description

closes: #1540

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/oracle/keeper/historic_price.go
```diff
@@ -0,0 +1,201 @@
...
+func median(prices []types.HistoricPrice) sdk.Dec {
+	lenPrices := len(prices)
+	if lenPrices == 0 {
+		return sdk.ZeroDec()
+	}
+
+	sort.Slice(prices, func(i, j int) bool {
+		return prices[i].ExchangeRates.ExchangeRate.BigInt().
+			Cmp(prices[j].ExchangeRates.ExchangeRate.BigInt()) > 0
+	})
```
adamewozniak, line 23: So we sort by rates, but what if the rates are the same? Is it randomized order?

x/oracle/keeper/historic_price.go
```diff
@@ -0,0 +1,201 @@
...
+
+	var medianPrice sdk.Dec
+	if lenPrices%2 == 0 {
```
adamewozniak, line 26:
```suggestion
	if lenPrices % 2 == 0 {
```

x/oracle/keeper/historic_price.go
```diff
@@ -0,0 +1,201 @@
...
+	block := uint64(ctx.BlockHeight())
+	bz := k.cdc.MustMarshal(&types.HistoricPrice{
+		ExchangeRates: exchangeRateTuple,
+		BlockNum:      block,
+	})
+	denom = strings.ToUpper(denom)
+	store.Set(types.GetHistoricPriceKey(denom, block), bz)
+
+	// update median after adding new historic price
+	k.setMedian(ctx, denom)
```
adamewozniak, line 165: You'll want to pass in the exchangeRate rather than querying for it again in `setMedian`
```suggestion
	k.setMedian(ctx, denom, exchangeRate)
```

x/oracle/keeper/historic_price.go
```diff
@@ -0,0 +1,201 @@
...
+
+	return historicPrice, nil
+}
+
+// GetHistoricPrices returns all the historic prices of a given denom.
+func (k Keeper) GetHistoricPrices(
+	ctx sdk.Context,
+	denom string,
+) []types.HistoricPrice {
+	denom = strings.ToUpper(denom)
```
adamewozniak, line 97: This will be consumed by ABCI endblocker where they're already uppercase, or through the querier, where we can just set it to uppercase. No need to put it in this keeper imo 
```suggestion
```

x/oracle/keeper/historic_price.go
```diff
@@ -0,0 +1,201 @@
...
+	handler func(sdk.Dec, uint64) bool,
+) {
+	store := ctx.KVStore(k.storeKey)
+
+	iter := sdk.KVStorePrefixIterator(store, types.KeyPrefixHistoricPrice)
+	defer iter.Close()
+
+	for ; iter.Valid(); iter.Next() {
+		key := iter.Key()
+		historicPriceDenom := string(key[len(types.KeyPrefixExchangeRate) : len(key)-9])
```
adamewozniak, line 129: You might want to put this denom parser in a util function

x/oracle/keeper/historic_price.go
```diff
@@ -0,0 +1,201 @@
...
+		if historicPriceDenom != denom {
+			continue
+		}
+		historicPrice := types.HistoricPrice{}
+
+		k.cdc.MustUnmarshal(iter.Value(), &historicPrice)
+		if handler(historicPrice.ExchangeRates.ExchangeRate, historicPrice.BlockNum) {
```
adamewozniak, line 136: This is a bit repetitive, in another PR do you think it would make sense to make the `HistoricPrice` struct look like this?

```golang
type HistoricPrice struct {
	Rate ExchangeRateTuple
	BlockNum      uint64
}
```

x/oracle/keeper/historic_price.go
```diff
@@ -0,0 +1,201 @@
...
+	blockNum uint64,
+) {
+	store := ctx.KVStore(k.storeKey)
+	store.Delete(types.GetHistoricPriceKey(denom, blockNum))
+
+	// update median after deleting historic price
+	k.setMedian(ctx, denom)
+}
+
+func (k Keeper) ClearHistoricPrices(ctx sdk.Context) {
```
adamewozniak, line 182: Keeper comment please 😄 

Also, we'll have to prune historic prices after a certain amount of time rather than just clearing all of them out, so idk if we need this. We can add that keeper function in another PR

x/oracle/keeper/params.go
```diff
@@ -63,6 +63,32 @@ func (k Keeper) MinValidPerWindow(ctx sdk.Context) (res sdk.Dec) {
 	return
 }
 
+// StampPeriod returns the amount of blocks the historacle module waits
```
adamewozniak, line 66:
```suggestion
// StampPeriod returns the amount of blocks the historacle endblocker waits
```

x/oracle/keeper/params.go
```diff
@@ -63,6 +63,32 @@ func (k Keeper) MinValidPerWindow(ctx sdk.Context) (res sdk.Dec) {
...
+// between recording a set of prices.
+func (k Keeper) StampPeriod(ctx sdk.Context) (res uint64) {
+	k.paramSpace.Get(ctx, types.KeyStampPeriod, &res)
+	return
+}
+
+// SetStampPeriod updates the amount of blocks the historacle module waits
```
adamewozniak, line 73:
```suggestion
// SetStampPeriod updates the amount of blocks the historacle endblocker waits
```

Same here, technically historacle isn't a module of its own ^

x/oracle/types/params.go
```diff
@@ -150,6 +166,10 @@ func (p Params) Validate() error {
 		return fmt.Errorf("oracle parameter MinValidPerWindow must be between [0, 1]")
 	}
 
+	if p.PrunePeriod < p.StampPeriod {
```
adamewozniak, line 169: perfect 💯

x/oracle/types/params.go
```diff
@@ -285,3 +305,29 @@ func validateMinValidPerWindow(i interface{}) error {
...
+
+func validatePrunePeriod(i interface{}) error {
+	v, ok := i.(uint64)
+	if !ok {
+		return fmt.Errorf("invalid parameter type: %T", i)
+	}
+
+	if v == 0 {
+		return fmt.Errorf("prune period must be positive: %d", v)
+	}
```
adamewozniak, line 330:
```suggestion
	if v < 1 {
		return fmt.Errorf("stamp period must be positive: %d", v)
	}

	return nil
}

func validatePrunePeriod(i interface{}) error {
	v, ok := i.(uint64)
	if !ok {
		return fmt.Errorf("invalid parameter type: %T", i)
	}

	if v < 1 {
		return fmt.Errorf("prune period must be positive: %d", v)
	}
```

adamewozniak (commented):
initial review - great job @rbajollari !!

x/oracle/keeper/historic_price.go
```diff
@@ -0,0 +1,201 @@
...
+	var medianPrice sdk.Dec
+	if lenPrices%2 == 0 {
+		medianPrice = prices[lenPrices/2-1].ExchangeRates.ExchangeRate.
+			Add(prices[lenPrices/2].ExchangeRates.ExchangeRate).
+			QuoInt64(2)
+	} else {
+		medianPrice = prices[lenPrices/2].ExchangeRates.ExchangeRate
+	}
+
+	return medianPrice
```
robert-zaremba, line 34:
```suggestion
	if lenPrices%2 == 0 {
		return prices[lenPrices/2-1].ExchangeRates.ExchangeRate.
			Add(prices[lenPrices/2].ExchangeRates.ExchangeRate).
			QuoInt64(2)
	}
	return prices[lenPrices/2].ExchangeRates.ExchangeRate
```

x/oracle/keeper/historic_price.go
```diff
@@ -0,0 +1,201 @@
...
+func (k Keeper) setMedian(
+	ctx sdk.Context,
+	denom string,
+) {
+	store := ctx.KVStore(k.storeKey)
+
+	historicPrices := k.GetHistoricPrices(ctx, denom)
+
+	median := median(historicPrices)
+
```
robert-zaremba, line 66:
```suggestion
	store := ctx.KVStore(k.storeKey)
	historicPrices := k.GetHistoricPrices(ctx, denom)
	median := median(historicPrices)
```

x/oracle/keeper/historic_price.go
```diff
@@ -0,0 +1,201 @@
...
+	ctx sdk.Context,
+	denom string,
+) []types.HistoricPrice {
+	denom = strings.ToUpper(denom)
+	historicPrices := []types.HistoricPrice{}
+
+	k.IterateHistoricPrices(ctx, denom, func(exchangeRate sdk.Dec, blockNum uint64) bool {
+		historicPrices = append(historicPrices, types.HistoricPrice{
+			ExchangeRates: types.ExchangeRateTuple{
+				Denom:        denom,
```
robert-zaremba, line 103: Why don't we create a list of exchange rates only? We don't need to repeat `denom` in all elements.

x/oracle/keeper/historic_price.go
```diff
@@ -0,0 +1,201 @@
...
+	k.cdc.MustUnmarshal(bz, &historicPrice)
+
+	return historicPrice, nil
+}
+
+// GetHistoricPrices returns all the historic prices of a given denom.
+func (k Keeper) GetHistoricPrices(
+	ctx sdk.Context,
+	denom string,
+) []types.HistoricPrice {
```
robert-zaremba, line 96: pagination must be added, otherwise it will blow up the tx time.

x/oracle/keeper/historic_price.go
```diff
@@ -0,0 +1,201 @@
...
+	defer iter.Close()
+
+	for ; iter.Valid(); iter.Next() {
+		key := iter.Key()
+		historicPriceDenom := string(key[len(types.KeyPrefixExchangeRate) : len(key)-9])
+		if historicPriceDenom != denom {
+			continue
+		}
+		historicPrice := types.HistoricPrice{}
+
```
robert-zaremba, line 134:
```suggestion
		var historicPrice types.HistoricPrice
```

x/oracle/keeper/historic_price.go
```diff
@@ -0,0 +1,201 @@
...
+		})
+
+		return false
+	})
+
+	return historicPrices
+}
+
+// IterateHistoricPrices iterates over historic prices of a given
+// denom in the store.
```
robert-zaremba, line 116:
```suggestion
// denom in the store.
// Iterator stops when exhausting the source, or when the handler returns `true`.
```

x/oracle/keeper/historic_price.go
```diff
@@ -0,0 +1,201 @@
...
+// block height when called to the store. Afterwards it will call
+// setMedian to update the median price of the denom in the store.
+func (k Keeper) AddHistoricPrice(
+	ctx sdk.Context,
+	denom string,
+	exchangeRate sdk.Dec,
+) {
+	store := ctx.KVStore(k.storeKey)
+	exchangeRateTuple := types.ExchangeRateTuple{
+		Denom:        denom,
```
robert-zaremba, line 152: why storing `denom`? It's already in the key. Let's avoid bloating the state.

x/oracle/types/keys.go
```diff
@@ -55,3 +59,20 @@ func GetAggregateExchangeRateVoteKey(v sdk.ValAddress) (key []byte) {
 	key = append(key, KeyPrefixAggregateExchangeRateVote...)
 	return append(key, address.MustLengthPrefix(v)...)
 }
+
+// GetMedianKey - stored by *denom*
+func GetMedianKey(denom string) (key []byte) {
```
robert-zaremba, line 64:
```suggestion
func MedianKey(denom string) (key []byte) {
```

robert-zaremba: We are not getting anything from anywhere.

x/oracle/types/keys.go
```diff
@@ -55,3 +59,20 @@ func GetAggregateExchangeRateVoteKey(v sdk.ValAddress) (key []byte) {
...
+	key = append(key, KeyPrefixMedian...)
+	key = append(key, []byte(denom)...)
+	return append(key, 0) // append 0 for null-termination
+}
+
+// GetHistoricPriceKey - stored by *denom* and *block*
+func GetHistoricPriceKey(denom string, blockNum uint64) (key []byte) {
```
robert-zaremba, line 71:
```suggestion
func HistoricPriceKey(denom string, blockNum uint64) (key []byte) {
```

x/oracle/keeper/historic_price.go
```diff
@@ -0,0 +1,201 @@
...
+	store := ctx.KVStore(k.storeKey)
+	iter := sdk.KVStorePrefixIterator(store, types.KeyPrefixHistoricPrice)
+	defer iter.Close()
+	for ; iter.Valid(); iter.Next() {
+		store.Delete(iter.Key())
+	}
+
+	// clear medians as well
+	k.clearMedians(ctx)
+}
```
robert-zaremba, line 192: We should have an option to efficiently delete prices older then some time (eg older than 2 weeks ago). Otherwise this will cause significant storage use.

x/oracle/keeper/params.go
robert-zaremba, line 66:
```suggestion
// StampPeriod returns the amount of blocks the oracle module waits
```

x/oracle/keeper/params.go
robert-zaremba, line 73:
```suggestion
// SetStampPeriod updates the amount of blocks the oracle module waits
```

x/oracle/types/params.go
```diff
@@ -18,13 +18,17 @@ var (
...
+	KeyPrunePeriod              = []byte("PrunePeriod")
 )
 
 // Default parameter values
 const (
 	DefaultVotePeriod               = BlocksPerMinute / 2 // 30 seconds
 	DefaultSlashWindow              = BlocksPerWeek       // window for a week
 	DefaultRewardDistributionWindow = BlocksPerYear       // window for a year
+	DefaultStampPeriod              = BlocksPerDay        // window for a day
+	DefaultPrunePeriod              = BlocksPerMonth      // window for a month
```
robert-zaremba, line 31: What if a block time will change? It can be updated without changing a code, through a gov voting.

x/oracle/keeper/historic_price.go
robert-zaremba, line 96: pagination must be added, otherwise it will blow up the tx time.

x/oracle/keeper/historic_price.go
robert-zaremba, line 165: this is super inefficient. You read all historical prices and recalculate median during every price update.

robert-zaremba (changes requested):
Blocking because:
* adding historical price is super inefficient (see comment)
* have concerns about state bloat: how the state will be managed and cleaned?
* why do we have one big median? Shouldn't we have median per "epoch"?

Is there any spec clarifying the pruning functionality? If not, I can help with that if needed.

x/oracle/keeper/historic_price.go
robert-zaremba: Median is median. It's OK to have same values in a sequence.

x/oracle/keeper/historic_price.go
rbajollari: I think it already is like that. Do you mean like this?
```golang
type HistoricPrice struct {
	Rate          sdk.Dec
	BlockNum      uint64
}
```

x/oracle/keeper/historic_price.go
rbajollari: I put ParseDemonFromHistoricPriceKey and ParseBlockFromHistoricPriceKey as their own methods in keys.go rather than in the utils folder since these don't seem general enough to belong in there. Lmk what you think about that.

x/oracle/keeper/historic_price.go
rbajollari: Removed for now and just going to keep the DeleteHistoricPrice method for this PR

x/oracle/keeper/historic_price.go
adamewozniak: Yeah, if the denom is a part of the key we don't need to read this

x/oracle/keeper/historic_price.go
rbajollari: This is now changed to not get called inside the keeper and be used like outlined in the spec. SetMedian will get called every MedianPeriod rather than every time a new price is added.

x/oracle/keeper/historic_price.go
rbajollari: Proto changed to be just exchangerate rather than tuple 👍

x/oracle/types/keys.go
rbajollari: While I don't disagree with the naming update the rest of the methods in this file have "Get" in front of them, so updating just these two methods would look inconsistent. I'd suggest doing this in another PR.

x/oracle/types/params.go
adamewozniak: Then we'll have to change these definitions for the entire oracle module, that's not a part of this PR

x/oracle/keeper/historic_price.go
adamewozniak: this `GetHistoricPrices` method will likely only be called internally. @rbajollari wdyt?

rbajollari: Updated to only iterate price keys with the given denom 👍

x/oracle/types/keys.go
adamewozniak: I agree @rbajollari

proto/umee/oracle/v1/oracle.proto
```diff
@@ -107,9 +111,10 @@ message ExchangeRateTuple {
 
 // HistoricPrice is an instance of a price "stamp"
 message HistoricPrice {
-  ExchangeRateTuple exchange_rates = 1 [
-    (gogoproto.castrepeated) = "ExchangeRateTuples",
-    (gogoproto.nullable)     = false
+  ExchangeRateTuple exchange_rate = 1 [
+    (gogoproto.moretags)   = "yaml:\"exchange_rate\"",
```
adamewozniak, line 115: We don't need yaml tags anymore:
```suggestion
```

proto/umee/oracle/v1/oracle.proto
```diff
@@ -47,6 +47,10 @@ message Params {
   // Prune Period represents the maximum amount of blocks which we want
   // to keep a record of our set of exchange rates.
   uint64 prune_period = 10;
+  // Medion Period reperesents the amount blocks we will wait between
```
adamewozniak, line 50:
```suggestion
  // Median Period represents the amount blocks we will wait between
```

x/oracle/keeper/historic_price.go
```diff
@@ -0,0 +1,224 @@
...
+// period since a given block.
+func (k Keeper) GetMedian(
+	ctx sdk.Context,
+	denom string,
+	blockNum uint64,
+) (sdk.Dec, error) {
+	store := ctx.KVStore(k.storeKey)
+	bz := store.Get(types.GetMedianKey(denom, blockNum))
+	if bz == nil {
+		return sdk.ZeroDec(), sdkerrors.Wrap(types.ErrUnknownDenom, denom)
```
adamewozniak, line 58: It might make more sense to do a new error message here, since initially medians will not exist until the first `Median Period` passes

x/oracle/keeper/historic_price.go
```diff
@@ -0,0 +1,224 @@
...
+// its median price in the last prune period since a given block.
+func (k Keeper) GetMedianDeviation(
+	ctx sdk.Context,
+	denom string,
+	blockNum uint64,
+) (sdk.Dec, error) {
+	store := ctx.KVStore(k.storeKey)
+	bz := store.Get(types.GetMedianDeviationKey(denom, blockNum))
+	if bz == nil {
+		return sdk.ZeroDec(), sdkerrors.Wrap(types.ErrUnknownDenom, denom)
```
adamewozniak, line 93: Same here, might want another error message

x/oracle/keeper/historic_price.go
```diff
@@ -0,0 +1,224 @@
...
+) {
+	store := ctx.KVStore(k.storeKey)
+	medianDeviation := medianDeviation(median, prices)
+	bz := k.cdc.MustMarshal(&sdk.DecProto{Dec: medianDeviation})
+	store.Set(types.GetMedianDeviationKey(denom, uint64(ctx.BlockHeight())), bz)
+}
+
+// GetHistoricPrice returns the historic price of a denom at a given
+// block.
+func (k Keeper) GetHistoricPrice(
```
adamewozniak, line 118: Does this need to be a public keeper function?
```suggestion
func (k Keeper) getHistoricPrice(
```

x/oracle/keeper/historic_price.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+
+	var historicPrice types.HistoricPrice
+	k.cdc.MustUnmarshal(bz, &historicPrice)
+
+	return historicPrice, nil
+}
+
+// GetHistoricPrices returns all the historic prices of a given denom.
+func (k Keeper) GetHistoricPrices(
```
adamewozniak, line 136: Same
```suggestion
func (k Keeper) getHistoricPrices(
```

x/oracle/keeper/historic_price.go
```diff
@@ -0,0 +1,224 @@
...
+		k.cdc.MustUnmarshal(iter.Value(), &historicPrice)
+		if handler(historicPrice.ExchangeRate, historicPrice.BlockNum) {
+			break
+		}
+	}
+}
+
+// AddHistoricPrice adds the historic price of a denom at the current
+// block height when called to the store. Afterwards it will call
+// setMedian to update the median price of the denom in the store.
```
adamewozniak, line 178:
```suggestion
// AddHistoricPrice adds the historic price of a denom at the current
// block height.
```

x/oracle/keeper/historic_price.go
```diff
@@ -0,0 +1,224 @@
...
+func (k Keeper) DeleteHistoricPrice(
+	ctx sdk.Context,
+	denom string,
+	blockNum uint64,
+) {
+	store := ctx.KVStore(k.storeKey)
+	store.Delete(types.GetHistoricPriceKey(denom, blockNum))
+}
+
+// DeleteMedian deletes a givem denom's median price in the last prune
```
adamewozniak, line 204:
```suggestion
// DeleteMedian deletes a given denom's median price in the last prune
```

adamewozniak (approved):
this looks good to me, just a few comments. 

I'd like to have a review from @zarazan , and maybe @toteki or @RafilxTenfen as well 🙏

x/oracle/keeper/historic_price.go
rbajollari: This method isn't called within the keeper so since it doesn't need to be public I've removed it

adamewozniak:
waiting for a review from @RafilxTenfen to merge 🙏

x/oracle/types/keys.go
```diff
@@ -55,3 +60,41 @@ func GetAggregateExchangeRateVoteKey(v sdk.ValAddress) (key []byte) {
...
 }
+
+// GetMedianKey - stored by *denom* and *block*
+func GetMedianKey(denom string, blockNum uint64) (key []byte) {
+	key = append(key, KeyPrefixMedian...)
+	key = append(key, []byte(denom)...)
+	block := make([]byte, 8)
+	binary.LittleEndian.PutUint64(block, blockNum)
+	key = append(key, block...)
+	return append(key, 0) // append 0 for null-termination
```
RafilxTenfen, line 71: We do this in the same way 3 times, maybe add a function to it?

RafilxTenfen (approved):
Just one comment
utACK :100:

x/oracle/keeper/historic_price.go
adamewozniak: Will be addressed in #1570

Issue #1521, opened 2022-10-24
Epic: Historacle pricing


adamewozniak:
## Summary

As an addition to the oracle module, we need to:

1. Stamp prices for every assets every X amount of blocks
2. Provide a keeper function & query that responds with the median of the last 30 days of prices
3. Provide a keeper function that takes a denomination and a price, and will return whether or not that price is within the standard deviation of the median of the last 30 days of prices

This should serve as an epic for some smaller tasks:

- [x] https://github.com/umee-network/umee/pull/1525
- [x] #1540
- [x] #1598 
- [x] #1571
- [x] #1541
- [x] #1542
- [x] #1543
- [x] #1545
- [x] #1544
- [x] #1619 
- [x] #1570
- [x] #1621 
- [ ] #1655 
---

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

toteki:
I can make this into a submodule of leverage, so the code stays fairly isolated, until we figure out what to do with it. It'll just
- Store prices every epoch
- Calculate and store median and SD each epoch, per asset
- Get stored values using a keeper method + query

That way leverage can, for every `MsgBorrow` and possibly `MsgDecollateralize` and `MsgWithdraw`, execute some custom logic.

adamewozniak:
> I can make this into a submodule of leverage, so the code stays fairly isolated, until we figure out what to do with it. It'll just
> 
> * Store prices every `epoch`, keeping the last `num_epochs`
> * Calculate and store median and SD of stored prices each epoch, per asset
> * Get stored values using a keeper method + query
> 
> That way leverage can, for every `MsgBorrow` and possibly `MsgDecollateralize` and `MsgWithdraw`, execute some custom logic.
> 
> (Only difference from issue desc. is the storing of calculated values since they should only change on epoch)

Ehh now that I think about it, it might be cleaner to do this within the oracle module or as an x/historacle module. Separation of concerns 🤷 @toteki

robert-zaremba:
related to: https://twitter.com/avi_eisen/status/1582368460645945345?s=46&t=OhTJAQXLelh-16GbUngmjw

adamewozniak:
@toteki I'm going to make a couple keeper functions for you to call on the oracle side (thinking 30 days now)

toteki:
[...]

PR #1562, opened 2022-11-09
docs: historacle spec

adamewozniak:
## Description

Initial spec for the historacle pricing oracle feature.

Please see #1521 for initial discussion 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/docs/design_docs/011-historacle-pricing.md
+++ b/docs/design_docs/011-historacle-pricing.md
@@ -1,0 +1,88 @@
+# Design Doc 011: Historacle Pricing
+
+## Status
+
+Draft
+
+## Abstract
+
+In order to support small-volume assets in a safe manner, we need to be able to calculate the weighted medians of our prices over the past X amount of time (30 days or so). This will allow the leverage module to make decisions around when to allow liquidations if an asset price is considered abnormal.
+
+## Context
+
+Right now, the volume of assets such as Juno, Umee, and Osmo are so low that it would be relatively easy to use the leverage protocol on UMEE to manipulate the prices of these assets and game the system.
+
+The attack goes:
+
+1. Account A deposits & collateralizes a large amount of USDC.
+2. Account A uses collateral to borrow a lot of the low-volume token (here, FOO).
+3. Account A sells the tokens to Account B, inflating the price of FOO.
+4. Account B collateralizes inflated FOO, and borrows USDC.
+5. Account B exits the market for a profit.
+
+In order to avoid this, and to continue with our goal of allowing users to collateralize low-volume assets, we need to have a safety net.
+
+Currently, this is defended against by disallowing the use of `Umee` as collateral, however, we would like to re-enable that and list other assets with low volume for collateral.
+
+Historacle pricing will provide an API for the leverage module to tell when prices have changed in an abnormal fashion and avoid such events.
+
+## Specification
+
+Each `Stamp Period`, the `x/oracle` module will now "stamp" the set of exchange rates in memory, and store it until a `Pruning Period` has passed (30 days). Another epoch, `Median Period`, will determine how often the Median and the `Standard Deviation around the Median` are calculated, which will also be stored in memory.
+
+These values are stored in state in order to avoid the `x/leverage` module from having to calculate them while making decisions around allowable positions for users to take.
+
+### Proposed API
+
+The leverage protocol will have access to a `Keeper` function from the `x/oracle` module, which will take a `denom`, and will return whether or not the current price of that asset is within the `Standard Deviation around the Median`.
+
+### Outcomes
+
+There are a few outcomes to consider:
+
+> This will require a chain upgrade & migration of the `x/oracle` module.
+> In order to balance performance and safety, we will have to debate the initial governance parameters.
+
+## Alternative Approaches
+
+### Feeder implementation
+
+This implementation takes the on-chain process of storing historic prices and calculating the median, and takes it off-chain. Validators would then vote on the prices of these assets in a similar fashion to the `x/oracle` voting process.
+
+There are a couple reasons for not doing this:
+
+1. Technical overhead - the x/oracle module would either have to be heavily augmented to deal with this, or we would have to build a new module to deal with the voting process, which would lead to code replication.
+2. Latency - validators take time to come to consensus in this model. Currently prices are 5 blocks behind the current APIs, this would make the median another 5 blocks behind that.
+
+## Consequences
+
+The major negative consequence of this is state bloat. Keeping the prices, medians, and standard deviations of each asset supported by the oracle module leads to a lot of bloat.
+
+The major positive is less technical overhead by utilizing an existing module, and allowing for more safety features for `x/leverage`.
+
+### Backwards Compatibility
+
+This will not introduce a new module, and it is relatively backwards compatible. We should allow for governance to disable this feature by voting to set `Stamp Period`, `Prune Period`, and `Median Period` to 0.
+
+### Positive
+
+- Efficient API for the `x/leverage` module to use for safety.
+- We can continue listing low-volume assets for collateral.
+- Use of an existing module rather than creating a new one.
+
+### Negative
+
+- State bloat.
+- Additional processing to do during each epoch.
+
+## Further Discussions
+
+1. Can we lessen state bloat without impacting the `x/leverage` module's efficiency?
+
+## Comments
+
+Originally a 24-hour median was thought to be effective enough to defend against this type of attack, but our modeling shows that we would need 30 days for adequate protection.
+
+## References
+
+- [Twitter Thread on Mango Markets Hack](https://twitter.com/joshua_j_lim/status/1579987648546246658?ref_src=twsrc%5Etfw%7Ctwcamp%5Etweetembed%7Ctwterm%5E1579987648546246658%7Ctwgr%5E4d9aa2fbcc251280df2e9f47258135bac802b986%7Ctwcon%5Es1_&ref_url=https%3A%2F%2Fdecrypt.co%2F111727%2Fsolana-defi-trading-platform-mango-markets-loses-100m-in-hack)


robert-zaremba (commented):
Thanks for the write up.

We need more details and think it through before implementing and having it in a state machine.

- Can we support more than one period length? For various use cases we most likely will need more the one window function.
- What if we need to change from 30 days to 10 days? How the parameter update will look like?
- We need to know how this will impact the performance. If we are going to track 100 of currencies from 100 of data sources. How many calculations will be performed and when?

It seams that design is broken. The spec says:
> the set of exchange rates in memory, and store it until a `Pruning Period` has passed (30 days).

* If a node is restarted, then it it will lose a data, and can not participate in consensus
* if you will like to add to the spec that we will backup it, then we need to measure the performance first. loading lot of data on every block may impact the node.

docs/design_docs/011-historacle-pricing.md
```diff
@@ -0,0 +1,88 @@
...
+
+Right now, the volume of assets such as Juno, Umee, and Osmo are so low that it would be relatively easy to use the leverage protocol on UMEE to manipulate the prices of these assets and game the system.
+
+The attack goes:
+
+1. Account A deposits & collateralizes a large amount of USDC.
+2. Account A uses collateral to borrow a lot of the low-volume token (here, FOO).
+3. Account A sells the tokens to Account B, inflating the price of FOO.
+4. Account B collateralizes inflated FOO, and borrows USDC.
+5. Account B exits the market for a profit.
```
toteki, line 21: We might use the classic scenario
```suggestion
1. Attacker deposits & collateralizes a large amount FOO, a token with low volume on exchanges.
2. Attacker spikes the price of FOO on the exchanges by buying a large amount.
3. Attacker borrows USDC using their FOO as collateral at its current (spiked) oracle price. 
4. The price of FOO returns to normal, and the value of the USDC exceeds the value of the FOO collateral.
5. Account B exits the market for a profit.
```

toteki: There is also another scenario that might be included:

1. Attacker collateralizes USDC and borrows FOO
2. Attacker sells borrowed FOO on exchanges, dumping the price
3. Attacker borrows an even larger amount of FOO using the same collateral, due to the lower price
4. The price of FOO returns to normal, and the value of the FOO exceeds the value of the USDC collateral.
5. Attacker exits the market for a profit.

docs/design_docs/011-historacle-pricing.md
```diff
@@ -0,0 +1,88 @@
...
+
+In order to avoid this, and to continue with our goal of allowing users to collateralize low-volume assets, we need to have a safety net.
```
toteki, line 23:
```suggestion
In order to avoid these attacks, and to continue with our goal of allowing users to collateralize and borrow low-volume assets, we need to have a safety net.
```
`these attacks` only if mentioning both scenarios

docs/design_docs/011-historacle-pricing.md
```diff
@@ -0,0 +1,88 @@
...
+## Specification
+
+Each `Stamp Period`, the `x/oracle` module will now "stamp" the set of exchange rates in memory, and store it until a `Pruning Period` has passed (30 days). Another epoch, `Median Period`, will determine how often the Median and the `Standard Deviation around the Median` are calculated, which will also be stored in memory.
+
+These values are stored in state in order to avoid the `x/leverage` module from having to calculate them while making decisions around allowable positions for users to take.
+
+### Proposed API
+
+The leverage protocol will have access to a `Keeper` function from the `x/oracle` module, which will take a `denom`, and will return whether or not the current price of that asset is within the `Standard Deviation around the Median`.
+
```
toteki, line 38:
```suggestion
The leverage protocol will have access to the following `keeper` functions from the `x/oracle` module:
- `historicalMedian(denom) (sdk.Dec, error)` returns the median price of an asset in the last `Pruning Period`
- `withinStandardDeviation(denom) (bool, error)` returns whether or not the current price of an asset is within the `Standard Deviation around the Median`.

```
Is this still the plan?

docs/design_docs/011-historacle-pricing.md
```diff
@@ -0,0 +1,88 @@
...
+## Consequences
+
+The major negative consequence of this is state bloat. Keeping the prices, medians, and standard deviations of each asset supported by the oracle module leads to a lot of bloat.
+
+The major positive is less technical overhead by utilizing an existing module, and allowing for more safety features for `x/leverage`.
+
+### Backwards Compatibility
+
+This will not introduce a new module, and it is relatively backwards compatible. We should allow for governance to disable this feature by voting to set `Stamp Period`, `Prune Period`, and `Median Period` to 0.
```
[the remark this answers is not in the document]
adamewozniak: `Stamp Period` is the amount of blocks we wait to record the price of assets in the state machine (ex. every 30 minutes, we save the set of prices). `Prune Period` is the amount of time after which we delete these prices (ex. every 30 days).

docs/design_docs/011-historacle-pricing.md
```diff
@@ -0,0 +1,88 @@
...
+
+In order to avoid this, and to continue with our goal of allowing users to collateralize low-volume assets, we need to have a safety net.
+
+Currently, this is defended against by disallowing the use of `Umee` as collateral, however, we would like to re-enable that and list other assets with low volume for collateral.
+
+Historacle pricing will provide an API for the leverage module to tell when prices have changed in an abnormal fashion and avoid such events.
+
+## Specification
+
+Each `Stamp Period`, the `x/oracle` module will now "stamp" the set of exchange rates in memory, and store it until a `Pruning Period` has passed (30 days). Another epoch, `Median Period`, will determine how often the Median and the `Standard Deviation around the Median` are calculated, which will also be stored in memory.
```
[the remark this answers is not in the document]
adamewozniak: Applied, we might want to also mention `Prune Period` as there are technically 3 epochs.

docs/design_docs/011-historacle-pricing.md
adamewozniak: I think it'd be good to have both. I'll add the 2nd scenario in the comments section.

docs/design_docs/011-historacle-pricing.md
adamewozniak: Do you need `historicalMedian` @toteki ? How else are you going to use it other than using `withinStandardDeviation`?

adamewozniak:
> Thanks for the write up.
> 
> We need more details and think it through before implementing and having it in a state machine.
> 
> * Can we support more than one period length? For various use cases we most likely will need more the one window function.
> * What if we need to change from 30 days to 10 days? How the parameter update will look like?
> * We need to know how this will impact the performance. If we are going to track 100 of currencies from 100 of data sources. How many calculations will be performed and when?
> 
> It seams that design is broken. The spec says:
> 
> > the set of exchange rates in memory, and store it until a `Pruning Period` has passed (30 days).
> 
> * If a node is restarted, then it it will lose a data, and can not participate in consensus
> * if you will like to add to the spec that we will backup it, then we need to measure the performance first. loading lot of data on every block may impact the node.

1. We could revise the implementation to support more than one period length, although I don't think this is necessary atm. @toteki let me know if you need this 
2. It would just be a governance update to change the `Prune Period`
3. I'll add some numbers on additional calculations
4. By "in memory", I meant "in the state machine". I'll revise.

docs/design_docs/011-historacle-pricing.md
```diff
@@ -0,0 +1,88 @@
...
+### Positive
+
+- Efficient API for the `x/leverage` module to use for safety.
+- We can continue listing low-volume assets for collateral.
+- Use of an existing module rather than creating a new one.
+
+### Negative
+
+- State bloat.
+- Additional processing to do during each epoch.
```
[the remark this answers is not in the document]
adamewozniak: wdym?

docs/design_docs/011-historacle-pricing.md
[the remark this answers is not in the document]
adamewozniak: clarified this in 75baef2c954abc7240b5962aaffe257948649865

docs/design_docs/011-historacle-pricing.md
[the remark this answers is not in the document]
adamewozniak: Clarified in 75baef2c954abc7240b5962aaffe257948649865

docs/design_docs/011-historacle-pricing.md
toteki: The current plan involves ensuring any `MsgBorrow`, `MsgWithdraw`, or `MsgDecollateralize` results in a position that is healthy (under borrow limit) using both current an historical median prices.

For that I need the specific price - the other option (if I just have a boolean `withinStandardDeviation`) would be to disable all three message types, even for tiny borrow/etc amounts, if one of a user's borrowed or collateral assets  returned `false`.

adamewozniak: Gotcha, we'll provide both APIs for ya then 🚀

adamewozniak: Please see 04c4082c48d642804c438c18de8d985b6850cfec, changed the function names a bit

docs/design_docs/011-historacle-pricing.md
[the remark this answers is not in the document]
adamewozniak: Defined the negatives better here in the computational cost section

adamewozniak:
Merging this in so that we can continue with #1548

docs/design_docs/011-historacle-pricing.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Further Discussions
+
+1. Can we lessen state bloat without impacting the `x/leverage` module's efficiency?
+
+## Comments
+
+Originally a 24-hour median was thought to be effective enough to defend against this type of attack, but our modeling shows that we would need 30 days for adequate protection.
```
robert-zaremba, line 90: where is this model? It sounds that 30 days is really a long. The price may not reflect the reality. Reality is _now_ modulo possible attack.

PR #1568, opened 2022-11-10
feat!: ibc transfer quota

gsk967:
## Description

closes: #

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


robert-zaremba (commented):
let's try to convert `x/ibctransfer` into a true sdk module. Hence I propose the following structure (which also is inline with the most recent recommendations):
```
x/ibctransfer
- "root" ->
  - generated types and codec will go here
  - ibc_module.go -> ibc module can go there, eg `IBCModule`
- /module -> cosmos-sdk module
- /quota -> keeper for ibc limits and IBC middleware implementation  (this doesn't implement anything related to gRPC service)
- /service -> sdk module keeper (it should import and encapsulate /quota) implementing gRPC service
- /client etc... - standard

- /ics20  -- move the existing ibctransfer/keeper package here
```

robert-zaremba (commented):
Let's keep generated types directly in `x/ibctransfer` and the SDK module code in `x/ibctransfer/module` as in the [authz](https://github.com/cosmos/cosmos-sdk/tree/main/x/authz) module.

proto/umee/ibctransfer/v1/query.proto
```diff
@@ -0,0 +1,56 @@
...
+
+// QueryRateLimitsOfIBCDenom defines request type
+message QueryRateLimitsOfIBCDenom {
+  string ibc_denom = 1;
+}
+
+// QueryRateLimitsOfIBCDenomResponse defines response type of QueryRateLimitsOfIBCDenom
+message QueryRateLimitsOfIBCDenomResponse {
+  RateLimit rate_limit = 1 [(gogoproto.nullable) = false];
+}
```
robert-zaremba, line 56: How about combining the two queries into one, and have `rate_limit` optional? What do you think?

robert-zaremba (commented):
Proto review only.
Let's upate the params.

proto/umee/ibctransfer/v1/tx.proto
```diff
@@ -0,0 +1,73 @@
...
+option (gogoproto.messagename_all)     = true;
+
+// Msg defines the x/ibctransfer module's Msg service.
+service Msg {
+  // UpdateIBCDenomsRateLimit adds new rate limits for ibc denoms or
+  // updates the rate limits for existed ibc denoms.
+  rpc UpdateIBCDenomsRateLimit(MsgUpdateIBCDenomsRateLimit) returns (MsgUpdateIBCDenomsRateLimitResponse);
+
+  // UpdateIBCTransferPauseStatus
+  rpc UpdateIBCTransferPauseStatus(MsgUpdateIBCTransferPauseStatus) returns (MsgUpdateIBCTransferPauseStatusResponse);
```
robert-zaremba, line 21: Let's use `Gov` prefix in bot messages above, and use Quota instead of RateLimit, and reduce the length (we are in the IBC module, so no need to repeat IBC):
```
rpc GovUpdateQuota(MsgGovUpdateQuota) returns (GovMsgUpdateQuotaResponse);
```

proto/umee/ibctransfer/v1/tx.proto
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// MsgUpdateIBCDenomsRateLimit defines the Msg/UpdateIBCDenomsRateLimit request type.
```
robert-zaremba, line 24:
```suggestion
// MsgUpdateIBCDenomsRateLimit defines the Msg/UpdateIBCDenomsRateLimit request for with x/gov proposals.
```

proto/umee/ibctransfer/v1/tx.proto
```diff
@@ -0,0 +1,73 @@
...
+  string authority   = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+  string title       = 2;
+  string description = 3;
+  // new_ibc_denoms_rate_limits defines the rate limits for ibc denoms tokens.
+  repeated MsgRateLimit new_ibc_denoms_rate_limits = 4 [(gogoproto.nullable) = false];
+  // update_ibc_denoms_rate_limits defines the new rate limits for ibc denoms tokens.
+  repeated MsgRateLimit update_ibc_denoms_rate_limits = 5 [(gogoproto.nullable) = false];
+}
+
+// MsgUpdateIBCDenomsRateLimitResponse defines response type for the Msg/UpdateIBCDenomsRateLimit
```
robert-zaremba, line 40:
```suggestion
// MsgUpdateIBCDenomsRateLimitResponse defines response type for the Msg/UpdateIBCDenomsRateLimit for with x/gov proposals.
```

proto/umee/ibctransfer/v1/tx.proto
```diff
@@ -0,0 +1,73 @@
...
+message MsgUpdateIBCDenomsRateLimitResponse {};
+
+// MsgUpdateIBCTransferPauseStatus defines request type for UpdateIBCTransferStatus
```
robert-zaremba, line 43:
```suggestion
// MsgUpdateIBCTransferPauseStatus defines request type for UpdateIBCTransferStatus for with x/gov proposals.
```

proto/umee/ibctransfer/v1/tx.proto
```diff
@@ -0,0 +1,73 @@
...
+  string title       = 2;
+  string description = 3;
+  // ibc_pause_status defines ibc transfer pause status
+  bool ibc_pause_status = 4;
+}
+
+// MsgUpdateIBCTransferPauseStatusResponse definesresponse type for Msg/MsgUpdateIBCTransferPauseStatus
+message MsgUpdateIBCTransferPauseStatusResponse {}
+
+// MsgRateLimit is format of ibc rate limit
```
robert-zaremba, line 59:
```suggestion
// MsgQuota is a structure of IBC rate limits
```

starport.ci.yml
```diff
@@ -1,34 +1,44 @@
+version: 1
 build:
```
robert-zaremba, line 2: are we still using this file?

x/ibctransfer/client/cli/query.go
```diff
@@ -0,0 +1,82 @@
...
+		},
+	}
+
+	flags.AddQueryFlagsToCmd(cmd)
+
+	return cmd
+}
+
+// GetRateLimitsForIBCDenoms
+func GetRateLimitsForIBCDenoms() *cobra.Command {
```
robert-zaremba, line 56:
```suggestion
func GetQuota() *cobra.Command {
```

proto/umee/ibctransfer/v1/query.proto
```diff
@@ -0,0 +1,43 @@
...
+
+// Query defines the gRPC querier service.
+service Query {
+  // Params queries the parameters of the x/ibc-rate-limit module.
+  rpc Params(QueryParams) returns (QueryParamsResponse) {
+    option (google.api.http).get = "/umee/ibctransfer/v1/params";
+  }
+
+  // RateLimitsOfIBCDenoms queries the rate limits of ibc denoms.
+  rpc RateLimitsOfIBCDenoms(QueryRateLimitsOfIBCDenoms) returns (QueryRateLimitsOfIBCDenomsResponse) {
```
robert-zaremba, line 20:
```suggestion
  rpc Quota(QueryQuota) returns (QueryQuotaResponse) {
```

x/ibctransfer/client/cli/tx.go
```diff
@@ -0,0 +1,24 @@
...
+	cmd := &cobra.Command{
+		Use:                        ibctransfer.ModuleName,
+		Short:                      fmt.Sprintf("Transaction commands for the %s module", ibctransfer.ModuleName),
+		DisableFlagParsing:         true,
+		SuggestionsMinimumDistance: 2,
+		RunE:                       client.ValidateCmd,
+	}
+
+	cmd.AddCommand()
+
```
robert-zaremba, line 22:
```suggestion
	cmd.AddCommand()
```

x/ibctransfer/client/cli/query.go
```diff
@@ -0,0 +1,82 @@
...
+			queryClient := ibctransfer.NewQueryClient(clientCtx)
+			if len(args) > 0 {
+				resp, err := queryClient.RateLimitsOfIBCDenoms(
+					cmd.Context(), &ibctransfer.QueryRateLimitsOfIBCDenoms{IbcDenom: args[0]},
+				)
+				return cli.PrintOrErr(resp, err, clientCtx)
+			}
+
+			resp, err := queryClient.RateLimitsOfIBCDenoms(cmd.Context(), &ibctransfer.QueryRateLimitsOfIBCDenoms{})
+			return cli.PrintOrErr(resp, err, clientCtx)
```
robert-zaremba, line 76:
```suggestion
			q := ibctransfer.QueryRateLimitsOfIBCDenoms{}
			if len(args) > 0 {
				q.IbcDenom = args[0]
			}
			resp, err := queryClient.RateLimitsOfIBCDenoms(cmd.Context(), &q)
			return cli.PrintOrErr(resp, err, clientCtx)
```

x/ibctransfer/errors.go
```diff
@@ -0,0 +1,18 @@
+package ibctransfer
+
+import (
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+var (
+	ErrIBCPauseStatus          = sdkerrors.Register(ModuleName, 1, "invalid ibc pause status")
+	ErrRateLimitExceeded       = sdkerrors.Register(ModuleName, 2, "rate limit exceeded")
```
robert-zaremba, line 9:
```suggestion
	ErrQuotaExceeded       = sdkerrors.Register(ModuleName, 2, "quota transfer exceeded")
```

x/ibctransfer/errors.go
```diff
@@ -0,0 +1,18 @@
...
+	ErrBadMessage              = sdkerrors.Register(ModuleName, 3, "bad message")
+	ErrContractError           = sdkerrors.Register(ModuleName, 4, "contract error")
+	ErrBadPacket               = sdkerrors.Register(ModuleName, 5, "bad packet")
+	ErrNoRateLimitsForIBCDenom = sdkerrors.Register(ModuleName, 6, "no rate limits for ibc denom")
```
robert-zaremba, line 13: Why do we need this error?  In the latest design we have two types of quota:
* total outflows
* per denom (same for all denoms)

proto/umee/ibctransfer/v1/tx.proto
```diff
@@ -0,0 +1,73 @@
...
+  option (cosmos.msg.v1.signer)       = "authority";
+
+  // authority is the address of the governance account.
+  string authority   = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+  string title       = 2;
+  string description = 3;
+  // new_ibc_denoms_rate_limits defines the rate limits for ibc denoms tokens.
+  repeated MsgRateLimit new_ibc_denoms_rate_limits = 4 [(gogoproto.nullable) = false];
+  // update_ibc_denoms_rate_limits defines the new rate limits for ibc denoms tokens.
+  repeated MsgRateLimit update_ibc_denoms_rate_limits = 5 [(gogoproto.nullable) = false];
```
robert-zaremba, line 37: There are only two quotas. So we should change to:
```suggestion
  // total quota for all combined outflows
  string total = 4 [(cosmos_proto.scalar)  = "cosmos.Int", (gogoproto.customtype) = "Int", (gogoproto.nullable)   = false  ];
  // per_denom quota for outflows per denom. All denoms have the same quota size.
  string per_denom = 5 [(cosmos_proto.scalar)  = "cosmos.Int", (gogoproto.customtype) = "Int", (gogoproto.nullable)   = false  ];
```

robert-zaremba: Few weeks ago we discussed to simplify the design. The idea is to have only 2 quotas: one for all combined outflows, and other one per token. But each token will have same quota.

proto/umee/ibctransfer/v1/ibc_rate_limit.proto
```diff
@@ -0,0 +1,52 @@
...
+
+import "gogoproto/gogo.proto";
+import "google/protobuf/duration.proto";
+import "google/protobuf/timestamp.proto";
+import "cosmos_proto/cosmos.proto";
+
+option go_package = "github.com/umee-network/umee/v3/x/ibctransfer";
+
+// RateLimit is format of ibc rate limit
+message RateLimit {
```
robert-zaremba, line 12:
```suggestion
message Quota {
```

robert-zaremba, line 12:
```suggestion
message Quota {
```

proto/umee/ibctransfer/v1/ibc_rate_limit.proto
```diff
@@ -0,0 +1,52 @@
...
+  // ibc_denom defines the ibc denom
+  string ibc_denom = 1;
```
robert-zaremba, line 14:
```suggestion
  // ibc_denom defines the ibc denom. When empty, it will store a total outflows.
  string ibc_denom = 1;
```

proto/umee/ibctransfer/v1/ibc_rate_limit.proto
```diff
@@ -0,0 +1,52 @@
...
+  // expired_time defines rate limit expires time based on quota_duration
+  google.protobuf.Timestamp expired_time = 2 [
+    (gogoproto.stdtime)  = true,
+    (gogoproto.nullable) = true,
+    (gogoproto.moretags) = "yaml:\"expired_time\""
+  ];
+  // outflow_sum defines the sum of outflow tokens from ibc-transfer
```
robert-zaremba, line 21:
```suggestion
  // outflow_sum defines the sum of USD value of outflow tokens through ibc-transfer
```

robert-zaremba, line 21:
```suggestion
  // outflow_sum defines the sum of USD value of outflow tokens through ibc-transfer
```

proto/umee/ibctransfer/v1/ibc_rate_limit.proto
```diff
@@ -0,0 +1,52 @@
...
+  string outflow_sum = 3 [
+    (cosmos_proto.scalar)  = "cosmos.Dec",
```
robert-zaremba, line 23: Why using Dec rather than Int?

proto/umee/ibctransfer/v1/ibc_rate_limit.proto
```diff
@@ -0,0 +1,52 @@
...
+import "cosmos_proto/cosmos.proto";
+
+option go_package = "github.com/umee-network/umee/v3/x/ibctransfer";
+
+// RateLimit is format of ibc rate limit
+message RateLimit {
+  // ibc_denom defines the ibc denom
+  string ibc_denom = 1;
+  // expired_time defines rate limit expires time based on quota_duration
+  google.protobuf.Timestamp expired_time = 2 [
```
robert-zaremba, line 16:
```suggestion
  google.protobuf.Timestamp expires = 2 [
```

proto/umee/ibctransfer/v1/ibc_rate_limit.proto
```diff
@@ -0,0 +1,52 @@
...
+package umee.ibctransfer.v1;
+
+import "gogoproto/gogo.proto";
+import "google/protobuf/duration.proto";
+import "google/protobuf/timestamp.proto";
+import "cosmos_proto/cosmos.proto";
+
+option go_package = "github.com/umee-network/umee/v3/x/ibctransfer";
+
+// RateLimit is format of ibc rate limit
```
robert-zaremba, line 11:
```suggestion
// Quota stores current sum of IBC outflow transfers.
```

app/app.go
```diff
@@ -118,14 +118,20 @@ import (
...
-	uibctransferkeeper "github.com/umee-network/umee/v3/x/ibctransfer/keeper"
+
 	"github.com/umee-network/umee/v3/x/leverage"
 	leveragekeeper "github.com/umee-network/umee/v3/x/leverage/keeper"
 	leveragetypes "github.com/umee-network/umee/v3/x/leverage/types"
 	"github.com/umee-network/umee/v3/x/oracle"
 	oraclekeeper "github.com/umee-network/umee/v3/x/oracle/keeper"
 	oracletypes "github.com/umee-network/umee/v3/x/oracle/types"
+
+	uibctransfer "github.com/umee-network/umee/v3/x/ibctransfer"
```
robert-zaremba, line 129: maybe we should call the module `uibc`? This way we would avoid renames and confusions? WDYT?
```
"github.com/umee-network/umee/v3/x/ubic"
ics20keeper "github.com/umee-network/umee/v3/x/uibc/ics20/keeper"
...
```

x/ibctransfer/events.go
```diff
@@ -0,0 +1,8 @@
+package ibctransfer
+
+const (
+	EventBadRevert          = "bad_revert"
+	AttributeKeyPacket      = "packet"
+	AttributeKeyAck         = "acknowledgement"
+	AttributeKeyFailureType = "failure_type"
```
robert-zaremba, line 7: Shall we use typed events?

x/ibctransfer/params.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// ParamKeyTable type declaration for parameters
+func ParamKeyTable() paramtypes.KeyTable {
+	return paramtypes.NewKeyTable().RegisterParamSet(&Params{})
+}
+
+// DefaultParams returns default genesis params
+func DefaultParams() *Params {
+	return &Params{
```
robert-zaremba, line 47:
```suggestion
func DefaultParams() Params {
	return Params{
```

x/ibctransfer/params.go
```diff
@@ -0,0 +1,141 @@
...
+
+	if v.IsNil() {
+		return fmt.Errorf("quota per token cannot be nil: %s", v)
+	}
+	if v.IsNegative() {
+		return fmt.Errorf("quota per token cannot be negative: %s", v)
+	}
+
+	return nil
+}
```
robert-zaremba, line 141: as discussed, let's don't use params module. Instead we should use x/gov/v1. See for example `MsgGovUpdateRegistry`.

robert-zaremba (changes requested):
Let's update the code to make it aligned with the latest design:
- [ ] we have same quota size for all denoms
- [ ] don't use params module
- [ ] simplify message names

app/app.go
gsk967: Okay , I will rename `ibctransfer` to `uibc`

x/uibc/quota/keeper/quota.go
```diff
@@ -0,0 +1,274 @@
...
+
+	// get the exchange rate of denom in USD
+	exchangeRate, err := k.oracleKeeper.GetExchangeRate(ctx, denom)
+	if err != nil {
+		return err
+	}
+
+	sendingAmount := sdk.NewDec(amount).Quo(sdk.NewDec(10).Power(uint64(tokenSettings.Exponent)))
+	amountInUSD := exchangeRate.Mul(sendingAmount)
+
```
robert-zaremba, line 163: x/oracle doesn't understand ibc/denoms. So if this is "atom", here we will see `ibc/...`, correct? We need to translate it to something like: `atom`.
- We should support `utokens` as well . cc: @toteki have function for that (not sure if `TokenValue()` handles that)

gsk967: `GetExchangeRateBase` will convert `ibc/xxx` to `symbol` and returns the exchange rate in oracle module

proto/umee/uibc/v1/quota.proto
```diff
@@ -0,0 +1,52 @@
...
+  string ibc_denom = 1;
+  // expires defines rate limit expires time based on quota_duration
+  google.protobuf.Timestamp expires = 2 [
+    (gogoproto.stdtime)  = true,
+    (gogoproto.nullable) = true,
+    (gogoproto.moretags) = "yaml:\"expires\""
+  ];
+  // outflow_sum defines the sum of USD value of outflow tokens through ibc-transfer
+  string outflow_sum = 3 [
+    (cosmos_proto.scalar)  = "cosmos.Dec",
```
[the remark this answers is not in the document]
gsk967: because `GetExchangeRateBase` returns sdk.Dec , so when we convert the dec to int we will lose the decimals of price

gsk967:
> Requests change for the functionality:
> 
>     * [x]  use `ibc_status` as an enum (see details in the comment)
> 
>     * [ ]   we should make quotas only for leverage supported tokens. If a token is not supported by leverage, and IBC quota is not disabled, then we should enable transfers. Consider this scenario: someone send to Umee HUAHUA tokens. He can't do nothing with it, because it's not supported by leverage, so user should be free to send it back, rather then lock them in umee.
> 
>     * [ ]  support uTokens
> 
>     * [ ]  verify the exponent: are we always providing based denom, or the default denom?
> 
>     * [ ]  make sure we handle `ibc/...` tokens correctly (eg, that ibc ATOM will be handled as `atom` or `uatom`?).

- we are supporting the quota for leverage supported tokens https://github.com/umee-network/umee/blob/267e9583f2f19bb085417172dd9acd81b1642cf9/x/uibc/quota/keeper/quota.go#L151
- we are getting the token exponent from leverage token settings 
- `GetExchangeRateBase` will returns the price from base denom `ibc/xxx`  https://github.com/umee-network/umee/blob/main/x/oracle/keeper/keeper.go#L91

starport.ci.yml
gsk967: Yes , we are using for ignite-cli to start the chains for testing the ibc transactions

x/uibc/quota/keeper/quota.go
```diff
@@ -0,0 +1,274 @@
...
+	quotaOfIBCDenom, err := k.getQuotaOfIBCDenom(ctx, denom)
+	if err != nil {
+		return err
+	}
+
+	if quotaOfIBCDenom == nil {
+		return nil
+	}
+
+	amount, err := strconv.ParseInt(sendAmount, 10, 64)
```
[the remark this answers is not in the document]
gsk967: FungibleTokenPacketData -> amount is string

x/uibc/quota/keeper/quota.go
toteki: If supporting uTokens (detect using `leveragetypes.HasUTokenPrefix`) you can use `leveragekeeper.ExchangeUToken` to convert to base tokens then `leveragekeeper.TokenValue` to get USD value.

The last step (`TokenValue`) might behave better for rounding than the math shown above - the reason is `sdk.Dec` having 18-point precision, so `.Quo(sdk.NewDec(10).Power(uint64(tokenSettings.Exponent)` will lose some detail with `exponent >= 18` assets like `DAI`.

robert-zaremba: @gsk967 today the uTokens are enabled over IBC, so we need to handle correct prices.
@toteki do we have an integration test for handling uToken and Token with a set of known prices? Could you point it out to @gsk967 ?

x/uibc/quota/keeper/quota.go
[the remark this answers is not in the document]
robert-zaremba: So, the idea is to to convert in different place, and update the argument of this function to have the accepted argument. Also, why we are limiting to 64bits ?

x/uibc/quota/keeper/quota.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+
+	// get the exchange rate of denom in USD
+	exchangeRate, err := k.oracleKeeper.GetExchangeRate(ctx, denom)
+	if err != nil {
+		return err
+	}
+
+	sendingAmount := sdk.NewDec(amount).Quo(sdk.NewDec(10).Power(uint64(tokenSettings.Exponent)))
+	amountInUSD := exchangeRate.Mul(sendingAmount)
```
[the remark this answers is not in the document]
gsk967: OK, Now, we are getting the exchange price from `leverageKeeper.TokenValue()` https://github.com/umee-network/umee/pull/1568/files#diff-0c1ecbc0c4238830ffea35cdd1e2f3efd4e2d0ec54d4f14e39ea6c3b4abcfac1R155

x/uibc/quota/keeper/keeper.go
```diff
@@ -0,0 +1,58 @@
...
+
+	return k.SetParams(ctx, params)
+}
+
+// UpdateTansferStatus update the ibc pause status in module params.
+func (k Keeper) UpdateTansferStatus(ctx sdk.Context, ibcStatus bool) error {
+	params := k.GetParams(ctx)
+
+	if params.IbcPause == ibcStatus {
+		return uibc.ErrIBCPauseStatus
```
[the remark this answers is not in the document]
gsk967: we are throwing error because the status is same as previous one

x/uibc/quota/ibc_module.go
```diff
@@ -0,0 +1,173 @@
+package quota
```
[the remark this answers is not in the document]
gsk967: I will raise separate pull request for documentation once everything is stable

x/uibc/events.go
```diff
@@ -0,0 +1,8 @@
+package uibc
+
+const (
+	EventBadRevert          = "bad_revert"
+	AttributeKeyPacket      = "packet"
+	AttributeKeyAck         = "acknowledgement"
+	AttributeKeyFailureType = "failure_type"
+)
```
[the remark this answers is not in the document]
gsk967: We are throwing error `BadRevert` when undo quota throws error   https://github.com/umee-network/umee/pull/1568/files#diff-7bd9ddb0e66dea157f1f1c954fbf8abe96a7724e83e704f6693df276f6033bceR101-R111

x/uibc/quota/keeper/quota.go
gsk967: If it is `uToken` , We can use above logic of toteki for checking quota of uTokens
https://github.com/umee-network/umee/pull/1568/commits/e485b2ae795163a7f71adcd1d637ddc6c4a924b0#diff-0c1ecbc0c4238830ffea35cdd1e2f3efd4e2d0ec54d4f14e39ea6c3b4abcfac1

x/uibc/events.go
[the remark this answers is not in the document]
gsk967: Ok , now added typed events

x/uibc/errors.go
```diff
@@ -0,0 +1,16 @@
+package uibc
+
+import (
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+var (
+	ErrIBCPauseStatus     = sdkerrors.Register(ModuleName, 1, "invalid ibc pause status")
+	ErrQuotaExceeded      = sdkerrors.Register(ModuleName, 2, "quota transfer exceeded")
+	ErrBadMessage         = sdkerrors.Register(ModuleName, 3, "bad message")
```
robert-zaremba, line 10: I think this is not used. Let's delete this.

x/uibc/errors.go
```diff
@@ -0,0 +1,16 @@
...
+	ErrContractError      = sdkerrors.Register(ModuleName, 4, "contract error")
```
robert-zaremba, line 11: same this. Let's review other errors as well

x/uibc/expected_keepers.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+// ICS4Wrapper defines the expected ICS4Wrapper for middleware
+type ICS4Wrapper interface {
+	WriteAcknowledgement(ctx sdk.Context, chanCap *capabilitytypes.Capability, packet ibcexported.PacketI,
+		acknowledgement ibcexported.Acknowledgement,
+	) error
+	SendPacket(ctx sdk.Context, channelCap *capabilitytypes.Capability, packet ibcexported.PacketI) error
+	GetAppVersion(ctx sdk.Context, portID, channelID string) (string, bool)
+}
```
robert-zaremba, line 26: this interface is already defined in `cosmos/ibc-go/modules/core/05-port/types/module.go`. Let's use this interface instead of copying it.

x/uibc/keys.go
```diff
@@ -0,0 +1,28 @@
...
+
+	// StoreKey defines the primary module store key
+	StoreKey = ModuleName
+
+	// RouterKey is the message route for uibc
+	RouterKey = ModuleName
+)
+
+var (
+	KeyPrefixForIBCDenom = []byte{0x01}
```
robert-zaremba, line 17:
```suggestion
	KeyPrefixDenomQuota = []byte{0x01}
```

x/uibc/keys.go
```diff
@@ -0,0 +1,28 @@
...
+	KeyTotalOutflowSum   = []byte{0x02}
```
robert-zaremba, line 18:
```suggestion
	KeyPrefixTotalOutflows   = []byte{0x02}
```

x/uibc/keys.go
```diff
@@ -0,0 +1,28 @@
...
+	// ParamsKey is the key to query all gov params
+	ParamsKey = []byte{0x03}
```
robert-zaremba, line 20:
```suggestion
	KeyPrefixParams = []byte{0x03}
```

x/uibc/keys.go
```diff
@@ -0,0 +1,28 @@
...
+	// QuotaExpiresKey is the key to store the next expire time of ibc-transfer quota
+	QuotaExpiresKey = []byte{0x04}
```
robert-zaremba, line 22:
```suggestion
	KeyPrefixQuotaExpires = []byte{0x04}
```

x/uibc/keys.go
robert-zaremba: Let's use same naming structure as in https://github.com/umee-network/umee/blob/main/x/oracle/types/keys.go

x/uibc/keys.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+func CreateKeyForQuotaOfIBCDenom(ibcDenom string) []byte {
```
robert-zaremba, line 25:
```suggestion
func KeyTotalOutflows(ibcDenom string) []byte {
```

x/uibc/module/abci.go
```diff
@@ -0,0 +1,45 @@
...
+
+		for _, quotaOfIBCDenom := range quotaOfIBCDenoms {
+			// reset the outflow sum to 0
+			quotaOfIBCDenom.OutflowSum = sdk.NewDec(0)
+			// storing the rate limits to store
+			if err := keeper.SetQuotaOfIBCDenom(ctx, quotaOfIBCDenom); err != nil {
+				panic(err)
+			}
+		}
+	}
```
robert-zaremba, line 39:
```suggestion
	params := keeper.GetParams(ctx)
	quotaExpires, err := keeper.GetExpire(ctx)
	if err != nil {
		panic(err)
	}

	// reset quotas
	if quotaExpires == nil || quotaExpires.Before(ctx.BlockTime()) {
	    newExpire := ctx.BlockTime().Add(params.QuotaDuration)
		if err := keeper.SetExpire(ctx, newExpire); err != nil {
			panic(err)
		}
		if err := keeper.SetTotalOutflowSum(ctx, sdk.NewDec(0)); err != nil {
			panic(err)
		}
        quotaOfIBCDenoms, err := keeper.GetQuotaOfIBCDenoms(ctx)
	if err != nil {
		panic(err)
	}
		for _, quotaOfIBCDenom := range quotaOfIBCDenoms {
			// reset the outflow sum to 0
			quotaOfIBCDenom.OutflowSum = sdk.NewDec(0)
			// storing the rate limits to store
			if err := keeper.SetQuotaOfIBCDenom(ctx, quotaOfIBCDenom); err != nil {
				panic(err)
			}
		}
	}
```

x/uibc/msg.go
```diff
@@ -0,0 +1,137 @@
...
+		return sdkerrors.Wrap(err, "invalid authority address")
+	}
+
+	if msg.Total.IsNil() {
+		return ErrInvalidQuota.Wrap("total quota shouldn't empty")
+	}
+
+	if !msg.Total.IsPositive() {
+		return ErrInvalidQuota.Wrap("total quota must be positive")
+	}
```
robert-zaremba, line 47:
```suggestion
	if msg.Total.IsNil() || !msg.Total.IsPositive() {
		return ErrInvalidQuota.Wrap("total quota must be positive")
	}
```

x/uibc/msg.go
```diff
@@ -0,0 +1,137 @@
...
+
+	if msg.PerDenom.IsNil() {
+		return ErrInvalidQuota.Wrap("quota per denom shouldn't empty")
+	}
+
+	if !msg.PerDenom.IsPositive() {
+		return ErrInvalidQuota.Wrap("quota per denom must be positive")
+	}
```
robert-zaremba, line 55:
```suggestion
	if msg.PerDenom.IsNil() || !msg.PerDenom.IsPositive() {
		return ErrInvalidQuota.Wrap("quota per denom must be positive")
	}
```

x/uibc/msg.go
```diff
@@ -0,0 +1,137 @@
...
+		return sdkerrors.ErrInvalidRequest.Wrap("ibc denom shouldn't be empty")
+	}
+
+	if q.OutflowSum.IsNil() {
+		return ErrInvalidQuota.Wrap("ibc denom quota expires shouldn't be empty")
+	}
+
+	if q.OutflowSum.IsNegative() {
+		return ErrInvalidQuota.Wrap("ibc denom quota expires shouldn't be empty")
+	}
```
robert-zaremba, line 134:
```suggestion
	if q.OutflowSum.IsNil() || q.OutflowSum.IsNegative()  {
		return ErrInvalidQuota.Wrap("ibc must be defined non negative number")
	}
```

robert-zaremba (commented):
* use `denom` rather than `ibc_denom` because we support our native token as well

x/uibc/errors.go
gsk967: We are using some of the errors from this file

x/uibc/quota/keeper/keeper.go
```diff
@@ -0,0 +1,54 @@
...
+	return Keeper{
+		cdc:            cdc,
+		storeKey:       key,
+		ics4Wrapper:    ics4Wrapper,
+		leverageKeeper: leverageKeeper,
+	}
+}
+
+// UpdateQuota update the quota for ibc denoms
+func (k Keeper) UpdateQuota(ctx sdk.Context, totalQuota, quotaPerDenom sdk.Dec, quotaDuration time.Duration) error {
```
robert-zaremba, line 35:
```suggestion
func (k Keeper) UpdateQuotaParams(ctx sdk.Context, totalQuota, quotaPerDenom sdk.Dec, quotaDuration time.Duration) error {
```

x/uibc/quota/keeper/keeper.go
```diff
@@ -0,0 +1,54 @@
...
+
+	params.TokenQuota = totalQuota
+	params.QuotaDuration = quotaDuration
+	params.TokenQuota = quotaPerDenom
+
+	return k.SetParams(ctx, params)
+}
+
+// UpdateTansferStatus update the ibc pause status in module params.
+func (k Keeper) UpdateTansferStatus(ctx sdk.Context, ibcStatus uibc.IBCTransferStatus) error {
```
robert-zaremba, line 46:
```suggestion
func (k Keeper) SetIBCPause(ctx sdk.Context, ibcStatus uibc.IBCTransferStatus) error {
```

proto/umee/uibc/v1/tx.proto
```diff
@@ -0,0 +1,73 @@
...
+option (gogoproto.messagename_all)     = true;
+
+// Msg defines the x/uibc module's Msg service.
+service Msg {
+  // GovUpdateQuota adds new quota for ibc denoms or
+  // updates the quota for existed ibc denoms.
+  rpc GovUpdateQuota(MsgGovUpdateQuota) returns (MsgGovUpdateQuotaResponse);
+
+  // GovUpdateTransferStatus update the status of ibc-transfer
+  rpc GovUpdateTransferStatus(MsgGovUpdateTransferStatus) returns (MsgGovUpdateTransferStatusResponse);
```
robert-zaremba, line 22:
```suggestion
  rpc GovSetIBCPause(MsgGovUpdateTransferStatus) returns (MsgGovUpdateTransferStatusResponse);
```

x/uibc/quota/keeper/quota.go
```diff
@@ -0,0 +1,266 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	transfertypes "github.com/cosmos/ibc-go/v5/modules/apps/transfer/types"
+	"github.com/cosmos/ibc-go/v5/modules/core/exported"
+
+	ltypes "github.com/umee-network/umee/v4/x/leverage/types"
+	"github.com/umee-network/umee/v4/x/uibc"
+)
+
+// GetQuotaOfIBCDenoms returns quota of all registered ibc denoms.
```
robert-zaremba, line 18:
```suggestion
// GetQuotaOfIBCDenoms returns quota of all tokens.
```

x/uibc/quota/keeper/quota.go
```diff
@@ -0,0 +1,266 @@
...
+		}
+
+		quotaOfIBCDenoms = append(quotaOfIBCDenoms, quotaOfIBCDenom)
+	}
+
+	return quotaOfIBCDenoms, nil
+}
+
+// GetQuotaOfIBCDenom retunes the rate limits of ibc denom.
+func (k Keeper) GetQuotaOfIBCDenom(ctx sdk.Context, ibcDenom string) (*uibc.Quota, error) {
```
robert-zaremba, line 41:
```suggestion
func (k Keeper) GetQuotaByDenom(ctx sdk.Context, ibcDenom string) (*uibc.Quota, error) {
```

x/uibc/quota/keeper/quota.go
```diff
@@ -0,0 +1,266 @@
...
+	rate, err := k.getQuotaOfIBCDenom(ctx, ibcDenom)
+	if err != nil {
+		return nil, err
+	}
+
+	return rate, nil
+}
+
+// getQuotaOfIBCDenom retunes the quota of ibc denom.
+func (k Keeper) getQuotaOfIBCDenom(ctx sdk.Context, ibcDenom string) (*uibc.Quota, error) {
```
robert-zaremba, line 51:
```suggestion
```

x/uibc/quota/keeper/quota.go
```diff
@@ -0,0 +1,266 @@
...
+	var quotaOfIBCDenom uibc.Quota
+	if err := k.cdc.Unmarshal(bz, &quotaOfIBCDenom); err != nil {
+		return nil, err
+	}
+
+	return &quotaOfIBCDenom, nil
+}
+
+// SetQuotaOfIBCDenoms save the rate limits of ibc denoms into store.
+func (k Keeper) SetQuotaOfIBCDenoms(ctx sdk.Context, quotaOfIBCDenoms []uibc.Quota) error {
```
robert-zaremba, line 67:
```suggestion
// SetDenomQuotas save the updated token quota.
func (k Keeper) SetDenomQuotas(ctx sdk.Context, quotas []uibc.Quota) error {
```

x/uibc/quota/keeper/quota.go
```diff
@@ -0,0 +1,266 @@
...
+	for _, quotaOfIBCDenom := range quotaOfIBCDenoms {
+		if err := k.SetQuotaOfIBCDenom(ctx, quotaOfIBCDenom); err != nil {
+			return err
+		}
```
robert-zaremba, line 71:
```suggestion
	for _, q := range quotaOfIBCDenoms {
		if err := k.SetDenomQuota(ctx, q); err != nil {
			return err
		}
```

x/uibc/quota/keeper/quota.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	return nil
+}
+
+// SetQuotaOfIBCDenom save the quota of ibc denom into store.
+func (k Keeper) SetQuotaOfIBCDenom(ctx sdk.Context, quotaOfIBCDenom uibc.Quota) error {
```
robert-zaremba, line 78:
```suggestion
// SetDenomQuota save the quota of ibc denom into store.
func (k Keeper) SetDenomQuota(ctx sdk.Context, q uibc.Quota) error {
```

x/uibc/quota/keeper/quota.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+	quotaExpires := ctx.BlockTime()
+	if err := quotaExpires.UnmarshalBinary(bz); err != nil {
+		return &quotaExpires, err
+	}
+	return &quotaExpires, nil
+}
+
+// CheckAndUpdateQuota checks the quota of ibc-transfer of denom
+func (k Keeper) CheckAndUpdateQuota(ctx sdk.Context, denom string, sendAmount string) error {
```
robert-zaremba, line 136:
```suggestion
func (k Keeper) CheckAndUpdateQuota(ctx sdk.Context, denom string, sendAmount sdk.Int) error {
```

x/uibc/quota/keeper/quota.go
```diff
@@ -0,0 +1,266 @@
...
+	if err := k.SetQuotaOfIBCDenom(ctx, *quotaOfIBCDenom); err != nil {
+		return err
+	}
+
+	// updating the total outflow sum
+	return k.SetTotalOutflowSum(ctx, totalOutflowSum)
+}
+
+// UndoUpdateQuota undo the quota of ibc denom
+func (k Keeper) UndoUpdateQuota(ctx sdk.Context, denom, sendAmount string) error {
```
robert-zaremba, line 195:
```suggestion
func (k Keeper) UndoUpdateQuota(ctx sdk.Context, denom, sendAmount sdk.Int) error {
```

x/uibc/quota/keeper/quota.go
```diff
@@ -0,0 +1,266 @@
...
+	// reset the outflow limit of token
+	quotaOfIBCDenom.OutflowSum = quotaOfIBCDenom.OutflowSum.Sub(exchangePrice)
+	if err := k.SetQuotaOfIBCDenom(ctx, *quotaOfIBCDenom); err != nil {
+		return err
+	}
+
+	// reset the total outflow sum
+	totalOutflowSum := k.GetTotalOutflowSum(ctx)
+	return k.SetTotalOutflowSum(ctx, totalOutflowSum.Sub(exchangePrice))
+}
```
robert-zaremba, line 239: there is some repetition here. Could you possible merge this functions? We could add a param (`undo boolean`) or split it into few functions.

x/uibc/quota/keeper/quota.go
```diff
@@ -0,0 +1,266 @@
...
+
+// GetLocalDenom retruns ibc denom
+// Expected denoms in the following cases:
+//
+// send non-native: transfer/channel-0/denom -> ibc/xxx
+// send native: denom -> denom
+// recv (B)non-native: denom
+// recv (B)native: transfer/channel-0/denom
+func (k Keeper) GetLocalDenom(denom string) string {
+
```
robert-zaremba, line 259:
```suggestion
```

x/uibc/quota/keeper/relay.go
```diff
@@ -0,0 +1,36 @@
...
+	ibcexported "github.com/cosmos/ibc-go/v5/modules/core/exported"
+)
+
+// SendPacket wraps IBC ChannelKeeper's SendPacket function
+func (k Keeper) SendPacket(ctx sdk.Context, chanCap *capabilitytypes.Capability, packet ibcexported.PacketI) error {
+	amount, denom, err := k.GetFundsFromPacket(packet)
+	if err != nil {
+		return sdkerrors.Wrap(err, "bad packet in rate limit's SendPacket")
+	}
+
```
robert-zaremba, line 16: we can add amount -> sdk.Int conversion here.

x/uibc/quota/ics4_wrapper.go
```diff
@@ -0,0 +1,26 @@
...
+	packet exported.PacketI,
+) error {
+	return im.keeper.SendPacket(ctx, chanCap, packet)
+}
+
+// WriteAcknowledgement implements types.Middleware
+func (im IBCMiddleware) WriteAcknowledgement(ctx sdk.Context, chanCap *capabilitytypes.Capability,
+	packet exported.PacketI, ack exported.Acknowledgement,
+) error {
+	return im.keeper.WriteAcknowledgement(ctx, chanCap, packet, ack)
```
robert-zaremba, line 25: similarly, I think this should call `app.WriteAcknowledgement`

robert-zaremba (commented):
- [ ] use `validateQuota` in ValidateBasic

and more comments

x/uibc/quota/ics4_wrapper.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	capabilitytypes "github.com/cosmos/cosmos-sdk/x/capability/types"
+	"github.com/cosmos/ibc-go/v5/modules/core/exported"
+)
+
+// GetAppVersion implements types.Middleware
+func (im IBCMiddleware) GetAppVersion(ctx sdk.Context, portID string, channelID string) (string, bool) {
+	return im.keeper.GetAppVersion(ctx, portID, channelID)
```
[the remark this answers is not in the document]
gsk967: No,
ibc middleware ics4Wrapper will call the main ibc module GetAppVersion 
> packet -> middleware -> ibc transfer module

x/uibc/quota/ics4_wrapper.go
gsk967: No

x/uibc/quota/keeper/grpc_query.go
```diff
@@ -0,0 +1,60 @@
...
+	return &uibc.QueryParamsResponse{Params: params}, nil
+}
+
+// Quota returns quota of ibc denoms.
+func (q Querier) Quota(goCtx context.Context, req *uibc.QueryQuota) (
+	*uibc.QueryQuotaResponse, error,
+) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "empty request")
+	}
```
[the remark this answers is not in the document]
gsk967: Yes, the req is pointer , so there is chance it will nil

x/uibc/errors.go
robert-zaremba: yes, but not all

x/uibc/quota/keeper/grpc_query.go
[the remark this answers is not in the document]
robert-zaremba: But grpc will never pass nil here. In Cosmos SDK sometimes we check `nil`, sometimes we don't.

x/uibc/errors.go
```diff
@@ -0,0 +1,14 @@
...
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+var (
+	ErrIBCPauseStatus     = sdkerrors.Register(ModuleName, 1, "invalid ibc pause status")
+	ErrQuotaExceeded      = sdkerrors.Register(ModuleName, 2, "quota transfer exceeded")
+	ErrNoQuotaForIBCDenom = sdkerrors.Register(ModuleName, 3, "no quota for ibc denom")
+	ErrInvalidIBCDenom    = sdkerrors.Register(ModuleName, 4, "invalid ibc denom")
+	ErrInvalidQuota       = sdkerrors.Register(ModuleName, 5, "invalid quota")
+	ErrInvalidAmount      = sdkerrors.Register(ModuleName, 6, "invalid transfer amount")
```
robert-zaremba, line 13: Use `status.Errorf(codes.InvalidArgument, "%v", ...)` or `sdk/errors.ErrInvalidRequest.Wrap(...)`

robert-zaremba: and remove the not needed errors here

x/uibc/module/abci.go
```diff
@@ -0,0 +1,46 @@
...
+
+		for _, quotaOfIBCDenom := range quotaOfIBCDenoms {
+			// reset the outflow sum to 0
+			quotaOfIBCDenom.OutflowSum = sdk.NewDec(0)
+			// storing the rate limits to store
+			if err := keeper.SetDenomQuota(ctx, quotaOfIBCDenom); err != nil {
+				panic(err)
+			}
+		}
+	}
```
robert-zaremba, line 40: move the function body to `keeper.ResetQuota(ctx)` and here just call this function.

x/uibc/msg.go
```diff
@@ -0,0 +1,125 @@
...
+
+	"github.com/umee-network/umee/v4/util/checkers"
+)
+
+var (
+	_ sdk.Msg = &MsgGovUpdateQuota{}
+	_ sdk.Msg = &MsgGovSetIBCPause{}
+)
+
+// GetTitle returns the title of the proposal.
```
robert-zaremba, line 17:
```suggestion
// GetTitle implements govv1b1.Content interface
```

x/uibc/msg.go
```diff
@@ -0,0 +1,125 @@
...
+func (msg *MsgGovUpdateQuota) GetTitle() string { return msg.Title }
+
+// GetDescription returns the description of the proposal.
```
robert-zaremba, line 20:
```suggestion
// GetDescription implements govv1b1.Content interface
```

x/uibc/msg.go
```diff
@@ -0,0 +1,125 @@
...
+) *MsgGovSetIBCPause {
+	return &MsgGovSetIBCPause{
+		Title:          title,
+		Description:    description,
+		Authority:      authority,
+		IbcPauseStatus: ibcPauseStatus,
+	}
+}
+
+// GetTitle returns the title of the proposal.
```
robert-zaremba, line 73:
```suggestion
// GetTitle implements govv1b1.Content interface
```

x/uibc/msg.go
```diff
@@ -0,0 +1,125 @@
...
+func (msg *MsgGovSetIBCPause) GetTitle() string { return msg.Title }
+
+// GetDescription returns the description of the proposal.
```
robert-zaremba, line 76:
```suggestion
// GetDescription implements govv1b1.Content interface
```

x/uibc/quota/ibc_module.go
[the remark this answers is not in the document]
robert-zaremba: OK, please add a task for that

x/uibc/quota/ibc_module.go
```diff
@@ -0,0 +1,173 @@
...
+			ctx.EventManager().EmitEvent(
+				sdk.NewEvent(
+					uibc.EventBadRevert,
+					sdk.NewAttribute(sdk.AttributeKeyModule, uibc.ModuleName),
+					sdk.NewAttribute(uibc.AttributeKeyFailureType, "timeout"),
+					sdk.NewAttribute(uibc.AttributeKeyPacket, string(packet.GetData())),
+				),
+			)
+		}
+	}
```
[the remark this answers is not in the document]
gsk967: Here we are reset the ibc-transfer quota  if we got the `acknowledgement error`

x/uibc/quota/ibc_module.go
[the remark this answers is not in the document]
gsk967: https://linear.app/umee/issue/UMEE-314/ibc-transfer-quota-docs
I will create a new issue for this in github

x/uibc/quota/ibc_module.go
```diff
@@ -0,0 +1,173 @@
...
+	channeltypes "github.com/cosmos/ibc-go/v5/modules/core/04-channel/types"
+	porttypes "github.com/cosmos/ibc-go/v5/modules/core/05-port/types"
+	"github.com/cosmos/ibc-go/v5/modules/core/exported"
+
+	"github.com/umee-network/umee/v3/x/uibc"
+	"github.com/umee-network/umee/v3/x/uibc/quota/keeper"
+)
+
+type IBCMiddleware struct {
+	app    porttypes.IBCModule
```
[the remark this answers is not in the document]
gsk967: `IBCMiddleware` struct implementing the IBCModule Interfaces, if don't use `app` then `im.OnChanOpenInit()` will call itself again again

x/uibc/errors.go
gsk967: ok

x/uibc/msg.go
```diff
@@ -0,0 +1,125 @@
...
+func (msg *MsgGovUpdateQuota) GetSignBytes() []byte {
+	return sdk.MustSortJSON(ModuleCdc.MustMarshalJSON(msg))
+}
+
+// GetSigners implements Msg
+func (msg *MsgGovUpdateQuota) GetSigners() []sdk.AccAddress {
+	return checkers.Signers(msg.Authority)
+}
+
+func NewUpdateIBCTransferPauseStatus(authority, title, description string,
```
robert-zaremba, line 62: This function is never used

x/uibc/params.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+func NewParams(ibcPause IBCTransferStatus, totalQuota, quotaPerDenom sdk.Dec, quotaDurationPerDenom int64) Params {
+	return Params{
+		IbcPause:      ibcPause,
+		TotalQuota:    totalQuota,
+		TokenQuota:    quotaPerDenom,
+		QuotaDuration: time.Second * time.Duration(quotaDurationPerDenom),
+	}
+}
```
robert-zaremba, line 31: This is used only once

x/uibc/quota/ibc_module.go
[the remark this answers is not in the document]
robert-zaremba: we can do it in a subsequent PR, I will show you what I mean

x/uibc/quota/ibc_module.go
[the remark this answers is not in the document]
robert-zaremba: yes, but if err is `nil`, why don't wa call `app.OnAcknowledgementPacket`?

x/uibc/quota/keeper/keeper.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/umee-network/umee/v4/x/uibc"
+)
+
+type Keeper struct {
+	storeKey       storetypes.StoreKey
+	cdc            codec.BinaryCodec
+	leverageKeeper uibc.LeverageKeeper
+	ics4Wrapper    porttypes.ICS4Wrapper
+}
```
robert-zaremba, line 20: would be great to use a concept of a keeper I presented with AvgKeeper. But we can do it later.

x/uibc/quota/keeper/keeper.go
```diff
@@ -0,0 +1,56 @@
...
+		ics4Wrapper:    ics4Wrapper,
+		leverageKeeper: leverageKeeper,
+	}
+}
+
+// UpdateQuotaParams update the ibc-transfer quota params for ibc denoms
+func (k Keeper) UpdateQuotaParams(ctx sdk.Context, totalQuota, quotaPerDenom sdk.Dec, quotaDuration time.Duration,
+) error {
+	params := k.GetParams(ctx)
+
```
robert-zaremba, line 38:
```suggestion
```

x/uibc/quota/keeper/keeper.go
```diff
@@ -0,0 +1,56 @@
...
+	return k.SetParams(ctx, params)
+}
+
+// SetIBCPause update the ibc pause status in module params.
+func (k Keeper) SetIBCPause(ctx sdk.Context, ibcStatus uibc.IBCTransferStatus) error {
+	params := k.GetParams(ctx)
+	if params.IbcPause == ibcStatus {
+		return uibc.ErrIBCPauseStatus.Wrapf("ibc-transfer status already have same status %s", ibcStatus.String())
+	}
```
robert-zaremba, line 51: I think this is not an error. I prefer to handle this without an error (accept that we can set the same value)

x/uibc/quota/keeper/quota.go
```diff
@@ -0,0 +1,276 @@
...
+	return quotaOfIBCDenoms, nil
+}
+
+// GetQuotaByDenom retunes the rate limits of ibc denom.
+func (k Keeper) GetQuotaByDenom(ctx sdk.Context, ibcDenom string) (*uibc.Quota, error) {
+	return k.getQuotaByDenom(ctx, ibcDenom)
+}
+
+// getQuotaByDenom retunes the quota of ibc denom.
+func (k Keeper) getQuotaByDenom(ctx sdk.Context, ibcDenom string) (*uibc.Quota, error) {
```
robert-zaremba, line 46:
```suggestion
func (k Keeper) GetQuotaByDenom(ctx sdk.Context, ibcDenom string) (*uibc.Quota, error) {
```

robert-zaremba: We can have one function, no need to have public which just calls the private and does nothing else

x/uibc/quota/keeper/quota.go
```diff
@@ -0,0 +1,276 @@
...
+// SetDenomQuota save the quota of denom into store.
+func (k Keeper) SetDenomQuota(ctx sdk.Context, quotaOfIBCDenom uibc.Quota) error {
+	store := ctx.KVStore(k.storeKey)
+	key := uibc.KeyTotalOutflows(quotaOfIBCDenom.IbcDenom)
+
+	bz, err := k.cdc.Marshal(&quotaOfIBCDenom)
+	if err != nil {
+		return err
+	}
+
```
robert-zaremba, line 81:
```suggestion
	bz, err := k.cdc.Marshal(&quotaOfIBCDenom)
	if err != nil {
		return err
	}
```

x/uibc/quota/keeper/quota.go
```diff
@@ -0,0 +1,276 @@
...
+	return sdk.MustNewDecFromStr(string(bz))
+}
+
+// SetTotalOutflowSum save the total outflow of ibc-transfer amount.
+func (k Keeper) SetTotalOutflowSum(ctx sdk.Context, amount sdk.Dec) error {
+	store := ctx.KVStore(k.storeKey)
+
+	store.Set(uibc.KeyPrefixTotalOutflows, []byte(amount.String()))
+
+	return nil
```
robert-zaremba, line 100:
```suggestion
	store := ctx.KVStore(k.storeKey)
	store.Set(uibc.KeyPrefixTotalOutflows, []byte(amount.String()))
	return nil
```

x/uibc/quota/ibc_module.go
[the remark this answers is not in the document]
gsk967: ok

x/uibc/quota/keeper/keeper.go
gsk967: ok

x/uibc/quota/keeper/keeper.go
gsk967: ok

x/uibc/quota/ibc_module.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// OnAcknowledgementPacket implements types.Middleware
+func (im IBCMiddleware) OnAcknowledgementPacket(ctx sdk.Context, packet channeltypes.Packet, acknowledgement []byte,
+	relayer sdk.AccAddress,
+) error {
+	var ack channeltypes.Acknowledgement
+	if err := json.Unmarshal(acknowledgement, &ack); err != nil {
+		return sdkerrors.ErrUnknownRequest.Wrapf("cannot unmarshal ICS-20 transfer packet acknowledgement: %v", err)
+	}
```
robert-zaremba, line 99: shouldn't we call `app.OnAcknowledgementPacket` as well? Otherwise, the app won't be called, and maybe the status will be broken. What do you think?

proto/umee/ibctransfer/v1/ibc_rate_limit.proto
gsk967: because price is in decimals

x/uibc/quota/keeper/quota.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+
+	if err := k.SetTotalOutflowSum(ctx, sdk.NewDec(0)); err != nil {
+		return err
+	}
+
+	quotaOfIBCDenoms, err := k.GetQuotaOfIBCDenoms(ctx)
+	if err != nil {
+		return err
+	}
```
robert-zaremba, line 136: no need to deserialize. Instead, let's iterate over the keys and overwrite the value directly.

x/uibc/quota/keeper/quota.go
```diff
@@ -0,0 +1,267 @@
...
+	// update the per token outflow sum
+	if err := k.SetDenomQuota(ctx, *quotaOfIBCDenom); err != nil {
+		return err
+	}
+
+	// updating the total outflow sum
+	return k.SetTotalOutflowSum(ctx, totalOutflowSum)
+}
+
+// getExchangePrice
```
robert-zaremba, line 194:
```suggestion
```

x/uibc/quota/keeper/quota.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+// GetFundsFromPacket returns transfer amount and denom
+func (k Keeper) GetFundsFromPacket(packet exported.PacketI) (string, string, error) {
+	var packetData transfertypes.FungibleTokenPacketData
+	err := json.Unmarshal(packet.GetData(), &packetData)
+	if err != nil {
+		return "", "", err
+	}
+	return packetData.Amount, k.GetLocalDenom(packetData.Denom), nil
```
robert-zaremba, line 250: Can we return `sdk.Int` here? It will help with other functions, and API of this function will be more clear (returning Int rather than string).

proto/umee/ibctransfer/v1/ibc_rate_limit.proto
robert-zaremba: I know, but maybe it's enough to trace USD as Ints (and round up)? Otherwise let's update the comment and say that this is a USD price sum.

x/uibc/quota/keeper/quota.go
gsk967: I will add that

x/uibc/quota/ibc_module.go
gsk967: We are calling `app.OnAcknowledgementPacket` end of the function , if marshal failed means some is not coming correctly from middlewares

x/uibc/quota/keeper/quota.go
gsk967: we are storing serialized quota struct into store

x/uibc/quota/ibc_module.go
gsk967: Added TODO to code to recheck the `app.OnAcknowledgementPacket` error handling

x/uibc/quota/ibc_module.go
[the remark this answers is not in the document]
robert-zaremba: @gsk967 , I resolved this here: https://github.com/umee-network/umee/pull/1776/files#diff-7bd9ddb0e66dea157f1f1c954fbf8abe96a7724e83e704f6693df276f6033bceR19

PR #1573, opened 2022-11-10
fix(oracle): historical iterator

robert-zaremba:
## Description

Historical prices stored in the x/oracle module can collide. More specifically, if we have tokens where one denom is a prefix of another token denom (eg `ETH` and `ETHW`), then an iterator for `ETH` will also return prices for `ETHW` as well and calculate the prices totally wrong. This is obviously a bug.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/x/oracle/types/keys.go
+++ b/x/oracle/types/keys.go
@@ -64,30 +64,27 @@ func GetAggregateExchangeRateVoteKey(v sdk.ValAddress) (key []byte) {
 // GetMedianKey - stored by *denom* and *block*
 func GetMedianKey(denom string, blockNum uint64) (key []byte) {
 	key = append(key, KeyPrefixMedian...)
-	key = appendDenomAndBlockToKey(key, denom, blockNum)
-	return append(key, 0) // append 0 for null-termination
+	return appendDenomAndBlock(key, denom, blockNum)
 }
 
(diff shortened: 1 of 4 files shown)


x/oracle/types/keys.go
```diff
@@ -61,33 +61,30 @@ func GetAggregateExchangeRateVoteKey(v sdk.ValAddress) (key []byte) {
 	return append(key, address.MustLengthPrefix(v)...)
 }
 
-// GetMedianKey - stored by *denom* and *block*
-func GetMedianKey(denom string, blockNum uint64) (key []byte) {
+// MedianKey - stored by *denom* and *block*
+func MedianKey(denom string, blockNum uint64) (key []byte) {
```
robert-zaremba, line 65: renamed, because we construct a value rather then extracting it.

rbajollari: Lets do this same renaming for the other oracle key constructor functions in another pr 👍

rbajollari (approved):

robert-zaremba:
> Lets do this same renaming for the other oracle key constructor functions in another pr +1

I didn't update the other functions, because that would be API breaking against what is already released. I guess nobody else is doing it so probably we are good to make that PR.

Issue #1585, opened 2022-11-14
temporarily disable historacle params


toteki:
Merging the historacle params into `proto` definitions early has caused a few problems:
- `main` branch is incompatible with `umee-1` mainnet genesis state (`panic: cannot unmarshal empty bytes` on start)
- a premature migration is being proposed for release `3.2` to address a feature (historacle) which is still being drafted

The new params should be disabled in `main` branch, especially during release to avoid unnecessary problems, and re-enabled later.

robert-zaremba:
Let's close this. It will be solved by https://github.com/umee-network/umee/issues/1668 and https://github.com/umee-network/umee/pull/1656

PR #1577, opened 2022-11-11
feat: liquidator query enabled by runtime flag

toteki:
## Description

The query `liquidation-targets` is enabled at runtime the node is started with a flag:

```bash
# Enabled
umeed start --enable-liquidator-query

# Enabled
umeed start -l

# Disabled
umeed start
```

closes: #1400

---

### Author Checklist

_All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues._

I have...

- [x] included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [x] added `!` to the type prefix if API or client breaking change
- [x] added appropriate labels to the PR
- [x] targeted the correct branch (see [PR Targeting](https://github.com/umee-network/umee/blob/main/CONTRIBUTING.md#pr-targeting))
- [ ] provided a link to the relevant issue or specification
- [x] added a changelog entry to `CHANGELOG.md`
- [x] included comments for [documenting Go code](https://blog.golang.org/godoc)
- [x] updated the relevant documentation or specification
- [x] reviewed "Files changed" and left comments if necessary
- [ ] confirmed all CI checks have passed

### Reviewers Checklist

_All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items._

I have...

- [ ] confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] confirmed all author checklist items have been addressed
- [ ] reviewed state machine logic
- [ ] reviewed API design and naming
- [ ] reviewed documentation is accurate
- [ ] reviewed tests and test coverage
- [ ] manually tested (if applicable)

PR #1586, opened 2022-11-14
fix: temporarily disable new historacle proto

toteki:
## Description

First step of #1585 - second would be re-enabling.

I think this is a cleaner path to `v3.2.0` release than leaving `main` branch incompatible or building a migration to put the params in early.


adamewozniak:
imo this is the cleaner way to go.
cc @zarazan @rbajollari

rbajollari (approved):

adamewozniak:
@toteki after this PR is merged will we be g2g on the gravity bridge upgrade?

toteki:
This and #1577 yeah

robert-zaremba:
I'm going sleep now. Personally i think it's better to add the fix rather then doing 3 similar commits (old push, this revert, and the new re-push with fix). Of course this only make sense if we are not under big pressure.l, and i already have most of the gov message part done. Need to update CLI .

robert-zaremba:
That being said I'm not against revert and redo if others support it.

robert-zaremba (changes requested):
As noted in other comment, we don't need to go back and forth with these proto changes. So blocking it because these are unnecessary back-and-forth changes. 

I've prepared a migration: https://github.com/umee-network/umee/pull/1593
still need to test it.

The full migration to go out of the param space is a bit more complex, so will be better to do it in v3.3

Issue #1596, opened 2022-11-16
Make sure we can build app without CGO


robert-zaremba:
## Summary

With `cosmwasm` enabled in an experimental build, we still need to compile with CGO enabled. This, unfortunately, doesn't provide a way how we can make a cross platform builds.

Related to: https://github.com/umee-network/umee/issues/1596

## Problem Definition

Run `CGO_ENABLED=0 make build`. You will see the following errros:

```
../../../../go/pkg/mod/github.com/!cosm!wasm/wasmvm@v1.1.1/internal/api/mocks.go:195:2: undefined: GasMeter
../../../../go/pkg/mod/github.com/!cosm!wasm/wasmvm@v1.1.1/internal/api/mocks.go:196:20: undefined: Gas
../../../../go/pkg/mod/github.com/!cosm!wasm/wasmvm@v1.1.1/internal/api/mocks.go:200:11: undefined: Gas
../../../../go/pkg/mod/github.com/!cosm!wasm/wasmvm@v1.1.1/internal/api/mocks.go:201:11: undefined: Gas
../../../../go/pkg/mod/github.com/!cosm!wasm/wasmvm@v1.1.1/internal/api/mocks.go:205:28: undefined: Gas
../../../../go/pkg/mod/github.com/!cosm!wasm/wasmvm@v1.1.1/internal/api/mocks.go:212:38: undefined: Gas
../../../../go/pkg/mod/github.com/!cosm!wasm/wasmvm@v1.1.1/internal/api/mocks.go:216:32: undefined: Gas
../../../../go/pkg/mod/github.com/!cosm!wasm/wasmvm@v1.1.1/internal/api/mocks.go:230:42: undefined: Gas
../../../../go/pkg/mod/github.com/!cosm!wasm/wasmvm@v1.1.1/internal/api/mocks.go:329:7: undefined: KVStore
../../../../go/pkg/mod/github.com/!cosm!wasm/wasmvm@v1.1.1/internal/api/mock_failure.go:15:27: undefined: GoAPI
../../../../go/pkg/mod/github.com/!cosm!wasm/wasmvm@v1.1.1/internal/api/mocks.go:329:7: too many errors
```

This also breaks goreleaser build at Github.

## Proposal

Make sure we can build a none experimental build without cgo. Probably we will need to check back the original approach with the functions, rather than `if Experimental` switch.

---

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
[...]

PR #1606, opened 2022-11-17
chore: make goreleaser cosmwasm compatible

robert-zaremba:
## Description

related to: https://github.com/umee-network/umee/issues/1596

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


robert-zaremba:
@adamewozniak I'm not closing https://github.com/umee-network/umee/issues/1596 here, because it doesn't solve the dependency issue.

.goreleaser.yml
```diff
@@ -20,19 +19,19 @@ builds:
...
-      - arm64
+#      - arm64  # github only supports linux @ amd64
 
 archives:
-  - id: bins
-    format: binary
-    name_template: "{{ .Binary }}-v{{ .Version }}-{{ .Os }}-{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}"
+  # - id: bins
+  #   format: binary
+  #   name_template: "{{ .Binary }}-v{{ .Version }}-{{ .Os }}-{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}"
```
robert-zaremba, line 30: @RafilxTenfen since we already have tar.gz archives, there is not much value in adding unpacked copies.

price-feeder/.goreleaser.yml
```diff
@@ -1,11 +1,11 @@
 project_name: price-feeder
 
 env:
-  - GO111MODULE=on
```
robert-zaremba, line 4: not needed, this is already a default.

robert-zaremba:
Adding other platforms through docker is possible, but better to do it through a different PR. Way more work and testing.

.goreleaser.yml
RafilxTenfen: IMO we should let the binaries and on the proposals we should point directly to the binaries
cc @umee-network/core-devs

robert-zaremba: why?

.goreleaser.yml
```diff
@@ -20,19 +19,19 @@ builds:
...
 
   - id: tarball
     format: tar.gz
+    wrap_in_directory: true
```
toteki, line 34: Wasn't this option related to why cosmovisor was failing on `.tar.gz`? I can't remember if it was failing on `true` or `false` though.

.goreleaser.yml
RafilxTenfen: why not? 
kkk the binaries are directly what they want, we can set the checksum, and also cosmovisor works easier with the binaries instead of the archives

With the archives need to set up one env variable more. So, what is the reason for making it more complex than it needs to be

.goreleaser.yml
RafilxTenfen: Don't remember as well, if we go with the route of archives releases, I recommend testing that with cosmovisor and updating the readme indication to add the env variable 
@robert-zaremba

.goreleaser.yml
robert-zaremba: Because it's a copy.

However, if validators will find it more useful then probably we should keep only binary. I will ask them for opinion.

.goreleaser.yml
robert-zaremba: OK

.goreleaser.yml
RafilxTenfen: the archives are a copy wrapped in an tarball

robert-zaremba: I know ;)

robert-zaremba: Majority (40:10) want's binary, so will update to provide binary.
https://discord.com/channels/815699313581490196/940283450089435216/1044412287676592178

.goreleaser.yml
robert-zaremba: confirmed, should be false.

.goreleaser.yml
```diff
@@ -20,36 +19,35 @@ builds:
...
+  #   name_template: "{{ .Binary }}-v{{ .Version }}-{{ .Os }}-{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}"
+  #   files:
+  #     - LICENSE
+  #     - README.md
 
 release:
   github:
     owner: umee-network
     name: umee
-  name_template: "v{{.Version}}"
```
toteki, line 45: ditto - can't tell if removing this will cause problems

price-feeder/.goreleaser.yml
```diff
@@ -35,17 +30,17 @@ archives:
     name_template: '{{ replace .Version "price-feeder/" "price-feeder-" }}-{{ .Os }}-{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}'
     files:
       - README.md
+      - ../LICENSE
 
 release:
   disable: false
-  name_template: "{{ .Tag }}"
```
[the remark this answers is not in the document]
robert-zaremba: yes, it's not needed

toteki (approved):
looks good - will look at the rc to be sure

jgimeno:
Sorry guys, where is the part that makes the cosmwasm compatible? I still get unlinked lib errors.

robert-zaremba:
Hi Jonathan. It's not mean to make a self contained library. It's a common issue with Cosmwasm. For this reason,.most of the chains recommend to build from source. 

If you want to use GitHub binary you need to assure to have linwasmvm in a right place in your system.

We explained and documented it in Release Notes. See also the validator announcements channel.

jgimeno:
Thanks @robert-zaremba!

PR #1624, opened 2022-11-28
docs: historic spec updates

adamewozniak:
## Description

Updates the historacle spec to allow for use by both the leverage module & ibc rate limiting

---

### Author Checklist

_All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues._

I have...

- [ ] included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] added `!` to the type prefix if API or client breaking change
- [ ] added appropriate labels to the PR
- [ ] targeted the correct branch (see [PR Targeting](https://github.com/umee-network/umee/blob/main/CONTRIBUTING.md#pr-targeting))
- [ ] provided a link to the relevant issue or specification
- [ ] added a changelog entry to `CHANGELOG.md`
- [ ] included comments for [documenting Go code](https://blog.golang.org/godoc)
- [ ] updated the relevant documentation or specification
- [ ] reviewed "Files changed" and left comments if necessary
- [ ] confirmed all CI checks have passed

### Reviewers Checklist

_All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items._

I have...

- [ ] confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] confirmed all author checklist items have been addressed
- [ ] reviewed state machine logic
- [ ] reviewed API design and naming
- [ ] reviewed documentation is accurate
- [ ] reviewed tests and test coverage
- [ ] manually tested (if applicable)

PR #1631, opened 2022-12-01
docs: update historacle API and computation docs

robert-zaremba:
## Description

follow up after https://github.com/umee-network/umee/pull/1624

1. We need to be able to return list of medians.
2. I tried to put more details in the computation part

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


docs/design_docs/011-historacle-pricing.md
```diff
@@ -53,18 +54,18 @@ We define two Maximum values, which correspond to the most we will store of a me
...
 ### Proposed API
 
 Modules will have access to the following `keeper` functions from the `x/oracle` module:
 
-- `HistoricMedian(denom) (sdk.Dec, error)` returns the median price of an asset in the last `Pruning Period`
-- `WithinHistoricDeviation(denom) (bool, error)` returns whether or not the current price of an asset is within the `Standard Deviation around the Median`.
-- `MedianOfMedians(denom string, blockNum int) sdk.Dec` returns the Median of the all the Medians recorded within the past `blockNum`.
+- `HistoricMedian(denom) ([]sdk.Dec, error)` returns list median prices of an asset in the last `Pruning Period`
+- `WithinHistoricDeviation(denom) (bool, error)` returns whether or not the current price of an asset is within the Standard Deviation around the Median.
+- `MedianOfMedians(denom string, blockNum int) sdk.Dec` returns the Median of the all the Medians recorded within the past `blockNum`. TODO: what does it mean "within the past blockNum"?
```
toteki, line 68: For example `MedianOfMedians(uumee,2500)` makes a computation involving all medians stamped "in the last 2500 blocks"

adamewozniak: @toteki we were actually thinking of using a range here as well, something like
`MedianOfMedians(denom string, fromBlock, toBlock int)`
cc: @rbajollari

adamewozniak: Please see: https://github.com/umee-network/umee/blob/a90346e3b8b721e8a4dc78707d5b3cd2db91d20b/x/oracle/keeper/historic_price.go#L192 

In https://github.com/umee-network/umee/blob/a90346e3b8b721e8a4dc78707d5b3cd2db91d20b/x/oracle/keeper/historic_price.go#L192

Might want to update that here

docs/design_docs/011-historacle-pricing.md
```diff
@@ -53,18 +54,18 @@ We define two Maximum values, which correspond to the most we will store of a me
...
+| `Median Stamp` | Of a given asset & block, the median of the stored `Price Stamps` | Every `Median Stamp Period`   | After `Maximum Median Stamps` is met. |
 
 ### Proposed API
 
 Modules will have access to the following `keeper` functions from the `x/oracle` module:
 
-- `HistoricMedian(denom) (sdk.Dec, error)` returns the median price of an asset in the last `Pruning Period`
-- `WithinHistoricDeviation(denom) (bool, error)` returns whether or not the current price of an asset is within the `Standard Deviation around the Median`.
-- `MedianOfMedians(denom string, blockNum int) sdk.Dec` returns the Median of the all the Medians recorded within the past `blockNum`.
+- `HistoricMedian(denom) ([]sdk.Dec, error)` returns list median prices of an asset in the last `Pruning Period`
```
adamewozniak, line 66: do we need this to return `[]sdk.Dec`?

zarazan: Should it be `HistoricMedians` then instead of `HistoricMedian`?

zarazan: If the response includes multiple medians should it also include a block number?

adamewozniak: @toteki need your input here

docs/design_docs/011-historacle-pricing.md
robert-zaremba: Shouldn't we use `MedianOfMedians(denom, number_of_stamps)` instead?

docs/design_docs/011-historacle-pricing.md
robert-zaremba: We need to return something. I think we should return medians, and let the client decide.

toteki: from leverage side, we'll probably be using the advanced functions (`Min/Max/Median` of medians) rather than the raw slice (or single most recent) returns. With that in mind, this one can do either and I'm fine.

docs/design_docs/011-historacle-pricing.md
toteki: So I don't think leverage will be using min/max blocks. Might be useful on ojo or similar though

docs/design_docs/011-historacle-pricing.md
zarazan: > We need to return something. I think we should return medians, and let the client decide.

I agree with this comment and we think we should re-visit this topic when moving the historacle feature over to Ojo. I don't see much difference in computing things like min/max/median_of_medians on the oracle side vs on the leverage side except we have to update proto files when the calculation changes on the oracle side. Having a single endpoint that returns X number of medians in the past could be enough.

That said, let's leave it like this for now in favor of getting v1 of this feature out. `this one can do either and I'm fine`

zarazan (approved):
I think we have agreement here.

docs/design_docs/011-historacle-pricing.md
robert-zaremba: good, so leaving as proposed here: `MedianOfMedians(denom, number_of_stamps)`

docs/design_docs/011-historacle-pricing.md
robert-zaremba: >  I don't see much difference in computing things like min/max/median_of_medians on the oracle side vs on the leverage side except we have to update proto files when the calculation changes on the oracle side. Having a single endpoint that returns X number of medians in the past could be enough.

The difference is that returning the values gives more flexibility, while complexity is the same (we don't save any CPU cycles).

Issue #1621, opened 2022-11-26
Refactor historacle pricing median tracking


rbajollari:
## Summary

1. Stamp the last x amount of medians (6 hours, etc) and provide additional APIs of
    - Median of stored Medians
    - Average of stored Medians
    - Max Median
2. Add extra param for amount of medians

3. Remove the acceptlist on the oracle side & record historacle for all assets
    - Leverage & IBC rate limiting will control their own set of assets to decide on which API to use

---

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1632, opened 2022-12-01
refactor: Update median tracking for historacle pricing

rbajollari:
## Description

closes: #1621

Things that still need to be done after this PR:
- Add new median stat keeper methods as grpc queries
- Build out oracle endblocker integration test to test historacle functionality

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/oracle/keeper/historic_price.go
```diff
@@ -47,55 +47,91 @@ func medianDeviation(median sdk.Dec, prices []sdk.Dec) sdk.Dec {
 	return medianDeviation
 }
 
-// GetMedian returns a given denom's median price in the last prune
-// period since a given block.
-func (k Keeper) GetMedian(
+// average returns the average value of a list of prices
```
zarazan, line 50: What do you think about moving all these functions that don't interact with the keeper somewhere else? They are mostly just calculations done on a set of prices, maybe something like `price_calculations.go`

x/oracle/keeper/migrations.go
```diff
@@ -18,9 +16,9 @@ func NewMigrator(keeper *Keeper) Migrator {
 
 // Migrate1to2 migrates from version 1 to 2.
 func (m Migrator) Migrate1to2(ctx sdk.Context) error {
-	m.keeper.SetStampPeriod(ctx, 1)
-	m.keeper.SetPrunePeriod(ctx, 1)
-	m.keeper.SetMedianPeriod(ctx, 1)
-	m.keeper.SetHistoricAcceptList(ctx, types.DenomList{})
+	m.keeper.SetHistoricStampPeriod(ctx, 1)
```
zarazan, line 19: Does it matter that we are not removing the old values?

x/oracle/keeper/historic_price.go
```diff
@@ -163,6 +253,25 @@ func (k Keeper) getHistoricPrices(
...
+// range.
+func (k Keeper) getMedians(
+	ctx sdk.Context,
+	denom string,
+	startBlock uint64,
+	endBlock uint64,
+) []sdk.Dec {
+	medians := []sdk.Dec{}
+	k.IterateMedians(ctx, denom, func(exchangeRate sdk.Dec, block uint64) bool {
+		if block >= startBlock && block <= endBlock {
```
robert-zaremba, line 266: `startBlock` should be included in the key prefix search, to avoid traversing and deserializing not needed records.

x/oracle/keeper/historic_price.go
```diff
@@ -187,6 +296,30 @@ func (k Keeper) IterateHistoricPrices(
...
+// Iterator stops when exhausting the source, or when the handler returns `true`.
+func (k Keeper) IterateMedians(
+	ctx sdk.Context,
+	denom string,
+	handler func(sdk.Dec, uint64) bool,
+) {
+	store := ctx.KVStore(k.storeKey)
+
+	// make sure we have one zero byte to correctly separate denoms
+	prefix := util.ConcatBytes(1, types.KeyPrefixMedian, []byte(denom))
```
robert-zaremba, line 310: we need to have separator. Otherwise the code is not valid. Consider 2 denoms: `eth`, ethpow`. When searching for `eth` one, `ethpow` will be returned as well.

robert-zaremba (changes requested):
Need to fix `IterateMedians`

x/oracle/keeper/historic_price.go
rbajollari: Added them in our utils package in pricestats.go 👍

x/oracle/keeper/historic_price.go
rbajollari: This is tricky with the KVStore iterator since the prefix is changing every iteration. I changed it to not use the iterator and accomplish the same result in `getMedians()`. Let me know what you think about that solution.

x/oracle/keeper/historic_price.go
rbajollari: I believe this already taken care of by the util methods `ConcatBytes()` and `UintWithNullPrefix()`. I updated `historic_price_test.go` to test for this and it looks like this edge case is covered for.

x/oracle/keeper/migrations.go
rbajollari: I'm not actually sure. @robert-zaremba would this be an issue?

zarazan (commented):
Great job getting this together! It needed a ton of changes since last time. I think we're almost there, a couple tiny comments from me and some consensus on the iteration method and I think its good.

x/oracle/keeper/historic_price.go
```diff
@@ -1,261 +1,302 @@
...
+        k.cdc.MustUnmarshal(iter.Value(), &decProto)
+        if handler(decProto.Dec) {
+            break
+        }
+    }
+}
+
+// getMedians returns median prices of a given denom in the store
+// starting at startblock until the latest block.
+func (k Keeper) getMedians(
```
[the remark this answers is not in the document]
zarazan: How does this resolve the issue Robert mentioned here: https://github.com/umee-network/umee/pull/1632#discussion_r1037723821 ? Is a separator not needed?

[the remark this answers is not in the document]
zarazan: Oh yeah, that will work. Does this match the key? The margin passed to `ConcatBytes` is 0. https://github.com/umee-network/umee/blob/a0630a097a82dfa3d5a4c5a1dd9f36efcca3bef5/x/oracle/types/keys.go#L65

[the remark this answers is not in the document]
rbajollari: Yeah I believe the UintWithNullPrefix util takes care of the null byte after the denom there. I also test for this in `historic_price_test.go` by adding prices and medians to a denom with a similar prefix ("UMEEtest" along with "UMEE") to check if the "UMEEtest" prices/medians being tracked mess with the calculations for "UMEE".

zarazan (approved):
I think @robert-zaremba's concerns are resolved but idk if he wants to take a look before merging.

x/oracle/keeper/historic_price.go
robert-zaremba: yes, now `ConcatBytes` solves this.

util/pricestats.go
```diff
@@ -0,0 +1,100 @@
...
+func Min(prices []sdk.Dec) (sdk.Dec, error) {
+	lenPrices := len(prices)
+	if lenPrices == 0 {
+		return sdk.ZeroDec(), ErrEmptyList
+	}
+
+	sort.Slice(prices, func(i, j int) bool {
+		return prices[i].BigInt().
+			Cmp(prices[j].BigInt()) < 0
+	})
```
robert-zaremba, line 97: don't use sort to find Min

util/pricestats.go
```diff
@@ -0,0 +1,100 @@
...
+func Max(prices []sdk.Dec) (sdk.Dec, error) {
+	lenPrices := len(prices)
+	if lenPrices == 0 {
+		return sdk.ZeroDec(), ErrEmptyList
+	}
+
+	sort.Slice(prices, func(i, j int) bool {
+		return prices[i].BigInt().
+			Cmp(prices[j].BigInt()) < 0
+	})
```
robert-zaremba, line 81: don't use sort to find `Max`

util/pricestats.go
```diff
@@ -0,0 +1,100 @@
+package util
```
robert-zaremba, line 1: let's move this file to a more specific package, eg : `util/decmath`

x/oracle/abci.go
```diff
@@ -63,14 +59,16 @@ func EndBlocker(ctx sdk.Context, k keeper.Keeper, experimental bool) error {
 			}
 
 			if experimental {
-				// Stamp rate every stamp period if asset is set to have historic stats tracked
-				if isPeriodLastBlock(ctx, params.StampPeriod) && params.HistoricAcceptList.Contains(ballotDenom.Denom) {
-					k.AddHistoricPrice(ctx, ballotDenom.Denom, exchangeRate)
+				// Stamp historic price if historic stamp period has passed
+				if isPeriodLastBlock(ctx, params.HistoricStampPeriod) {
+					k.AddHistoricPrice(ctx, strings.ToUpper(ballotDenom.Denom), exchangeRate)
```
robert-zaremba, line 64: shouldn't this be normalized when the ballot is created?

util/pricestats_test.go
```diff
@@ -0,0 +1,99 @@
...
+	require.ErrorIs(err, ErrEmptyList)
+}
+
+func TestAverage(t *testing.T) {
+	require := require.New(t)
+	prices := []sdk.Dec{
+		sdk.MustNewDecFromStr("1.1"),
+		sdk.MustNewDecFromStr("1.05"),
+		sdk.MustNewDecFromStr("1.15"),
+		sdk.MustNewDecFromStr("1.2"),
```
robert-zaremba, line 53: let's use different data, such that avg != median.

x/oracle/keeper/historic_price.go
```diff
@@ -11,91 +10,65 @@ import (
...
+func (k Keeper) GetHistoricMedians(
 	ctx sdk.Context,
 	denom string,
-) (sdk.Dec, error) {
-	store := ctx.KVStore(k.storeKey)
-	bz := store.Get(types.KeyMedian(denom))
-	if bz == nil {
-		return sdk.ZeroDec(), sdkerrors.Wrap(types.ErrNoMedian, fmt.Sprintf("denom: %s", denom))
-	}
+	numStamps uint64,
```
robert-zaremba, line 17: :+1:

x/oracle/keeper/migrations.go
robert-zaremba: In migration? But on mainnet, nor canon-2 the medians and historic prices won't be recorded. So there is nothing to remove.
The historic prices should only be recorded (now) when `experimental == true`).

robert-zaremba (commented):
thanks for fixing the denom prefix for iterations.

- [ ] Please verify if these change are in line with the updated: https://github.com/umee-network/umee/commit/3909192e39288bdeccfd6a634cfb2ccac2463c32

robert-zaremba (changes requested):
And let's remove the new files from the /util package

x/oracle/abci.go
rbajollari: If `VotePeriod` == `HistoricStampPeriod` yes, but we decided on having `HistoricStampPeriod` be a parameter to be able to test whether or not prices need to be stamped as frequent as the `VotePeriod`.

util/pricestats.go
rbajollari: Updated 👍

util/pricestats.go
rbajollari: Updated 👍

rbajollari:
@robert-zaremba @zarazan @adamewozniak Updated the keeper methods in the historacle design doc to match the actual implementation in this PR (some of the stuff agreed upon in #1631 wasn't put into the doc). Lmk if anything looks wrong in there.

docs/design_docs/011-historacle-pricing.md
```diff
@@ -63,12 +63,12 @@ We define two Maximum values, which correspond to the most we will store of a me
...
-- `MaxMedian(denom string, blockNum int) sdk.Dec` returns the Maximum of all the Medians recorded within the past `blockNum`.
-- `MinMedian(denom string, blockNum int) sdk.Dec` returns the Minimum of all the Medians recorded within the past `blockNum`.
+- `GetHistoricMedians(denom string, numStamps uint64) []sdk.Dec` returns list of last `numStamps` amount of median prices of an asset
+- `WithinHistoricDeviation(denom string) (bool, error)` returns whether or not the current price of an asset is within the Standard Deviation around the Median.
+- `GetMedianOfMedians(denom string, numStamps uint64) (sdk.Dec, error)` returns the Median of the all the Medians recorded within the past `numStamps` of medians.
+- `GetAverageOfMedians(denom string, numStamps uint64) (sdk.Dec, error)` returns the Average of all the Medians recorded within the past `numStamps` of medians.
+- `GetMaxOfMedians(denom string, numStamps uint64) (sdk.Dec, error)` returns the Maximum of all the Medians recorded within the past `numStamps` of medians.
+- `GetMinOfMedians(denom string, numStamps uint64) (sdk.Dec, error)` returns the Minimum of all the Medians recorded within the past `numStamps` of medians.
```
robert-zaremba, line 71: This is not Java. Why prefixing every function with `Get`?

rbajollari: This was to be consistent with other keeper methods e.g. `GetAggregateExchangeRatePrevote` and to avoid query functions and keeper functions sharing the same name like `MedianOfMedians`. I agree with not using `Get` though, so I've just added the `historic` key word to those method names instead of using `Get` to address the second issue.

robert-zaremba: Many of these methods do computation (eg, Avg of medians), rather than just fetching a data. 
Following a bad practice is a bad practice ;) 

Anyway, this is not a blocking comment.

x/oracle/abci.go
```diff
@@ -108,11 +106,15 @@ func EndBlocker(ctx sdk.Context, k keeper.Keeper, experimental bool) error {
...
-	// Prune historic prices every prune period
-	if isPeriodLastBlock(ctx, params.PrunePeriod) && experimental {
-		pruneBlock := uint64(ctx.BlockHeight()) - params.PrunePeriod
-		for _, v := range params.HistoricAcceptList {
-			k.DeleteHistoricPrice(ctx, v.String(), pruneBlock)
+	// Prune historic prices and medians outside pruning period determined by
+	// the stamp period multiplied by the max stamps.
+	if experimental {
+		pruneHistoricBlock := uint64(ctx.BlockHeight()) - (params.HistoricStampPeriod * params.MaximumPriceStamps)
+		pruneMedianBlock := uint64(ctx.BlockHeight()) - (params.MedianStampPeriod * params.MaximumMedianStamps)
```
robert-zaremba, line 113: how about avoiding prunning every block, and instead batch it, eg once every 100 blocks (10min).

x/oracle/keeper/genesis.go
```diff
@@ -43,8 +43,12 @@ func (k Keeper) IterateAllMedianPrices(
 	for ; iter.Valid(); iter.Next() {
 		var decProto sdk.DecProto
 		k.cdc.MustUnmarshal(iter.Value(), &decProto)
-		denom := types.ParseDenomFromMedianKey(iter.Key())
-		median := types.ExchangeRateTuple{ExchangeRate: decProto.Dec, Denom: denom}
+		denom := types.ParseDenomFromKey(iter.Key(), types.KeyPrefixMedian)
+		blockNum := types.ParseBlockFromKey(iter.Key())
```
robert-zaremba, line 47: can we have one function which will return both denom and a block number?

x/oracle/keeper/grpc_query_test.go
```diff
@@ -203,33 +203,39 @@ func (s *IntegrationTestSuite) TestQuerier_AggregateVotesAppendVotes() {
...
 
 func (s *IntegrationTestSuite) TestQuerier_Medians() {
+	app, ctx := s.app, s.ctx
+
 	atomMedian := sdk.DecCoin{Denom: "atom", Amount: sdk.MustNewDecFromStr("49.99")}
 	umeeMedian := sdk.DecCoin{Denom: "umee", Amount: sdk.MustNewDecFromStr("6541.48")}
 
-	s.app.OracleKeeper.SetMedian(s.ctx, atomMedian.Denom, atomMedian.Amount)
-	s.app.OracleKeeper.SetMedian(s.ctx, umeeMedian.Denom, umeeMedian.Amount)
+	app.OracleKeeper.SetMedianStampPeriod(ctx, 1)
```
robert-zaremba, line 211: Let's have a test with period != 1

x/oracle/keeper/historic_price.go
```diff
@@ -106,50 +80,126 @@ func (k Keeper) GetMedianDeviation(
...
+		return false, sdkerrors.Wrap(types.ErrNoMedian, fmt.Sprintf("denom: %s", denom))
+	}
+	median := medians[0]
+
+	// get latest historic price
+	prices := k.historicPrices(ctx, denom)
+	if len(prices) == 0 {
+		return false, sdkerrors.Wrap(types.ErrNoHistoricPrice, fmt.Sprintf("denom: %s", denom))
 	}
+	price := prices[0]
```
robert-zaremba, line 102: if we want only the latest price, why do we deserialize all of them? Please update, (you can use reverse iterator) to only fetch max one price. (it's possible that there will be zero)

x/oracle/keeper/historic_price.go
```diff
@@ -187,6 +237,33 @@ func (k Keeper) IterateHistoricPrices(
...
+	iter := sdk.KVStoreReversePrefixIterator(store, prefix)
+	defer iter.Close()
+
+	for ; iter.Valid(); iter.Next() {
+		decProto := sdk.DecProto{}
+		k.cdc.MustUnmarshal(iter.Value(), &decProto)
+		block := types.ParseBlockFromKey(iter.Key())
+		if handler(decProto.Dec) || block <= startBlock {
+			break
+		}
```
robert-zaremba, line 263: Let's add start block to the iterator prefix. No need to access DB records we don't need.

x/oracle/keeper/historic_price.go
rbajollari: Update to iterate until a start block like the medians iterator 👍

x/oracle/keeper/grpc_query_test.go
rbajollari: I'll do so in the next pr for updating the grpc queries

x/oracle/keeper/genesis.go
rbajollari: Done👍

x/oracle/abci.go
rbajollari: I changed it to prune every historic stamp period. Any less frequent would result in us storing more than the maximum price stamps.

util/decmath/decmath.go
```diff
@@ -0,0 +1,104 @@
...
+	for _, price := range prices {
+		sumPrices = sumPrices.Add(price)
+	}
+
+	return sumPrices.QuoInt64(int64(lenPrices)), nil
+}
+
+// Max returns the max value of a list of prices. Returns error
+// if prices is empty list.
+func Max(prices []sdk.Dec) (sdk.Dec, error) {
```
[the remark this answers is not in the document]
rbajollari: Done👍

x/oracle/abci.go
rbajollari: Initially yes but I changed it back to not convert to upper case to keep consistent with the `SetExchangeRate` method in the endblocker

x/oracle/keeper/historic_price.go
```diff
@@ -106,50 +80,126 @@ func (k Keeper) GetMedianDeviation(
...
+	denom string,
+	numStamps uint64,
+) (sdk.Dec, error) {
+	medians := k.HistoricMedians(ctx, denom, numStamps)
+	max, err := decmath.Max(medians)
+	if err != nil {
+		return sdk.ZeroDec(), sdkerrors.Wrap(err, fmt.Sprintf("denom: %s", denom))
+	}
+	return max, nil
+}
```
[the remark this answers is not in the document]
zarazan: https://github.com/umee-network/umee/pull/1631 Let's leave them in for now since that is what we already decided over here: https://github.com/umee-network/umee/pull/1631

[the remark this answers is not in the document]
robert-zaremba: It should be avoided. User should not call `MedianOfMedian` or `MaxMedian` through the oracle keeper, because it will cause to fetch all medians twice. So at minimum, if we want to keep it, we should clearly describe it in the function comment.

x/oracle/keeper/historic_price.go
robert-zaremba: No, you can iterate from a given key,

x/oracle/keeper/grpc_query_test.go
robert-zaremba: OK, then please add todo , or a task.

x/oracle/keeper/historic_price.go
robert-zaremba: I don't see the update

x/oracle/keeper/historic_price.go
rbajollari: Changed it to use KVStoreReversePrefixIteratorPaginated

x/oracle/keeper/historic_price.go
rbajollari: Im calling for only the latest historic stamp now rather than calling for all of them; `prices := k.historicPrices(ctx, denom, 1)` the "1" parameter being numstamps

x/oracle/keeper/historic_price.go
```diff
@@ -164,18 +211,44 @@ func (k Keeper) getHistoricPrices(
 }
 
 // IterateHistoricPrices iterates over historic prices of a given
-// denom in the store.
+// denom in the store in reverse.
 // Iterator stops when exhausting the source, or when the handler returns `true`.
 func (k Keeper) IterateHistoricPrices(
 	ctx sdk.Context,
 	denom string,
+	numStamps uint,
```
robert-zaremba, line 219: :+1:

robert-zaremba (approved):
utACK. Thanks for keeping improving the PR.

toteki:
Note: be sure to run `make proto-all` after these PRs.

In this case, there's a little whitespace that gets changed by `make proto-format` and some changes not reflected in `swagger/` folder.

PR #1649, opened 2022-12-07
chore: fix lint errors

toteki:
Removes or renames to `_` any unused parameters that are being caught by linters.

In general, I've tried to remove params unless there's a reason to keep `_`, for example:
- experimental flag (unused in some places which we may need later)
- required for interface (legacy querier)
- work in progress (historacle related)

Some additional lint fixes (redundant error check before return, unnecessary break, `Id` -> `ID`) also included.

Re-enables the newly cleaned linters for CI.


gsk967 (approved):
LGTM

price-feeder/oracle/provider/binance.go
```diff
@@ -239,7 +239,7 @@ func (p *BinanceProvider) getCandlePrices(key string) ([]types.CandlePrice, erro
 	return candleList, nil
 }
 
-func (p *BinanceProvider) messageReceived(messageType int, bz []byte) {
+func (p *BinanceProvider) messageReceived(_ int, bz []byte) {
```
robert-zaremba, line 242: I think it would be better to have the name, rather than removing it, like `_messageType`. It serves documentation puroposes.

toteki: In this case, hopefully going back to the `Provider` interface which it's implementing should work.

Otherwise, we could have added a `//nolint:` to some of these.

PR #1651, opened 2022-12-09
refactor: Update historic medians query

rbajollari:
---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/umee/oracle/v1/query.proto
```diff
@@ -93,6 +93,38 @@ service Query {
...
+        "/umee/historacle/v1/denoms/min_of_medians";
+  }
+
+  // MaxOfMedians returns the max of a specified amount of
+  // stamped medians
+  rpc MaxOfMedians(QueryMaxOfMedians)
+      returns (QueryMaxOfMediansResponse) {
+    option (google.api.http).get =
+        "/umee/historacle/v1/denoms/max_of_medians";
+  }
```
robert-zaremba, line 127: Why do we expose this methods? Who is going to use it? As noted in https://github.com/umee-network/umee/pull/1632#discussion_r1043840468, these methods should be avoided, unless really needed.

robert-zaremba (changes requested):
Where and why there was a decision to expose the aggregate methods outside? These can be expensive and should not be exposed by default.

proto/umee/oracle/v1/query.proto
toteki: We could keep them out of the grpc queries, but still expose on the Keeper for leverage to use

adamewozniak: Yeah they still need to be exposed for x/leverage. Iirc these queries were intended to be useful when debugging, but if they're too expensive we can just leave them out

rbajollari: I'm good with that. I'm gonna take those out and keep the update on the Medians query for this PR.

robert-zaremba: > We could keep them out of the grpc queries

I thought that's the plan. But my preference is in fact to implement it as a method over the series of prices (as is done in the `util` module), and return sequence, and let the client (leverage module) do whatever calculation it needs without risking that someone will call `Max` and `Min` within the same transaction for example. Again, see https://github.com/umee-network/umee/pull/1632#discussion_r1043840468

x/oracle/keeper/grpc_query.go
```diff
@@ -298,6 +296,7 @@ func (q querier) MedianDeviations(
 
 	if len(req.Denom) > 0 {
 		exchangeRate, err := q.HistoricMedianDeviation(ctx, req.Denom)
+
```
adamewozniak, line 299:
```suggestion
```

adamewozniak (approved):
this lgtm! we should rename the title of this PR though

proto/umee/oracle/v1/query.proto
```diff
@@ -252,6 +252,8 @@ message QueryMedians {
   option (gogoproto.goproto_getters) = false;
   // denom defines the denomination to query for.
   string denom = 1;
+  // numStamps defines the number of median stamps to query for.
+  uint64 numStamps = 2;
```
gsk967, line 256: is there any lower and upper bound for this param

gsk967 (commented):
Can you add min or max limit for `numStamps` params , and can do `make proto-all`

proto/umee/oracle/v1/query.proto
rbajollari: I added a check in the Medians query method that the numStamps parameter is not greater than the MaximumMedianStamps param so that the bound is based to what that param is set to.

gsk967 (approved):
LGTM

adamewozniak:
@robert-zaremba please re-review

proto/umee/oracle/v1/query.proto
robert-zaremba, line 256:
```suggestion
  uint32 numStamps = 2;
```

robert-zaremba: uint32 is more than enough

proto/umee/oracle/v1/query.proto
```diff
@@ -252,6 +252,8 @@ message QueryMedians {
   option (gogoproto.goproto_getters) = false;
   // denom defines the denomination to query for.
   string denom = 1;
+  // numStamps defines the number of median stamps to query for.
```
robert-zaremba, line 255: Let's define what happens if `numStamp > max amount of stamps`. I suggest that we will handle it as the max.

robert-zaremba: Also, add a comment that must be bigger than zero.

robert-zaremba (commented):
- [ ] regenerate proto based on updated types
- [ ] in ValidateBasic check that `numStamps > 0` and add a test

x/oracle/abci_test.go
```diff
@@ -42,11 +42,11 @@ func clearHistoricPrices(
...
 	denom string,
-	) {
-	stampPeriod	:= int(k.HistoricStampPeriod(ctx))
+) {
+	stampPeriod := int(k.HistoricStampPeriod(ctx))
 	numStamps := int(k.MaximumPriceStamps(ctx))
-	for i := 0; i < numStamps; i++  {
-		k.DeleteHistoricPrice(ctx, denom, uint64(ctx.BlockHeight()) - uint64(i*stampPeriod))
+	for i := 0; i < numStamps; i++ {
+		k.DeleteHistoricPrice(ctx, denom, uint64(ctx.BlockHeight())-uint64(i*stampPeriod))
```
robert-zaremba, line 49: gofmt didn't work with the editor?

x/oracle/keeper/grpc_query.go
```diff
@@ -262,15 +263,17 @@ func (q querier) Medians(
...
 	if len(req.Denom) > 0 {
-		// TODO: Add numStamps param to request to enable responding with a variable amount
-		// of last medians rather than just returning the most recent stamp
-		medianList := q.HistoricMedians(ctx, req.Denom, 1)
+		medianList := q.HistoricMedians(ctx, req.Denom, req.NumStamps)
 
-		if len(medianList) != 0 {
-			medians = medians.Add(sdk.NewDecCoinFromDec(req.Denom, medianList[0]))
+		for _, median := range medianList {
+			medians = medians.Add(sdk.NewDecCoinFromDec(req.Denom, median))
```
robert-zaremba, line 276: Don't use `Add`. Create the slice with the expected size and just set the values. It's faster. We need to be fast in this query.

x/oracle/keeper/grpc_query_test.go
```diff
@@ -216,11 +218,21 @@ func (s *IntegrationTestSuite) TestQuerier_Medians() {
 	s.Require().NoError(err)
 	s.Require().Equal(res.Medians, sdk.NewDecCoins(atomMedian, umeeMedian))
 
-	res, err = s.queryClient.Medians(ctx.Context(), &types.QueryMedians{Denom: atomMedian.Denom})
+	res, err = s.queryClient.Medians(ctx.Context(), &types.QueryMedians{Denom: atomMedian.Denom, NumStamps: 1})
+
 	s.Require().NoError(err)
 	s.Require().Equal(res.Medians, sdk.NewDecCoins(atomMedian))
```
robert-zaremba, line 224: let's add a test with `2` stamps.

robert-zaremba (changes requested):
Clicked too quickly, see my previous review.

adamewozniak:
cc @rbajollari

proto/umee/oracle/v1/query.proto
rbajollari: Handle that case as max and added the comment👍

x/oracle/abci_test.go
rbajollari: That is strange 🤔

adamewozniak:
@robert-zaremba is there any reason to block this? let's get this merged in

robert-zaremba:
@adamewozniak yes, see my review. Especially the first one. Few important things were missing. going to do second review now.

robert-zaremba (approved):
utACK

PR #1653, opened 2022-12-09
feat: implement incentive msg interface

toteki:
## Description

This implements `Msg` interface for the incentive message types and also takes care of some supporting details (codec, errors, etc)

It also moves `validateProposal` used in leverage to `util/checkers`, then reuses it for incentive proposals.

Some `make proto-all` effects are also included (oracle formatting / swagger)


x/incentive/params.go
```diff
@@ -0,0 +1,128 @@
...
+	return nil
+}
+
+func validateCommunityFundAddress(i interface{}) error {
+	_, ok := i.(string)
+	if !ok {
+		return fmt.Errorf("invalid parameter type: %T", i)
+	}
+
+	// TODO: enable once defaultCommunityFundAddress is known
```
toteki, line 115: Ideally we should ensure that this address is not a module account as well.

Current default is an empty string in this PR, so validation is commented out.

toteki (commented):
Added notes

x/incentive/msgs.go
```diff
@@ -0,0 +1,214 @@
...
+
+func (msg *MsgClaim) ValidateBasic() error {
+	_, err := sdk.AccAddressFromBech32(msg.Account)
+	return err
+}
+
+func (msg *MsgClaim) GetSigners() []sdk.AccAddress {
+	return checkers.Signers(msg.Account)
+}
+
```
gsk967, line 34: Can you implement `Route` and `Type` methods for all messages , For Ledger support we need to implement this two methods
```go
// Route implements the sdk.Msg interface.
func (m MsgType) Route() string { return types.RouterKey }

// Type implements the sdk.Msg interface.
func (m MsgType) Type() string { return sdk.MsgTypeURL(&m) }
```

proto/umee/oracle/v1/genesis.proto
```diff
@@ -22,7 +22,7 @@ message GenesisState {
       [(gogoproto.nullable) = false];
   repeated AggregateExchangeRateVote aggregate_exchange_rate_votes = 6
       [(gogoproto.nullable) = false];
-  repeated Price               historic_prices  = 8 [(gogoproto.nullable) = false];
+  repeated Price historic_prices  = 8 [(gogoproto.nullable) = false];
```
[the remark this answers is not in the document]
gsk967: can you make `historic_prices = 8` to `historic_prices = 7` or move below the `medians`

gsk967 (changes requested):
Can you implement `Route` and `Type` methods for all messages , For Ledger support we need to implement this two methods
```go
// Route implements the sdk.Msg interface.
func (m MsgType) Route() string { return types.RouterKey }

// Type implements the sdk.Msg interface.
func (m MsgType) Type() string { return sdk.MsgTypeURL(&m) }
```

proto/umee/oracle/v1/genesis.proto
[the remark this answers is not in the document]
toteki: kept the current numbers, moved to correct line order

x/incentive/msgs.go
toteki: Got it

gsk967 (approved):
LGTM

PR #1654, opened 2022-12-12
feat: leverage + historacle

toteki:
## Description

Ensure that every `MsgBorrow`, `MsgDecollateralize`, and `MsgWithdraw` is valid under both current and recent historacle prices.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


adamewozniak:
@toteki would you like a review here? Is it just in draft because we haven't cut v3.3 yet?

toteki:
@adamewozniak I've got a massive diff locally on the unit tests that I need to finish first - then a few other things to clean up (modify `QueryMaxWithdraw` to do both prices, add a parameter) and I'll mark R4R.

Also some new queries (as opposed to modified above) will be a separate PR

x/leverage/keeper/oracle.go
```diff
@@ -51,7 +55,20 @@ func (k Keeper) TokenDefaultDenomPrice(ctx sdk.Context, baseDenom string) (sdk.D
...
+	var price sdk.Dec
+
+	if historic {
+		// historic price
+		var numStamps uint32
+		price, numStamps, err = k.oracleKeeper.MedianOfHistoricMedians(ctx, t.SymbolDenom, numHistoracleStamps)
+		if err == nil && numStamps == 0 {
+			// if no price medians were available, current price is used as the historic price
+			price, err = k.oracleKeeper.GetExchangeRate(ctx, t.SymbolDenom)
+		}
```
toteki, line 67: Note here:

An `err != nil` return from `MedianOfHistoricMedians` will result in an error and stop any transactions which rely on historic prices from executing.

A `numStamps == 0` return, meaning no historic prices are stored for this denom, will attempt to use current prices instead. If they're available, the function will proceed. If an error occurs instead, it will likely be the usual `unknown denom: UMEE` oracle error for missing prices.

This behavior (assuming the current price is available) is _fail available_ rather than _fail safe_, since historical prices are a guardrail that acts in addition to the main (current prices) borrow restrictions.

x/leverage/keeper/oracle.go
```diff
@@ -11,6 +11,9 @@ import (
 
 var ten = sdk.MustNewDecFromStr("10")
 
+// TODO: Parameterize and move this
+const numHistoracleStamps = uint64(10)
```
rbajollari, line 15: Rather than adding an extra parameter for this would it work to just use `MaximumMedianStamps` since that param's value should be set in a way where we're only storing the amount of median data leverage needs?

toteki: Maybe - but if we decide in the future to have ibc rate limiting, or some other feature, use a longer amount than leverage, if would be good to have this param at the ready.

robert-zaremba (commented):
I'm not sure why we move the check a level above. Does it help with testing or not?
Also,
- [x] let's ensure we covered new borrow and withdraw scenarios in test cases. Ideally, if we can have unit test possible, but it seams, in this setup, it's too complex.
- [ ] ensure we don't call historacle median / denom more than once in the same flow.

x/leverage/keeper/keeper.go
```diff
@@ -313,19 +244,12 @@ func (k Keeper) Collateralize(ctx sdk.Context, borrowerAddr sdk.AccAddress, uTok
...
-		return err
-	}
-
-	return k.checkCollateralShare(ctx, uToken.Denom)
+	return k.bankKeeper.SendCoinsFromAccountToModule(ctx, borrowerAddr, types.ModuleName, sdk.NewCoins(uToken))
 }
 
 // Decollateralize disables selected uTokens for use as collateral by a single borrower.
+// This function does NOT check that a borrower remains under their borrow limit.
+// That assertion has been moved to MsgServer.
```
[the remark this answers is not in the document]
toteki: Since we're refactoring it anyway to check for both prices, I decided to add an additional refactor which will help with flash loans in the future. (It allows us to take multiple steps - i.e. `Supply + Collateralize` or `Borrow + Liquidate` - before the health check.)

x/leverage/keeper/grpc_query_test.go
```diff
@@ -14,7 +14,7 @@ func (s *IntegrationTestSuite) TestQuerier_RegisteredTokens() {
 
 	resp, err := s.queryClient.RegisteredTokens(ctx.Context(), &types.QueryRegisteredTokens{})
 	require.NoError(err)
-	require.Len(resp.Registry, 3, "token registry length")
+	require.Len(resp.Registry, 5, "token registry length")
```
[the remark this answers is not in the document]
toteki: Added test tokens `PUMP` and `DUMP` with rising or falling historic prices.

x/leverage/keeper/msg_server_test.go
```diff
@@ -168,3 +169,1347 @@ func (s *IntegrationTestSuite) TestUpdateRegistry() {
...
+		{
+			"borrow limit",
+			borrower,
+			coin("u/"+atomDenom, 50_000000),
+			nil,
+			nil,
+			sdk.Coin{},
+			types.ErrUndercollaterized,
+		},
+	}
```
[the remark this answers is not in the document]
toteki: I can add some specifics for those

x/leverage/keeper/borrows.go
```diff
@@ -90,7 +132,7 @@ func (k Keeper) CalculateBorrowLimit(ctx sdk.Context, collateral sdk.Coins) (sdk
 		// ignore blacklisted tokens
 		if !ts.Blacklist {
 			// get USD value of base assets
-			v, err := k.TokenValue(ctx, baseAsset)
+			v, err := k.TokenValue(ctx, baseAsset, historic)
```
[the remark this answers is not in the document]
toteki: They are all good (mostly because the checks are at the end of each message) except for `MsgMaxWithdraw` which does the (query-like) `maxWithdraw` first but still checks health at the end.

I was considering ways to streamline that, but in the case of `MsgMaxWithdraw` it would require a lot of spaghetti code to save that gas.

x/leverage/keeper/msg_server_test.go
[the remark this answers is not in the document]
toteki: Added tests for both cases for each affected message type (`Borrow`, `Withdraw`, `Decollateralize`) as well as `MaxWithdraw` (which uses the most restrictive amount given from the two calculations)

x/leverage/keeper/borrows.go
```diff
@@ -7,6 +7,47 @@ import (
...
+		return err
+	}
+	historicLimit, err := k.CalculateBorrowLimit(ctx, collateral, true)
+	if err != nil {
+		return err
+	}
+	if historicValue.GT(historicLimit) {
+		return types.ErrUndercollaterized.Wrapf(
+			"borrowed: %s, limit: %s (historic prices)", historicValue, historicLimit)
+	}
```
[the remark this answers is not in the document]
toteki: done

x/leverage/keeper/oracle.go
toteki: Outdated - now fails safe

x/leverage/keeper/oracle.go
```diff
@@ -11,6 +11,11 @@ import (
 
 var ten = sdk.MustNewDecFromStr("10")
 
+// TODO: Parameterize and move this
+const (
+	minHistoracleStamps = uint64(24)
```
robert-zaremba, line 16: Why no adding this directly to `Params`? 
Could you add a task to linear for this, and include link here?

x/leverage/keeper/msg_server_test.go
```diff
@@ -172,3 +173,1545 @@ func (s *IntegrationTestSuite) TestUpdateRegistry() {
...
+			"unregistered base token",
+			supplier,
+			coin("abcd", 80_000000),
+			nil,
+			nil,
+			sdk.Coin{},
+			types.ErrNotUToken,
+		},
+		{
+			"base token",
```
robert-zaremba, line 371:
```suggestion
			"only uToken can be withdrawn",
```

x/leverage/keeper/msg_server_test.go
```diff
@@ -172,3 +173,1545 @@ func (s *IntegrationTestSuite) TestUpdateRegistry() {
...
+			"withdraw from balance",
+			supplier,
+			coin("u/"+umeeDenom, 10_000000),
+			sdk.NewCoins(coin("u/"+umeeDenom, 10_000000)),
+			nil,
+			coin(umeeDenom, 10_000000),
+			nil,
+		},
+		{
+			"some from collateral",
```
robert-zaremba, line 398: please provide more meaningful name for this test

x/leverage/keeper/msg_server_test.go
```diff
@@ -172,3 +173,1545 @@ func (s *IntegrationTestSuite) TestUpdateRegistry() {
...
+			nil,
+			nil,
+			sdk.Coin{},
+			types.ErrInsufficientBalance,
+		},
+		{
+			"withdraw from balance",
+			supplier,
+			coin("u/"+umeeDenom, 10_000000),
+			sdk.NewCoins(coin("u/"+umeeDenom, 10_000000)),
```
robert-zaremba, line 392: let's create a helper function `ucoin` which will do the `"u/"+`, and panic if denom is already a utoken,

x/leverage/keeper/msg_server_test.go
```diff
@@ -172,3 +173,1545 @@ func (s *IntegrationTestSuite) TestUpdateRegistry() {
...
+	// using DUMP (historic price 1.00, current price 0.50) collateral
+	pumpborrower := s.newAccount(coin(dumpDenom, 100_000000))
+	s.supply(pumpborrower, coin(dumpDenom, 100_000000))
+	s.collateralize(pumpborrower, coin("u/"+dumpDenom, 100_000000))
+	s.borrow(pumpborrower, coin(pumpDenom, 5_000000))
+	// collateral value is $50 (current) or $100 (historic)
+	// borrowed value is $10 (current) or $5 (historic)
+	// collateral weights are always 0.25 in testing
+
+	tcs := []testCase{
```
robert-zaremba, line 578:
```suggestion
	tcs := []struct {
		msg                  string
		addr                 sdk.AccAddress
		denom                string
		expectedWithdraw     sdk.Coin
		expectFromCollateral sdk.Coin
		expectedTokens       sdk.Coin
		err                  error
	}{
```

robert-zaremba: same for other test cases, makes it easier to read if we have type definition right next to the values.

x/leverage/keeper/msg_server_test.go
```diff
@@ -172,3 +173,1545 @@ func (s *IntegrationTestSuite) TestUpdateRegistry() {
...
+func (s *IntegrationTestSuite) TestMsgMaxWithdraw() {
+	type testCase struct {
+		msg                  string
+		addr                 sdk.AccAddress
+		denom                string
+		expectedWithdraw     sdk.Coin
+		expectFromCollateral sdk.Coin
+		expectedTokens       sdk.Coin
+		err                  error
+	}
```
robert-zaremba, line 528:
```suggestion
```

x/leverage/keeper/msg_server_test.go
```diff
@@ -172,3 +173,1545 @@ func (s *IntegrationTestSuite) TestUpdateRegistry() {
...
+			"unregistered base token",
+			supplier,
+			"abcd",
+			sdk.Coin{},
+			sdk.Coin{},
+			sdk.Coin{},
+			types.ErrMaxWithdrawZero,
+		},
+		{
+			"uToken",
```
robert-zaremba, line 589:
```suggestion
			"can't borrow uToken",
```

x/leverage/keeper/msg_server_test.go
```diff
@@ -172,3 +173,1545 @@ func (s *IntegrationTestSuite) TestUpdateRegistry() {
...
+			require.ErrorIs(err, tc.err, tc.msg)
+		} else {
+			expectFromBalance := tc.expectedWithdraw.Sub(tc.expectFromCollateral)
+
+			// initial state
+			iBalance := app.BankKeeper.GetAllBalances(ctx, tc.addr)
+			iCollateral := app.LeverageKeeper.GetBorrowerCollateral(ctx, tc.addr)
+			iUTokenSupply := app.LeverageKeeper.GetAllUTokenSupply(ctx)
+			iExchangeRate := app.LeverageKeeper.DeriveExchangeRate(ctx, tc.denom)
+			iBorrowed := app.LeverageKeeper.GetBorrowerBorrows(ctx, tc.addr)
```
robert-zaremba, line 660: maybe we can have a helper function:
```
type balances struct {
  balance, collateral, utokenSupply, exchangeRate, borrowed
}
func getBalances(ctx, addr) balances
```
And use it across test functions : 
```
before := getBalances(...)
...
after := getBalances(...)
```
Will make the code easier to read (and much shorter)

x/leverage/keeper/msg_server_test.go
```diff
@@ -172,3 +173,1545 @@ func (s *IntegrationTestSuite) TestUpdateRegistry() {
...
+			nil,
+		},
+		{
+			"insufficient collateral",
+			supplier,
+			coin("u/"+umeeDenom, 40_000000),
+			types.ErrInsufficientCollateral,
+		},
+		{
+			"borrow limit",
```
robert-zaremba, line 876:
```suggestion
			"above borrow limit",
```

x/leverage/keeper/msg_server_test.go
```diff
@@ -172,3 +173,1545 @@ func (s *IntegrationTestSuite) TestUpdateRegistry() {
...
+		},
+
+		{
+			"acceptable decollateralize (dump borrower)",
+			dumpborrower,
+			coin("u/"+pumpDenom, 20_000000),
+			nil,
+		},
+		{
+			"borrow limit (undercollateralized under historic prices but ok with current prices)",
```
robert-zaremba, line 889:
```suggestion
			"above borrow limit (undercollateralized under historic prices but ok with current prices)",
```

x/leverage/keeper/msg_server_test.go
```diff
@@ -172,3 +173,1545 @@ func (s *IntegrationTestSuite) TestUpdateRegistry() {
...
+			types.ErrUndercollaterized,
+		},
+		{
+			"acceptable decollateralize (pump borrower)",
+			pumpborrower,
+			coin("u/"+dumpDenom, 20_000000),
+			nil,
+		},
+		{
+			"borrow limit (undercollateralized under current prices but ok with historic prices)",
```
robert-zaremba, line 901:
```suggestion
			"above borrow limit (undercollateralized under current prices but ok with historic prices)",
```

x/leverage/keeper/oracle.go
```diff
@@ -51,7 +57,23 @@ func (k Keeper) TokenDefaultDenomPrice(ctx sdk.Context, baseDenom string) (sdk.D
...
+		if err == nil && numStamps < uint32(minHistoracleStamps) {
+			return sdk.ZeroDec(), t.Exponent, types.ErrNoHistoricMedians.Wrapf(
+				"requested %d, got %d",
+				minHistoracleStamps,
+				numStamps,
+			)
+		}
+	} else {
+		// current price
+		price, err = k.oracleKeeper.GetExchangeRate(ctx, t.SymbolDenom)
```
robert-zaremba, line 75: why do we return a current price when `historic=true && minHistoracleStamps = 0`?

x/leverage/keeper/oracle_test.go
```diff
@@ -74,33 +97,101 @@ func (s *IntegrationTestSuite) TestOracle_TokenBasePrice() {
...
+
+	p, e, err = app.LeverageKeeper.TokenDefaultDenomPrice(ctx, pumpDenom, false)
+	require.NoError(err)
+	require.Equal(sdk.MustNewDecFromStr("2.0"), p)
+	require.Equal(uint32(6), e)
+
+	p, e, err = app.LeverageKeeper.TokenDefaultDenomPrice(ctx, dumpDenom, false)
+	require.NoError(err)
+	require.Equal(sdk.MustNewDecFromStr("0.5"), p)
+	require.Equal(uint32(6), e)
```
robert-zaremba, line 123: let's create a helper function:
```
check := func(err, expectedDec, p) {...}
...
check(err, "0.5", p)
```

robert-zaremba (commented):
thanks for adding tests. I suggest to make them more readable by breaking down repeated code into reusable functions.

x/leverage/keeper/msg_server_test.go
```diff
@@ -172,3 +173,1545 @@ func (s *IntegrationTestSuite) TestUpdateRegistry() {
...
+	other := s.newAccount(coin(umeeDenom, 100_000000))
+	s.supply(other, coin(umeeDenom, 100_000000))
+	s.collateralize(other, coin("u/"+umeeDenom, 100_000000))
+	s.borrow(other, coin(umeeDenom, 10_000000))
+
+	// create an additional supplier (UMEE, DUMP, PUMP tokens)
+	surplus := s.newAccount(coin(umeeDenom, 100_000000), coin(dumpDenom, 100_000000), coin(pumpDenom, 100_000000))
+	s.supply(surplus, coin(umeeDenom, 100_000000))
+	s.supply(surplus, coin(pumpDenom, 100_000000))
+	s.supply(surplus, coin(dumpDenom, 100_000000))
```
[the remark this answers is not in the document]
toteki: There are differences in some of the setups - since there is bound to be repetition in unit test setup, I might not be able to reduce all of it.

I'll think about giving `s.supply` and others the ability to accept `sdk.Coin...` args to eliminate some of this.

x/leverage/keeper/oracle.go
toteki: `minHistoracleStamps` is going to be the parameter which determines how many medians the leverage module asks for - but if we're asking for `0` medians, we don't want to ask the oracle module to compute a median from a slice of zero elements.

Once I add that parameter in properly, I will use earlier if statements to stop `historic == true` calls to price functions when `minHistoracleStamps == 0`

x/leverage/keeper/oracle.go
toteki: Follow-up task added [to linear](https://linear.app/umee/issue/UMEE-253/leverage-historacle-params-and-migration) (includes migration)

toteki:
Extracted style comments to new issue
https://github.com/umee-network/umee/issues/1681
to accelerate review

robert-zaremba (approved):
utACK. The main change is here: https://github.com/umee-network/umee/pull/1654/files#diff-9257a67f3050fe241bb803e3f669921d63c9091d7fe50d3c243d5bf255d96c7aR21
Minimum tests have been added to cover the historacle case.
We discussed to add more unit tests as well as reduce code duplication in tests in a followup PR.

PR #1657, opened 2022-12-14
feat: Add AvailableMedians method

rbajollari:
## Description

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/x/oracle/keeper/historic_price.go
+++ b/x/oracle/keeper/historic_price.go
@@ -27,6 +27,15 @@ func (k Keeper) HistoricMedians(
 	return medians
 }
 
+func (k Keeper) AvailableMedians(
+	ctx sdk.Context,
+	denom string,
+) uint32 {
+	medians := k.HistoricMedians(ctx, denom, k.MaximumMedianStamps(ctx))
+
+	return uint32(len(medians))
+}
+
 // CalcAndSetHistoricMedian uses all the historic prices of a given denom to
 // calculate its median price at the current block and set it to the store.
 // It will also call setMedianDeviation with the calculated median.
--- a/x/oracle/keeper/historic_price_test.go
+++ b/x/oracle/keeper/historic_price_test.go
@@ -32,12 +32,15 @@ func (s *IntegrationTestSuite) TestSetHistoraclePricing() {
 		ctx = ctx.WithBlockHeight(ctx.BlockHeight() + 1)
 	}
 
-	// check median and median standard deviation
+	// check medians, num of available medians, and median standard deviation
 	medians := app.OracleKeeper.HistoricMedians(ctx, displayDenom, 3)
 	s.Require().Equal(len(medians), 3)
 	s.Require().Equal(medians[0], sdk.MustNewDecFromStr("1.2"))
 	s.Require().Equal(medians[1], sdk.MustNewDecFromStr("1.125"))
 	s.Require().Equal(medians[2], sdk.MustNewDecFromStr("1.1"))
+
+	availableMedians := app.OracleKeeper.AvailableMedians(ctx, displayDenom)
+	s.Require().Equal(availableMedians, uint32(3))
 
 	medianDeviation, err := app.OracleKeeper.HistoricMedianDeviation(ctx, displayDenom)
 	s.Require().NoError(err)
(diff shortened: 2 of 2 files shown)


x/oracle/keeper/historic_price.go
```diff
@@ -27,6 +27,15 @@ func (k Keeper) HistoricMedians(
 	return medians
 }
 
+func (k Keeper) AvailableMedians(
```
adamewozniak, line 30: nit: add a godoc

adamewozniak (approved):

toteki (approved):
Nice.

On performance, is it much of a problem to be calling `AvailableMedians` and then (usually) `MedianOfMedians` right after?

rbajollari:
> Nice.
> 
> On performance, is it much of a problem to be calling `AvailableMedians` and then (usually) `MedianOfMedians` right after?

They're both going to call `HistoricMedians()` which iterates through medians in the store. The other option instead of having this method is to add an extra parameter to `MedianOfHistoricMedians()` like `minMedians` which would enforce a minimum amount of medians for it to return something by checking the length of the medians returned by `HistoricMedians()`. This would be one less call to `HistoricMedians()` than first calling `AvailableMedians()` and then `MedianOfHistoricMedians()`. What do you think about this approach @toteki?

toteki:
I think that enforcement would create an error return, which would be indistinguishable from other more serious errors (invalid denom, etc) with inspecting the error.

What about keeping the `AvailableMedians` method but also adding a return to the other functions, for example `MedianOfMedians(ctx,denom,numStamps) (sdk.Dec, uint32, error)` where the `uint32` is the number of stamps actually read?

rbajollari:
> I think that enforcement would create an error return, which would be indistinguishable from other more serious errors (invalid denom, etc) with inspecting the error.
> 
> What about keeping the `AvailableMedians` method but also adding a return to the other functions, for example `MedianOfMedians(ctx,denom,numStamps) (sdk.Dec, uint32, error)` where the `uint32` is the number of stamps actually read?
> 
> Would reduce the load from 2 -> 1 `HistoricMedians` and allow leverage to determine whether a lower than expected number of stamps constitutes an error or just a case.

Made that update to `MedianOfMedians` and the other methods as well i.e. `AverageOfMedians`, `MaxOfMedians`, and `MinOfMedians`👍

x/oracle/keeper/historic_price.go
```diff
@@ -27,6 +27,17 @@ func (k Keeper) HistoricMedians(
 	return medians
 }
 
+// AvailableMedians returns the number of medians in the store for a given denom
+// when called.
+func (k Keeper) AvailableMedians(
+	ctx sdk.Context,
+	denom string,
+) uint32 {
+	medians := k.HistoricMedians(ctx, denom, k.MaximumMedianStamps(ctx))
```
robert-zaremba, line 36: avoid data deserialization if not needed. Use native store traversal (not an Iterator function defined in this module) to count elements, rather than doing unnecessary operations

robert-zaremba (changes requested):
we should strictly avoid unnecessary operations in blockchain, which will consume extra gas. Here you only need to count elements, you don't even need to deserialize them, nor load them.

x/oracle/keeper/historic_price.go
toteki: If it's an issue, actually we can just remove this function (though I originally requested it) since it's been more efficiently replaced with the extra return on the others.

As far as leverage is using this, I can now avoid calling `AvailableMedians` directly since I'm using `MedianOfHistoricMedians`. (Assuming my other comment is addressed first)

x/oracle/keeper/historic_price.go
```diff
@@ -134,63 +145,63 @@ func (k Keeper) SetHistoricMedianDeviation(
...
 // MedianOfHistoricMedians calculates and returns the median of the last stampNum
-// historic medians.
+// historic medians as well as the amount of medians used to calculate that median.
 func (k Keeper) MedianOfHistoricMedians(
 	ctx sdk.Context,
 	denom string,
 	numStamps uint64,
-) (sdk.Dec, error) {
+) (sdk.Dec, uint32, error) {
 	medians := k.HistoricMedians(ctx, denom, numStamps)
```
toteki, line 154:
```suggestion
	medians := k.HistoricMedians(ctx, denom, numStamps)
	if len(medians) == 0 {
		sdk.ZeroDec(), 0, nil
	}
```
Prevents error return "empty list passed in" when dealing with this case.

Can also do this with the other functions, and add to each of their godoc comments:
> Also returns the actual number of median stamps retrieved, which may be less than
> numStamps requested. If no medians are available, all returns are zero and error is nil.

toteki (changes requested):
Change to treatment of zero case requested

x/oracle/keeper/historic_price.go
rbajollari: Removed `AvailableMedians` method👍

x/oracle/keeper/historic_price.go
rbajollari: Done👍

toteki (approved):
Nice, everything's been addressed

robert-zaremba (approved):
lgtm

PR #1664, opened 2022-12-15
fix:  fix the price-feeder docker image release

gsk967:
## Description

Instead of using own meta tags for docker build , changing to default docker meta tags for price-feeder image 

closes: #XXXX

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/.github/workflows/price-feeder-docker.yml
+++ b/.github/workflows/price-feeder-docker.yml
@@ -22,7 +22,6 @@ jobs:
         uses: docker/metadata-action@v4
         with:
           images: umee-network/price-feeder
-          tags: type=semver,pattern=v{{version}}
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
--- a/.github/workflows/price-feeder-docker.yml
+++ b/.github/workflows/price-feeder-docker.yml
@@ -36,7 +36,7 @@ jobs:
       - name: Build and push
         uses: docker/build-push-action@v3
         with:
-          context: ./contrib/images/price-feeder.dockerfile
+          context: contrib/images/price-feeder.dockerfile
           push: true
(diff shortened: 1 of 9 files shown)


.github/workflows/price-feeder-docker.yml
```diff
@@ -22,7 +22,6 @@ jobs:
         uses: docker/metadata-action@v4
         with:
           images: umee-network/price-feeder
-          tags: type=semver,pattern=v{{version}}
```
robert-zaremba, line 25: Why do we need to remove this? I think we should change pattern to `price-feeder/v{{version}}`

robert-zaremba (commented):
I think we should keep `tags`. Let's test it with other pattern.

.github/workflows/price-feeder-docker.yml
gsk967: I think `pattern` is only used to generate tags output , it will not check the release tag 
https://github.com/docker/metadata-action#typesemver

If we add valid semver tag , then only it will generate valid tag output for docker image 

> price-feeder/v{{version}} tag is not a valid semver

If we don't add any tags here , it will take [default tags input]( https://github.com/docker/metadata-action#tags-input), so based on default tags it will generate respective output tag for docker image

robert-zaremba: Can we only keep version as a tagged docker release? I don't think keeping branch or PR makes sense.

gsk967: Default tags are generating below tags for docker image
```json
 "tags": [
      "umee-network/price-feeder:price-feeder-v2.0.2-rc3",
      "umee-network/price-feeder:latest"
    ]
```

robert-zaremba (approved):

PR #1670, opened 2022-12-20
chore: remove latest flavor for price-feeder docker release

robert-zaremba:
## Description

Update release tags for price feeder docker release

### DIFF
--- a/.github/workflows/price-feeder-release.yml
+++ b/.github/workflows/price-feeder-release.yml
@@ -19,6 +19,9 @@ jobs:
           go-version: 1.19
           cache: true
           cache-dependency-path: price-feeder/go.sum
+          flavor: |
+            latest=false
+
       # Parse 'v*.*.*' semantic version from 'price-feeder/v*.*.*' and save to
       # the $GITHUB_ENV environment variable.
       - name: Set env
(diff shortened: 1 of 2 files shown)


.github/workflows/price-feeder-release.yml
```diff
@@ -19,6 +19,9 @@ jobs:
           go-version: 1.19
           cache: true
           cache-dependency-path: price-feeder/go.sum
+          flavor: |
```
gsk967, line 22: We need to add this one inside the `price-feeder-docker.yml` file at docker metadata

gsk967 (changes requested):
Please add 
```
flavor: |
  latest=false
```
inside the `price-feeder-docker.yml` file at docker metadata
https://github.com/docker/metadata-action#flavor-input

gsk967 (approved):
LGTM

.github/workflows/price-feeder-release.yml
robert-zaremba: thanks for catching. Updated. I'm testing in a local fork and will report if everything works.

robert-zaremba:
thanks for catching the issue.

PR #1689, opened 2023-01-05
DNM: param breaking tests

toteki:
This is related to #1685 - it proves that a marshaled `OldToken` (which is `Token` from `v3.3.0`) can be unmarshaled into a`Token` with the new param.

This test _might_ be enough to prove that we don't need a migration for leverage. I have also run some manual tests locally, but I would like to check gov queries with `OldToken` as well.


x/leverage/keeper/token_test.go
```diff
@@ -27,3 +28,52 @@ func (s *IntegrationTestSuite) TestGetToken() {
...
+		MaxSupplyUtilization:   sdk.MustNewDecFromStr("0.9"),
+		MinCollateralLiquidity: sdk.MustNewDecFromStr("0"),
+		MaxSupply:              sdk.NewInt(100_000_000000),
+	}
+
+	// Store the old token in state - simulating existing state
+	require.NoError(app.LeverageKeeper.SetOldToken(ctx, oldToken))
+
+	// Try to get it from state - using new token struct
+	t, err := app.LeverageKeeper.GetTokenSettings(ctx, "uold")
```
toteki, line 61: Here's the meat of it. We can set and old token and get a new one no problem (since proto tolerates adding fields)

PR #1706, opened 2023-01-11
fix: add err check in e2e historacle tests

robert-zaremba:
## Description

Err check is missing in the e2e historacle tests


toteki (approved):

zarazan (approved):
👍🏻 This check might get pull out of the CheckMedian function in the future so we can more efficiently loop over it multiple times / run other checks that require prices to exists, etc.

tests/grpc/historical.go
```diff
@@ -30,13 +32,21 @@ func MedianCheck(val1Client *UmeeClient) error {
...
 		if err == nil && len(exchangeRates) == len(denomAcceptList) {
 			break
 		}
 		<-chainHeight.HeightChanged
 	}
+	// error if the loop above didn't succeed
+	if err != nil {
+		return err
+	}
+	if len(exchangeRates) == len(denomAcceptList) {
```
zarazan, line 47:
```suggestion
	if len(exchangeRates) != len(denomAcceptList) {
```

PR #1715, opened 2023-01-16
feat: add spot_borrow_limit to account summary and use min borrow limit

toteki:
`borrow_limit` is now the minimum of `spot_borrow_limit` and `historic_borrow_limit` in the account summary query.

Actual leverage logic (see README) is unaffected.

Also contains some output from `make proto-all` in incentive module which has been ignored.

Non-state-breaking PR.


gsk967 (approved):
LGTM

PR #1720, opened 2023-01-17
fix(oracle): Update price pruning

zarazan:
- Update pruning for historic stamps, medians, and median deviations to prune all prices older than a specific block instead of only pruning one price at a time.
- Remove `VotePeriod` from the pruning block calculations
- Sort the query output for median and median deviations by block and then denom

Fixes no medians or median deviations being pruned

### DIFF
--- a/tests/e2e/e2e_setup_test.go
+++ b/tests/e2e/e2e_setup_test.go
@@ -275,6 +275,7 @@ func (s *IntegrationTestSuite) initGenesis() {
 	oracleGenState.Params.HistoricStampPeriod = 5
 	oracleGenState.Params.MaximumPriceStamps = 4
 	oracleGenState.Params.MedianStampPeriod = 20
+	oracleGenState.Params.MaximumMedianStamps = 2
 
 	bz, err = cdc.MarshalJSON(&oracleGenState)
 	s.Require().NoError(err)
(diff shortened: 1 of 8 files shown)


tests/e2e/e2e_setup_test.go
```diff
@@ -275,6 +275,7 @@ func (s *IntegrationTestSuite) initGenesis() {
 	oracleGenState.Params.HistoricStampPeriod = 5
 	oracleGenState.Params.MaximumPriceStamps = 4
 	oracleGenState.Params.MedianStampPeriod = 20
+	oracleGenState.Params.MaximumMedianStamps = 2
```
rbajollari, line 278: I think we should confirm the test still passes with more median stamps (maybe 5) unless that requires rewriting things beyond this parameter

rbajollari (approved):
LGTM!

tests/e2e/e2e_setup_test.go
zarazan: It will pass regardless of this number since we look for the median at a specific block height. I was going to put some stuff at the bottom checking that they were pruned correctly but I'd need to add another keeper method. I'll set it to 5 just for funzies.

adamewozniak (approved):
looks good to me folks 🙏

tests/e2e/e2e_setup_test.go
zarazan: I'll make another PR for this next.

PR #1724, opened 2023-01-18
chore(tests): Add checks to oracle ABCI test

zarazan:
- Update oracle ABCI test to check for the correct amount of pruning prices
- Update oracle ABCI test to check for correct median deviations


rbajollari (approved):
Endblocker looks much cleaner now!🔥

PR #1756, opened 2023-01-28
docs: liquidation manual

toteki:
Provides a basic manual for running liquidations, and a link to it in the readme.


docs/liquidators/manual.md
```diff
@@ -0,0 +1,162 @@
...
+- Repay `0.084001 ATOM` on Umee (our borrow accrued a few minutes of interest)
+
+For reference, the repayment looks like:
+
+```sh
+% umeed tx leverage repay 84001ibc/C4CFF46FD6DE35CA4CF4CE031E643C8FDC9BA4B99AE598E9B0ED98FE3A2319F9 --from my-key --chain-id umee-1 --gas auto --gas-adjustment 3.0 --gas-prices 0.1uumee -y --broadcast-mode block
+```
+
+The liquidator account has returned to its initial state, except:
+- An extra `0.04 ATOM` in its balance on Umee (profit)
```
toteki, line 156:
```suggestion
- An extra `0.004 ATOM` in its balance on Umee (profit)
```

docs/liquidators/manual.md
```diff
@@ -0,0 +1,162 @@
...
+Now the required `ATOM` is in our account.
+
+Another option would have been to buy `ATOM` on an exchange, or have an initial balance prepared before we selected our liquidation target.
+
+### Liquidation Transaction
+
+In a liquidation transaction, you need to specify three things:
+- target address (`umee1l2jv...`)
+- repayment coin (`84000ibc/C4CFF...`, which is `0.084 ATOM`)
+- reward token (`ibc/49788...`, which is USDC)
```
toteki, line 125:
```suggestion
- reward token (`ibc/49788...`, which is `USDC`)
```

docs/liquidators/manual.md
```diff
@@ -0,0 +1,162 @@
+# Manual Liquidations
```
robert-zaremba, line 1: shall we put this in `/x/leverage/docs`?

docs/liquidators/manual.md
```diff
@@ -0,0 +1,162 @@
...
+- Preparing a liquidator account
+- Querying for eligible liquidation targets
+- Executing a liquidation transaction
+- Selling liquidation rewards for a profit
+
+This document contains notes on how to do these manually - a more sophisticated approach would be to automate one or more steps and monitor the results.
+
+## Preparing a Liquidator Account
+
+Any wallet you control can be used as a liquidator, as long as it has a pool of assets and tokens for gas on the relevant chains.
```
robert-zaremba, line 14: Why talking about other chains? I think it is not relevant. Liquidation happens on Umee. So a liquidator has to have:
Umee account with tokens required for liquidation. Eg, to liquidate `100 AXL-USDC` and `50 ATOM`, the liquidator has to have at least:
1.  `100 AXL-USDC` and `50 ATOM`
2. `1UMEE` for transaction fees (variable).

docs/liquidators/manual.md
```diff
@@ -0,0 +1,162 @@
...
+
+Here is an example setup:
+- `100 AXL-USDC` from axelar collateralized on [app.umee.cc](https://app.umee.cc/) with no active borrows
+- `1 UMEE` on Umee
+- `1 ATOM` on Cosmos Hub
+- `1 AXL` on Axelar
+- `1 OSMO` on Osmosis
+
+The main collateral will allow you to borrow any tokens required during liquidation, and the other tokens are for gas on their native chains.
```
robert-zaremba, line 23: > gas on their native chains

This is not required for liquidation. We can create a new section by explaining additional flows, a liquidator can take.

docs/liquidators/manual.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Querying for Eligible Liquidation Targets
+
+There is a query on Umee nodes that lists addresses that are eligible for liquidation.
```
robert-zaremba, line 27:
```suggestion
Umee provides a query method to lists addresses that are eligible for liquidation.
```

docs/liquidators/manual.md
```diff
@@ -0,0 +1,162 @@
...
+
+On the web, it looks like `https://api.your.node.here/umee/leverage/v1/liquidation-targets`
+
+Using the CLI, it is `umeed q leverage liquidation-targets`
+
+### Enabling the Query
```
robert-zaremba, line 33:
```suggestion
### Enabling the `liquidation-targets` Query
```

docs/liquidators/manual.md
```diff
@@ -0,0 +1,162 @@
...
+
+For performance reasons, this query is disabled by default:
+
+> Error: rpc error: code = Unknown desc = node has disabled liquidator queries
```
robert-zaremba, line 37: Can we fix the error code? `Unknown desc` sounds weird.

docs/liquidators/manual.md
```diff
@@ -0,0 +1,162 @@
...
+  denom: ibc/49788C29CD84E08D25CA7BE960BC1F61E88FEFC6333F58557D236D693398466A
+```
+
+Here (and with the help of `umeed q leverage registered-tokens` to interpret ibc denoms) we see that the user has borrowed `0.083 ATOM` with collateral of `1.077 USDC`.
+
+If the borrower had multiple tokens collateralized or borrowed, this would be the time to choose which ones to repay and receive as a reward. In this case, the only option for the liquidator is to repay `ATOM` and receive `USDC`.
+
+## Executing a Liquidation Transaction
+
+Our intent is to repay `0.083 ATOM` on the borrower's behalf, and receive `USDC` in return. However, our liquidator account doesn't currently have `ATOM` on the Umee chain - it only has collateralized `USDC` and some `UMEE` left over for gas.
```
robert-zaremba, line 101: let's add a note and example about the maximum repay. Position may not be eligible for full liquidation. Let's explain that, and provide a way how to calculate the maximum repay.

robert-zaremba (commented):
Let's break down the flow of minimum liquidation requirements from the flow of using external chains (osmosis, juno...) to exchange the collateral. The latter is not required and liquidators could use different strategies (eg use Axelar -> Polygon to do an exchange).

docs/liquidators/manual.md
toteki: It's docs for our chain operations, not just a module. We probably need a more central location for that kind of thing

docs/liquidators/manual.md
toteki: That's not the full picture - this manual covers the full operation because it is meant to help validators and other third parties get started.

docs/liquidators/manual.md
toteki: Unfortunately the `node has disabled liquidator queries` is all I control.

docs/liquidators/manual.md
robert-zaremba: The last thing we should have is "15+" (read many) places for documentation. We already have `design_docs` and `x/leverage`. The former is more for discussion, the latter is more about usage. Hence:
* Let's use `x/leverage` for everything `x/leverage` user related
* let's use `docs/...` for everything general (setting up a chain, general requirements).

docs/liquidators/manual.md
robert-zaremba: I know, but it reads that a user has to have other native tokens, which is not a hard requirements (again: user doesn't need to have native Osmo to liquidate AXL-USDC - this creates confusion when doing a first read). Hence my proposal to break it down, and firstly cover the basic usage, and in other section cover other recommending steps.

toteki: I'll try to modify the first section to make it clear that those tokens are for gas. I do still like the chronological order (setup, query, liquidate, sell) for this guide, since it allows the user to follow step by step.

docs/liquidators/manual.md
toteki: I'll move it to `x/leverage` - having a link in the root readme as well should keep it accessible enough.

docs/liquidators/manual.md
toteki: Added

docs/liquidators/manual.md
robert-zaremba: Chronological order is good. But I would still separate it in two flows. It will make more sense.

x/leverage/LIQUIDATION.md
```diff
@@ -0,0 +1,166 @@
...
+- `1 OSMO` on Osmosis
+
+The main collateral will allow you to borrow any tokens required during liquidation, and the other tokens are for gas on their native chains. (Gas tokens will vary based on which chains and exchanges are being used.)
+
+## Querying for Eligible Liquidation Targets
+
+Umee provides a query to list addresses that are eligible for liquidation.
+* REST: `https://api.your.node.here/umee/leverage/v1/liquidation-targets`
+* CLI: `umeed q leverage liquidation-targets`
+
```
toteki, line 30:
```suggestion

There is also a courtesy endpoint at `https://api.mainnet.network.umee.cc/umee/leverage/v1/liquidation_targets` which provides a cached output for the query every 5 minutes when available.

```

x/leverage/LIQUIDATION.md
```diff
@@ -0,0 +1,168 @@
...
+- `1 UMEE` on Umee
+- `1 ATOM` on Cosmos Hub
+- `1 AXL` on Axelar
+- `1 OSMO` on Osmosis
+
+The main collateral will allow you to borrow any tokens required during liquidation, and the other tokens are for gas on their native chains. (Gas tokens will vary based on which chains and exchanges are being used.)
+
+## Querying for Eligible Liquidation Targets
+
+Umee provides a query to list addresses that are eligible for liquidation.
```
robert-zaremba, line 27:
```suggestion
Umee provides a query to list addresses that are eligible for liquidation:
```

robert-zaremba (approved):

Issue #1759, opened 2023-01-30
Currency Provider Tracker coinIDSymbolMap


rbajollari:
## Summary of Bug

`coinIDSymbolMap` in currency provider tracker gets colliding IDs for symbols from CoinGecko (i.e. symbol `axl` has id `axelar` and `axl-inu`).

## Version

Please provide the output of the following commands:

- `$ umeed version`
- `$ go version`
- `$ uname -a`

## Steps to Reproduce

1. ...

----

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #1781, opened 2023-02-03
fix: currency provider tracker causing price feeder to error

rbajollari:
## Description

closes: #1759 

- prevent collision of coin id to coin symbol in `coinIDSymbolMap`
- don't stop the price feeder if coin gecko api request errors (log error and have minimum providers for the pair be 3)
- take out btc in example config
- change price-feeder timeout back to 1 minute in e2e test

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


price-feeder/price-feeder.example.toml
```diff
@@ -211,21 +207,6 @@ providers = [
...
-]
-quote = "USD"
-
-[[currency_pairs]]
-base = "BTC"
-providers = [
-  "huobi",
-  "kraken",
-]
-quote = "USDT"
```
robert-zaremba, line 227: why removing these important pair from kraken and coinbase?

price-feeder/config/currency_provider_tracker.go
```diff
@@ -168,7 +192,7 @@ func (t *CurrencyProviderTracker) setCurrencyProviders() error {
 // set to 3 providers.
 func (t *CurrencyProviderTracker) setCurrencyProviderMin() {
 	for base, exchanges := range t.CurrencyProviders {
-		if len(exchanges) < 3 {
+		if len(exchanges) < 3 && len(exchanges) > 0 { // can be 0 if request errored
 			t.CurrencyProviderMin[base] = len(exchanges)
 		} else {
 			t.CurrencyProviderMin[base] = 3
```
robert-zaremba, line 198: What if it's zero? Do we handle that case anywhere?

price-feeder/config/currency_provider_tracker.go
```diff
@@ -111,51 +117,69 @@ func (t *CurrencyProviderTracker) setCoinIDSymbolMap() error {
...
+		Timeout: requestTimeout,
+	}
+	t.osmosisAPIPairs = make(map[string]string)
 
-		var tickerResponse coinTickerResponse
-		if err = json.NewDecoder(coinGeckoResp.Body).Decode(&tickerResponse); err != nil {
+	osmosisResp, err := client.Get(fmt.Sprintf("%s/%s", osmosisV2RestURL, osmosisV2AssetPairsEndpoint))
+	if err != nil {
+		t.logger.Error().Err(err).Msg("Failed to query osmosis api assetpairs endpoint")
+	} else {
```
robert-zaremba, line 138:
```suggestion
		t.logger.Error().Err(err).Msg("Failed to query osmosis api assetpairs endpoint")
		return err
	} 
```

robert-zaremba: prefer early return, rather than if-else.
Also, need to remove one `}` at the end.

price-feeder/config/currency_provider_tracker.go
```diff
@@ -111,51 +117,69 @@ func (t *CurrencyProviderTracker) setCoinIDSymbolMap() error {
...
+	client := &http.Client{
+		Timeout: requestTimeout,
+	}
+	t.osmosisAPIPairs = make(map[string]string)
 
-		var tickerResponse coinTickerResponse
-		if err = json.NewDecoder(coinGeckoResp.Body).Decode(&tickerResponse); err != nil {
+	osmosisResp, err := client.Get(fmt.Sprintf("%s/%s", osmosisV2RestURL, osmosisV2AssetPairsEndpoint))
+	if err != nil {
+		t.logger.Error().Err(err).Msg("Failed to query osmosis api assetpairs endpoint")
```
robert-zaremba, line 137: why do we log here?

price-feeder/config/currency_provider_tracker.go
```diff
@@ -111,51 +117,69 @@ func (t *CurrencyProviderTracker) setCoinIDSymbolMap() error {
...
-		if err = json.NewDecoder(osmosisResp.Body).Decode(&assetPairsResponse); err != nil {
-			return err
+// setCurrencyProviders queries CoinGecko's tickers endpoint to get all the exchanges
+// that support each price feeder currency pair and store it in the CurrencyProviders map.
+func (t *CurrencyProviderTracker) setCurrencyProviders() error {
+	client := &http.Client{
+		Timeout: requestTimeout,
+	}
+	for _, pair := range t.pairs {
+		t.CurrencyProviders[pair.Base] = []string{}
```
robert-zaremba, line 160: What if this pair is already set? Do we want to overwrite?

price-feeder/config/currency_provider_tracker.go
```diff
@@ -111,51 +117,69 @@ func (t *CurrencyProviderTracker) setCoinIDSymbolMap() error {
...
+
+		// check if its a pair supported by the osmosis api
+		if t.osmosisAPIPairs[strings.ToUpper(pair.Base)] == strings.ToUpper(pair.Quote) {
+			t.CurrencyProviders[pair.Base] = append(t.CurrencyProviders[pair.Base], "osmosisv2")
```
robert-zaremba, line 164: I'm not sure about the flow you want to get here.
1. firstly we set currency providers here based on coinGeck data
2. but we also have osmosis provider, which we set here and in the function above. Why in two places? I would update `CurrencyProviders` when calling `setOsmosisAPIPairs` rather than making this weird relationship. Other idea would be to provide the osmosis pairs as the function argument, rather than keeping them in the struct.

price-feeder/config/currency_provider_tracker.go
```diff
@@ -80,6 +82,10 @@ func NewCurrencyProviderTracker(
 		return nil, err
 	}
 
+	if err := currencyProviderTracker.setOsmosisAPIPairs(); err != nil {
+		return nil, err
+	}
+
 	if err := currencyProviderTracker.setCurrencyProviders(); err != nil {
```
robert-zaremba, line 89:
```suggestion
	osmosisPairs, err := currencyProviderTracker.setOsmosisAPIPairs()
	if err != nil {
		return nil, err
	}

	if err := currencyProviderTracker.setCurrencyProviders(osmosisPairs); err != nil {
```

robert-zaremba: see comment below

robert-zaremba (commented):
let's rethink the flow

price-feeder/config/currency_provider_tracker.go
rbajollari: Sending in the osmosis pairs as a function arg now👍

price-feeder/config/currency_provider_tracker.go
rbajollari: Took it out after changing to early return

price-feeder/config/currency_provider_tracker.go
rbajollari: Agreed! If else is taken out

price-feeder/config/currency_provider_tracker.go
rbajollari: I added a log if the coin gecko response actually returns 0 providers without (not from erroring out). I think it should still default to a three provider minimum in this case, which is what would happen in the current implementation.

price-feeder/price-feeder.example.toml
rbajollari: Our accept list in the oracle is tracking the price of  `wbtc` not `btc`, so I don't see where the price data of `btc` would be getting used atm.

price-feeder/config/currency_provider_tracker.go
robert-zaremba: I'm still not sure why we cap it to 3, and how do we handle the zero case. Could you explain it please?

rbajollari: Three is just the minimum providers we enforced before supporting the lower cap assets which is why I'm setting that as a default minimum if the coingecko request for the providers supporting a certain pair errors out. In the case where coin gecko just returns 0 (not from erroring out), that would suggest that coin gecko knows no providers for the supplied pair. In that scenario, it makes sense to default our three provider minimum rule since we have no way of knowing how many providers support that pair if coin gecko doesn't have data on it. This is something that would most likely happen from someone putting in a pair that actually doesn't have any providers, so the log would inform them that coin gecko isn't aware of any providers supporting that pair.

price-feeder/price-feeder.example.toml
robert-zaremba: Could you add in changelog that we are removing BTC from the template, and explain why. Also add a note to the RELEASE_NOTES file in price-feeder.

price-feeder/config/currency_provider_tracker.go
robert-zaremba: OK, makes sense. Please add a comment in the code.

robert-zaremba (approved):
pre-approving. Let's add missing docs: changelog, release notes, comment about the zero case for min providers...

PR #1787, opened 2023-02-06
chore: remove rc upgrade handlers

robert-zaremba:
## Description

Let's cleanup "test" upgrade handlers. Especially that we are going to remove canon-2


adamewozniak (approved):
thanks for this, I forgot to clean up 🥳

gsk967 (approved):

PR #1793, opened 2023-02-07
feat: incentive module KVstore and structs

toteki:
## Description

Finished creating and moving over KVStore functions for incentive module.

A couple of things to notice:
- `utils/store` package now mimics `x/leverage/keeper/store.go`. It is fully used for incentive module, and in a seperate PR, we can get rid of that file and move leverage over to `util` version. (I didn't do it in this PR because scope is restricted)
  - Helper functions for un/marshaling types, like `store.SetInt`, `store.GetDec`, `store.SetAddress`
  - All functions enforce proper encoding and return errors on set (or panic on get)
  - Automatically enforce minimum values as specified by the module - for example never set an `Amount` field below zero
  - They produce neater error messages. For example: "0 is below the minimum InsterestScalar of 1" instead of "bad value". If we ever see a report one of these on mainnet or a testnet, debugging will be much much easier.
- The `keyPrefix` functions have been moved inside keeper and lowercased (unexported). This pattern was recommended by @robert-zaremba as a side note in today's test patterns workshop.
- Other than that, it's just getters, setters, and proto struct changes.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


util/store/store.go
```diff
@@ -0,0 +1,194 @@
+package store
+
```
toteki, line 2: This utilities package standardizes the functions used in `x/leverage/keeper/store.go`. It's being used for incentive module right away but I plan to move leverage over to it as well, in a separate PR.

x/incentive/keeper/keys.go
```diff
@@ -0,0 +1,124 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/address"
+
+	"github.com/umee-network/umee/v4/util"
+	"github.com/umee-network/umee/v4/x/incentive"
+)
+
+// KVStore key prefixes
+var (
```
toteki, line 14: moved to keeper package (outside of which they should not be used) and lowercased (unexported)

x/incentive/keeper/store.go
```diff
@@ -0,0 +1,274 @@
...
+
+// GetCommunityFundAddress retrieves the community fund address parameter. It is guaranteed to be
+// either valid (by sdk.ValidateAddressFormat) or empty.
+func (k Keeper) GetCommunityFundAddress(ctx sdk.Context) sdk.AccAddress {
+	return store.GetAddress(k.KVStore(ctx),
+		keyPrefixParamCommunityFundAddress, "community fund address")
+}
+
+// setParams validates and sets the incentive module parameters
+func (k Keeper) setParams(ctx sdk.Context, params incentive.Params) error {
```
toteki, line 54: Params are stored and fetched separately (efficiency), but set together (only happens via governance)

x/incentive/keeper/store.go
```diff
@@ -0,0 +1,274 @@
...
+		sdk.MustAccAddressFromBech32(params.CommunityFundAddress), "community fund address")
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// GetIncentiveProgram gets an incentive program by ID, regardless of the program's status.
+// Returns the program's status if found, or an error if it does not exist.
+func (k Keeper) GetIncentiveProgram(ctx sdk.Context, id uint32) (
```
toteki, line 99: This concept of storing incentive programs in separate lists will help iterate through them in practice - for any given query or function (e.g. EndBlocker), the module will want to iterate over upcoming, ongoing, and completed programs at different times.

The upcoming and ongoing lists in particular will benefit from not being cluttered by completed or cancelled programs.

x/incentive/keeper/store.go
```diff
@@ -0,0 +1,274 @@
...
+func (k Keeper) GetLastRewardsTime(ctx sdk.Context) uint64 {
+	return store.GetUint64(k.KVStore(ctx), keyPrefixLastRewardsTime, 0, "last reward time")
+}
+
+// setLastRewardsTime sets the last unix time incentive rewards were computed globally by EndBlocker.
+func (k Keeper) setLastRewardsTime(ctx sdk.Context, time uint64) error {
+	prev := k.GetLastRewardsTime(ctx)
+	if time < prev {
+		return incentive.ErrDecreaseLastRewardTime.Wrapf("%s to %s", time, prev)
+	}
```
toteki, line 173: Small enforcement of invariant here - it will be completely impossible, through this setter, to introduce bugs related to reversing block time.

x/incentive/keys.go
```diff
@@ -14,26 +14,19 @@ const (
...
+	BondTierLong BondTier = iota
+	BondTierMiddle
+	BondTierShort
+)
+
+const (
+	ProgramStatusUpcoming ProgramStatus = iota
+	ProgramStatusOngoing
+	ProgramStatusCompleted
 )
```
toteki, line 32: Helpful enumerations

x/leverage/keeper/internal_test.go
```diff
@@ -20,7 +19,6 @@ type TestKeeper struct {
 // an additional TestKeeper that exposes normally
 // unexported methods for testing.
 func NewTestKeeper(
-	require *require.Assertions,
```
toteki, line 23: lint (unused)

x/incentive/keeper/store.go
```diff
@@ -0,0 +1,275 @@
...
+func (k Keeper) GetTierWeightMiddle(ctx sdk.Context) sdk.Dec {
+	return store.GetDec(k.KVStore(ctx),
+		keyPrefixParamTierWeightMiddle, sdk.ZeroDec(), "middle tier weight")
+}
+
+// GetTierWeightShort gets the ratio of rewards received by the middle tier of bonded assets. Ranges 0 - 1.
+func (k Keeper) GetTierWeightShort(ctx sdk.Context) sdk.Dec {
+	return store.GetDec(k.KVStore(ctx),
+		keyPrefixParamTierWeightShort, sdk.ZeroDec(), "short tier weight")
+}
```
robert-zaremba, line 44: if you will need to read multiple params in a single request, then probably it's better to serialize all params as a single record (Params), rather than breaking it down into multiple records.

x/incentive/keeper/store.go
```diff
@@ -0,0 +1,275 @@
...
+func (k Keeper) SetIncentiveProgram(ctx sdk.Context,
+	program incentive.IncentiveProgram, status incentive.ProgramStatus,
+) error {
+	keys := [][]byte{
+		keyIncentiveProgram(program.Id, incentive.ProgramStatusUpcoming),
+		keyIncentiveProgram(program.Id, incentive.ProgramStatusOngoing),
+		keyIncentiveProgram(program.Id, incentive.ProgramStatusCompleted),
+	}
+
+	kvStore := ctx.KVStore(k.storeKey)
```
robert-zaremba, line 134:
```suggestion
	kvStore := k.KVStore(ctx)
```

x/incentive/keeper/store.go
```diff
@@ -0,0 +1,275 @@
...
+		incentive.ProgramStatusOngoing,
+		incentive.ProgramStatusCompleted,
+	}
+
+	kvStore := ctx.KVStore(k.storeKey)
+
+	// Looks for an incentive program with the specified ID in upcoming, ongoing, then completed program lists.
+	for _, status := range statuses {
+		program := incentive.IncentiveProgram{}
+		bz := kvStore.Get(keyIncentiveProgram(id, status))
```
robert-zaremba, line 113:
```suggestion
		bz := k.KVStore(ctx)
```

x/incentive/keeper/keeper.go
```diff
@@ -41,3 +41,8 @@ func (k Keeper) ModuleBalance(ctx sdk.Context, denom string) sdk.Coin {
 	amount := k.bankKeeper.SpendableCoins(ctx, authtypes.NewModuleAddress(incentive.ModuleName)).AmountOf(denom)
 	return sdk.NewCoin(denom, amount)
 }
+
```
robert-zaremba, line 44: would be good to use `Keeper` only for data access, and implement the logic in the msg_server or another struct.

robert-zaremba (approved):
pre-approving. Before merging:
- [ ] let's remove duplications by using go generics
- [ ] to make the storage utils more useful, let's make the "min" check param optional.
- [ ] no need to make checks during storage read - we don't do that in SDK.
- [ ] consider saving all params as a one struct (if in the logic implementation you usually need to read few params).

util/store/store.go
```diff
@@ -0,0 +1,194 @@
...
+// GetInt retrieves an sdkmath.Int from a KVStore, or a provided minimum value if no value is stored.
+// It panics if a stored value fails to unmarshal or is less than or equal to the minimum value.
+// Accepts an additional string which should describe the field being retrieved in custom error messages.
+func GetInt(store sdk.KVStore, key []byte, min sdkmath.Int, desc string) sdkmath.Int {
+	if bz := store.Get(key); bz != nil {
+		val := sdk.ZeroInt()
+		if err := val.Unmarshal(bz); err != nil {
+			panic(fmt.Sprintf("error unmarshaling %s into %T: %s", desc, val, err))
+		}
+		if val.IsNil() || val.LTE(min) {
```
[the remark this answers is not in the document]
toteki: fixed

util/store/store.go
```diff
@@ -0,0 +1,194 @@
...
+
+	sdkmath "cosmossdk.io/math"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	gogotypes "github.com/gogo/protobuf/types"
+)
+
+// GetInt retrieves an sdkmath.Int from a KVStore, or a provided minimum value if no value is stored.
+// It panics if a stored value fails to unmarshal or is less than or equal to the minimum value.
+// Accepts an additional string which should describe the field being retrieved in custom error messages.
+func GetInt(store sdk.KVStore, key []byte, min sdkmath.Int, desc string) sdkmath.Int {
```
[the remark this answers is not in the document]
toteki: Simplified instead by removing the parameter and using zero as the minimum

(I looked for cases where the minimum is nonzero, and found none for incentive module and only one for leverage: interest scalar minimum is `1.0`. Those can be handled uniquely if it keeper everything else simple.)

Also considered optional input using `...sdk.Dec` but ultimately decided against it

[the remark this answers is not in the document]
toteki: renamed

util/store/store.go
```diff
@@ -0,0 +1,194 @@
...
+			panic(fmt.Sprintf("error unmarshaling %s into %T: %s", desc, val, err))
+		}
+		if val.IsNil() || val.LTE(min) {
+			panic(fmt.Sprintf("%s is not above the minimum %s of %s", val, desc, min))
+		}
+		return val
+	}
+	// No stored bytes at key: return minimum value
+	return min
+}
```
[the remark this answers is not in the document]
toteki: turned this into a github issue since they resisted my initial attempts

(unmarshal uses a pointer receiver, the zero values `sdk.ZeroInt` and `sdk.ZeroDec` are not the same as initializing (nil) those types, and several other issues made it not worth the lines it would save)

x/incentive/errors.go
```diff
@@ -24,4 +24,11 @@ var (
 	ErrNonzeroRemainingRewards = errors.Register(ModuleName, 300, "remaining rewards must be zero in proposal")
 	ErrNonzeroFundedRewards    = errors.Register(ModuleName, 301, "funded rewards must be zero in proposal")
 	ErrEmptyProposal           = errors.Register(ModuleName, 302, "proposal contains no incentive programs")
+
+	// 4XX = Programs
+	ErrNoProgramWithID = errors.Register(ModuleName, 400, "no incentive program with specified ID")
```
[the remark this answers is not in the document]
toteki: removed

x/incentive/keeper/keys.go
```diff
@@ -0,0 +1,124 @@
...
+	default:
+		panic("invalid incentive program status in key")
+	}
+	return util.ConcatBytes(0, prefix, bz)
+}
+
+// keyTotalBonded returns a KVStore key for total bonded uTokens for a single tier.
+func keyTotalBonded(denom string, tier incentive.BondTier) []byte {
+	// totalBondedPrefix | denom | 0x00 | tier
+	return util.ConcatBytes(0, keyTotalBondedNoTier(denom), []byte{byte(tier)})
```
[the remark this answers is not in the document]
toteki: It might acquire a few more uses when we add more queries - specifically the kind that iterate over all three tiers of a denom

x/incentive/keeper/store.go
toteki: fixed (in 3 places)

util/store/store.go
```diff
@@ -0,0 +1,194 @@
...
+// It panics if a stored value fails to unmarshal or is less than or equal to the minimum value.
+// Accepts an additional string which should describe the field being retrieved in custom error messages.
+func GetInt(store sdk.KVStore, key []byte, min sdkmath.Int, desc string) sdkmath.Int {
+	if bz := store.Get(key); bz != nil {
+		val := sdk.ZeroInt()
+		if err := val.Unmarshal(bz); err != nil {
+			panic(fmt.Sprintf("error unmarshaling %s into %T: %s", desc, val, err))
+		}
+		if val.IsNil() || val.LTE(min) {
+			panic(fmt.Sprintf("%s is not above the minimum %s of %s", val, desc, min))
```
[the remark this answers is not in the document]
toteki: Some of these checks have disappeared now that the `minimum` is always zero (recent commit). The remaining error cases are:
- unmarshal fail (we can't remove this possibility)
- negative value (negative value should never have been set, but we can choose not to check it if we want)

My reasoning for keeping as-is:
- these panics are not expected (they should never happen unless we do something very wrong, like skip a migration or corrupt a database) so we don't need to add an `error` return to every getter
- the remaining `IsNegative` check is computationally cheap, and protects all downstream functions from exotic bugs while providing an exact diagnosis via error message

x/incentive/keeper/store.go
toteki: I'll have to think about this one - I'm thinking
- in most transactions only one param is needed
- in endblocker multiple params will be needed
- in non-iterating queries zero or one are likely to be needed (e.g. unless a person has locked multiple tiers of the same token)

Might keep as-is for now, but I can check up on it after implementation PRs

util/store/store.go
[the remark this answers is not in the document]
toteki: Note: marking unresolved so this is easier to spot

PR #1807, opened 2023-02-13
fix: 4.1 BNB migration

toteki:
This will switch the BNB ibc denom in the leverage registry and oracle accept list to the correct denom.


x/leverage/types/errors.go
```diff
@@ -24,6 +24,7 @@ var (
 	ErrDuplicateToken       = errors.Register(ModuleName, 207, "duplicate token")
 	ErrCollateralWeightZero = errors.Register(ModuleName, 206,
 		"collateral weight of Token is zero: can't be used as a collateral")
+	ErrBadSupplyDetected = errors.Register(ModuleName, 207, "supply of a bad token detected")
```
gsk967, line 27: `207` is already registered `ErrDuplicateToken = errors.Register(ModuleName, 207, "duplicate token")`, Can you change the errorCode

x/oracle/keeper/migrations.go
```diff
@@ -32,3 +32,22 @@ func (m Migrator) HistoracleParams3x4(ctx sdk.Context) error {
...
+	correctDenom := "ibc/8184469200C5E667794375F5B0EC3B9ABB6FF79082941BF5D0F8FF59FEBA862E"
+	acceptList := m.keeper.AcceptList(ctx)
+	for _, d := range acceptList {
+		// Switch the base denom of the token with changing anything else
+		if d.BaseDenom == badDenom {
+			d.BaseDenom = correctDenom
+		}
+	}
+	// Overwrite previous accept list
+	m.keeper.SetAcceptList(ctx, acceptList)
```
gsk967, line 48: `aceeptList` is not a pointer of list , so i think the values will not be override 

```suggestion
	for index := range acceptList {
		// Switch the base denom of the token with changing anything else
		if acceptList[index].BaseDenom == badDenom {
			 acceptList[index].BaseDenom = correctDenom
		}
	}
	// Overwrite previous accept list
	m.keeper.SetAcceptList(ctx, acceptList)
```

gsk967 (changes requested):
Can you check the error code and github lints

x/leverage/types/errors.go
toteki: fixed

x/oracle/keeper/migrations.go
toteki: applied suggestion

gsk967 (approved):
LGTM

x/leverage/keeper/migrations.go
```diff
@@ -0,0 +1,46 @@
...
+	store.Delete(badKey)
+	// Add back to store, but bypass the hooks in SetRegisteredToken
+	trueKey := types.KeyRegisteredToken(correctDenom)
+	bz, err := m.keeper.cdc.Marshal(&token)
+	if err != nil {
+		return err
+	}
+	store.Set(trueKey, bz)
+	return nil
+}
```
robert-zaremba, line 46: let's move this structure / file to `leverage/migrations`. This package is already overloaded.

robert-zaremba (commented):
We should avoid growing the keeper package.

x/leverage/keeper/migrations.go
robert-zaremba: moving to a new package shouldn't create any cycle, if there was no cycle before.

PR #1810, opened 2023-02-13
DNM: oracle keeper enforces uppercase symbol denoms

toteki:
This needs discussion to see if it's reasonable.

Adds a migration to turn all symbol denoms currently stored in the oracle keeper (generally as keys, but sometimes as values) into uppercase. Proceeds to enforce that rule in all `Set` functions.

### DIFF
--- a/x/oracle/abci_test.go
+++ b/x/oracle/abci_test.go
@@ -173,17 +173,17 @@ func (s *IntegrationTestSuite) TestEndBlockerVoteThreshold() {
 	// umee has 100% power, and atom has 30%
 	val1Tuples = types.ExchangeRateTuples{
 		types.ExchangeRateTuple{
-			Denom:        "umee",
+			Denom:        "UMEE",
 			ExchangeRate: sdk.MustNewDecFromStr("1.0"),
 		},
 	}
 	val2Tuples = types.ExchangeRateTuples{
 		types.ExchangeRateTuple{
-			Denom:        "umee",
+			Denom:        "UMEE",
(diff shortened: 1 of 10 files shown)


adamewozniak:
imo I don't think this is the way to go. For one, you should always write software assuming caps won't match when comparing strings. Also our oracle's settings are still in parameters, so we don't have custom gov handling logic.
cc @zarazan @rbajollari

toteki:
I'm hoping to solve some issues:
- Leverage (and soon ibc quota) module needs to know what denom (all caps, exact match to leverage registry, or something else?) to pass into functions like `GetExchangeRate` since it can't try multiple possibilities without calling the function multiple times. Currently historic prices use whatever the leverage module set first, and exchange rates always use uppercase.
- Oracle module is vulnerable to leverage shenanigans, like updating setting `stOSMO` vs `STOSMO` in the registry.
- Calling functions with the wrong case currently gives us "unknown denom", but if they were always stored as uppercase in the oracle `KVStore`, then any getter would also know what `key` to search for and convert to uppercase.

zarazan:
> imo I don't think this is the way to go. For one, you should always write software assuming caps won't match when comparing strings. Also our oracle's settings are still in parameters, so we don't have custom gov handling logic. cc @zarazan @rbajollari

I agree that if we don't ever need to support handling capitalization differences as different assets ("tOKEN" vs "TOKEN") then the oracle should store everything in the same case (all upper or all lower). Preferably we would enforce this requirement further upstream (such as when submitting proposals) instead of putting a ton of `toUpper` in the keepers but like Woz said, we don't have custom gov handling logic yet.

PR #1811, opened 2023-02-13
chore: simplify upgrades

robert-zaremba:
reuse `onlyModuleMigrations` in upgrade handler.


gsk967 (changes requested):
Why we have two upgrade handler for 4.1 
`registerUpgrade4_1`  and `registerUpgrade4_0_1` this two funs are not same ?

gsk967 (approved):
LGTM

PR #1812, opened 2023-02-13
fix: collateral share restriction works during partial price outages

toteki:
When some token prices are missing from the oracle, it becomes impossible to calculate a token's collateral share.

However, if the collateralized token's price is known, and enough other token prices are known to prove that it cannot possibly be over its `MaxCollateralShare`, then we can let `MsgSupplyCollateral` or `MsgCollateralize` succeed anyway.

### DIFF
--- a/x/leverage/keeper/collateral.go
+++ b/x/leverage/keeper/collateral.go
@@ -78,6 +78,33 @@ func (k Keeper) CalculateCollateralValue(ctx sdk.Context, collateral sdk.Coins) 
 	return total, nil
 }
 
+// VisibleCollateralValue uses the price oracle to determine the value (in USD) provided by
+// collateral sdk.Coins, using each token's uToken exchange rate. Always uses spot price.
+// Unlike CalculateCollateralValue, this function will not return an error if value calculation
+// fails on a token - instead, that token will contribute zero value to the total.
+func (k Keeper) VisibleCollateralValue(ctx sdk.Context, collateral sdk.Coins) sdk.Dec {
+	total := sdk.ZeroDec()
+
+	for _, coin := range collateral {
+		// convert uToken collateral to base assets
+		baseAsset, err := k.ExchangeUToken(ctx, coin)
+		if err != nil {
+			continue
+		}
+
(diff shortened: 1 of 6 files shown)


x/leverage/keeper/collateral.go
```diff
@@ -78,6 +78,33 @@ func (k Keeper) CalculateCollateralValue(ctx sdk.Context, collateral sdk.Coins)
...
+		// convert uToken collateral to base assets
+		baseAsset, err := k.ExchangeUToken(ctx, coin)
+		if err != nil {
+			continue
+		}
+
+		// get USD value of base assets
+		v, err := k.TokenValue(ctx, baseAsset, types.PriceModeSpot)
+		if err != nil {
+			continue
```
robert-zaremba, line 98: let's log this error with an appropriate message.

x/leverage/keeper/collateral.go
```diff
@@ -112,17 +139,16 @@ func (k Keeper) CollateralLiquidity(ctx sdk.Context, denom string) sdk.Dec {
...
 }
 
-// CollateralShare calculates the portion of overall collateral
-// (measured in USD value) that a given uToken denom represents.
-func (k *Keeper) CollateralShare(ctx sdk.Context, denom string) (sdk.Dec, error) {
+// VisibleCollateralShare calculates the portion of overall collateral (measured in USD value) that a
+// given uToken denom represents. If an asset other than the denom requested is missing an oracle
+// price, it ignores that asset's contribution to the system's overall collateral, thus potentially
+// overestimating the requested denom's collateral share while improving availability.
+func (k *Keeper) VisibleCollateralShare(ctx sdk.Context, denom string) (sdk.Dec, error) {
```
robert-zaremba, line 146: Do we need to rename this function? I would prefer to not rename it if possible.

robert-zaremba (approved):
pre-approving.

x/leverage/keeper/collateral.go
toteki: Added logging at info level so it doesn't flood the error level with noise

PR #1814, opened 2023-02-13
chore: update Quota storage

robert-zaremba:
## Description

No need to pack denom and quota in a struct. We can do it directly, and reduce number of types.


x/uibc/quota/keeper/quota.go
```diff
@@ -12,63 +13,66 @@ import (
...
-	var quotaOfIBCDenoms []uibc.Quota
-
+// GetAllQuotas returns quota of all tokens.
+func (k Keeper) GetAllQuotas(ctx sdk.Context) (sdk.DecCoins, error) {
+	var quotas sdk.DecCoins
 	prefix := uibc.KeyPrefixDenomQuota
 	store := ctx.KVStore(k.storeKey)
 
 	iter := sdk.KVStorePrefixIterator(store, prefix)
 	defer iter.Close()
```
gsk967, line 28: Can you do same like this (prefixStore) https://github.com/umee-network/umee/pull/1814/files#diff-0c1ecbc0c4238830ffea35cdd1e2f3efd4e2d0ec54d4f14e39ea6c3b4abcfac1R127-R129 , so we follow same style

gsk967 (approved):
Prr-Approving

x/uibc/quota/keeper/quota.go
robert-zaremba: done, actually there was a bug here

PR #1821, opened 2023-02-14
fix!: allow safe leverage operations during partial oracle outages

toteki:
## Description

This one does a lot of things:
- Allows `account_summary` to work during price outages, treating all unknown prices as zero.
  - In such cases, supplied value and other fields will appear lower than it really is, since some assets were skipped
  - Liquidation threshold will be null when it can't be computed, since there's no safe way to do that with missing prices
- `borrow_limit` in queries as well as messages is computed using only collateral with known prices
  - If the portion of your collateral with known prices is enough to cover a borrow, then it still works.
  - Same for withdraw and decollateralize
- `MaxWithdraw` and `MaxBorrow` (both queries and messages) now function with missing collateral prices, respecting the borrow limit policy above. The queries return zero on missing borrow prices, or a missing price for the specific token being asked for.
- `liquidation_targets` query skips addresses where liquidation threshold cannot be computed, instead of returning an error for the whole query
- `MsgLiquidate` will be able to function with some missing prices on the target's borrowed assets, if their other borrows are  high enough to still put them above their liquidation threshold

API Breaking:
- `liquidation_threshold` field in `account_summary` field can now be null

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


x/leverage/keeper/msg_server.go
```diff
@@ -330,6 +342,10 @@ func (s msgServer) MaxBorrow(
 		return nil, err
 	}
 
+	// If a price is missing for the borrower's collateral,
+	// but not this uToken or any of their borrows, error
```
toteki, line 346:
```suggestion
	// but not this token or any of their borrows, error
```

x/leverage/keeper/msg_server.go
```diff
@@ -330,6 +342,10 @@ func (s msgServer) MaxBorrow(
 		return nil, err
 	}
 
+	// If a price is missing for the borrower's collateral,
+	// but not this token or any of their borrows, error
+	// will be nil and the resulting value will be what
+	// can safely be withdrawn even with missing prices.
```
toteki, line 348:
```suggestion
	// can safely be borrowed even with missing prices.
```

x/leverage/keeper/borrows.go
```diff
@@ -127,6 +130,43 @@ func (k Keeper) CalculateBorrowLimit(ctx sdk.Context, collateral sdk.Coins) (sdk
...
+		// ignore blacklisted tokens
+		if !ts.Blacklist {
+			// get USD value of base assets using the chosen price mode
+			v, err := k.TokenValue(ctx, baseAsset, types.PriceModeLow)
+			if err == nil {
+				// if both spot and historic (if requeired) prices exist,
+				// add each collateral coin's weighted value to borrow limit
+				limit = limit.Add(v.Mul(ts.CollateralWeight))
+			}
+
```
robert-zaremba, line 163: let's add `else` and check `err`, in some cases we should return an error, rather than ignoring.

robert-zaremba: in other words, let's ignore if an error `Is` of a specific type.

x/leverage/keeper/grpc_query.go
```diff
@@ -320,11 +323,13 @@ func (q Querier) MaxWithdraw(
...
+		// but not this uToken or any of their borrows, error
+		// will be nil and the resulting value will be what
+		// can safely be withdrawn even with missing prices.
+		// On non-nil error here, max withdraw is zero.
 		uToken, err := q.Keeper.maxWithdraw(ctx, addr, denom)
-		if err != nil {
-			return nil, err
-		}
-		if uToken.IsPositive() {
+		if err == nil && uToken.IsPositive() {
```
robert-zaremba, line 332: if the `err` is not nil, then we should return or check it at least.

x/leverage/keeper/grpc_query.go
```diff
@@ -374,14 +379,15 @@ func (q Querier) MaxBorrow(
...
+		// but not this token or any of their borrows, error
+		// will be nil and the resulting value will be what
+		// can safely be borrowed even with missing prices.
+		// On non-nil error here, max borrow is zero.
 		maxBorrow, err := q.Keeper.maxBorrow(ctx, addr, denom)
-		if err != nil {
-			return nil, err
-		}
-		if maxBorrow.IsPositive() {
+		if err == nil && maxBorrow.IsPositive() {
```
robert-zaremba, line 388: ditto

x/leverage/keeper/iter.go
```diff
@@ -168,20 +168,16 @@ func (k Keeper) GetEligibleLiquidationTargets(ctx sdk.Context) ([]sdk.AccAddress
...
+		// to the next borrower instead of stopping the entire query
 		liquidationLimit, err := k.CalculateLiquidationThreshold(ctx, collateral)
-		if err != nil {
-			return err
-		}
-
-		// If liquidation limit is smaller than borrowed value then the
-		// address is eligible for liquidation.
-		if liquidationLimit.LT(borrowValue) {
+		if err == nil && liquidationLimit.LT(borrowValue) {
```
robert-zaremba, line 178: ditto -> err check

robert-zaremba (commented):
I'm not sure if this is a right solution.
I'm worried about  _borrowed value will skip borrowed tokens with unknown oracle prices, treating them as zero value_ -- this can prevent liquidations, which normally should happen.

Also, 
- [x] proper error checks are missing.
- [ ] let's add a follow up task to test new edge cases
- [ ] see , document or make a task to solve panic in case of missing collateral price when we try to liquidate a position.
- [ ] probably, this will also block uncollaterize operations.

robert-zaremba (approved):
pre-approving to make the planned released sooner.
I've updated the previous review note: https://github.com/umee-network/umee/pull/1821#pullrequestreview-1298514938
Let's make sure we add all relevant tasks, and follow ups in the task board.

x/leverage/keeper/borrows.go
toteki: I'll see if I can skip the three possible "missing price" errors:
- `decmath.Median`'s `ErrEmptyList` (from historic price mode)
- `leverage.ErrNoHistoricMedians` (same, but leverage side)
- `oracle.ErrUnknownDenom` (regular price mode)

toteki:
> see , document or make a task to solve panic in case of missing collateral price when we try to liquidate a position.

This is an intended error return - we can't declare a borrower unhealthy (liquidation eligible) without knowing the value of all their collateral.

I'll check off the rest on your comment as they are done

robert-zaremba:
> Decollateralize and withdraw both use assertBorrowerHealth, which requires the value of all borrows is known, and that the borrow limit derived from their collateral with known prices (some might be missing) is sufficient to cover the borrows.

We should enable a user to decollaterize the asset with zero price.

PR #1825, opened 2023-02-15
fix: skip BNB migration instead of error if no BNB denom present

toteki:
If the BNB denom is not present in the leverage registry, skips the BNB migration instead of causing an error.

This shouldn't affect mainnet or `canon-2`, but is good form in case we ever run this migration on a different testnet.

### DIFF
--- a/x/leverage/keeper/migrations.go
+++ b/x/leverage/keeper/migrations.go
@@ -29,7 +29,7 @@ func (m Migrator) MigrateBNB(ctx sdk.Context) (bool, error) {
 	}
 	token, err := m.keeper.GetTokenSettings(ctx, badDenom)
 	if err != nil {
-		ctx.Logger().Info("leverage migration skipped due to missing token", "err", err.Error())
+		ctx.Logger().Error("leverage migration skipped due to missing token", "err", err.Error())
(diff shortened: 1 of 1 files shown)


x/leverage/keeper/migrations.go
```diff
@@ -29,7 +29,9 @@ func (m Migrator) MigrateBNB(ctx sdk.Context) (bool, error) {
 	}
 	token, err := m.keeper.GetTokenSettings(ctx, badDenom)
 	if err != nil {
-		return false, err
+		ctx.Logger().Error("leverage migration skipped due to missing token", "err", err.Error())
+		// If it's not a registered token, then we don't need to run this migration
+		return false, nil
```
robert-zaremba, line 34: Is there other reason when the error is returned then not found? If yes then let's check the error before nullifying it

robert-zaremba (approved):

x/leverage/keeper/migrations.go
toteki: Since this line skips / aborts the migration and logs it at error level, it's already the same treatment that an unexpected error (unmarshal fail for example) would get

PR #1832, opened 2023-02-16
fix: upgrade plan name to v4.1.0

toteki:


robert-zaremba (approved):
We will have a bit of inconsistency with other upgrade names...

gsk967 (approved):

PR #1834, opened 2023-02-16
Move IBC out from the experimental build

robert-zaremba:


robert-zaremba:
OK, thanks for the check!

gsk967 (approved):
LGTM

PR #1835, opened 2023-02-16
test: use experimental for default CI tests

robert-zaremba:
## Description

Experimental tests should be run by default and we should not break the code, even for the experimental features.


toteki (approved):
Looks good

PR #1840, opened 2023-02-17
chore(CI): set merge groups

robert-zaremba:
## Description

+ sets `merge-groups` targets in the workflows, to enable merge-group trigger to pass the required jobs
+ removes some push targes, which waist unnecessary CI cycles on merges to main. We don't need them, because we have all that checks on PR, and direct pushing to `main` is blocked anyway.


gsk967 (approved):
LGTM

PR #1850, opened 2023-02-18
chore: update upgrades structure

robert-zaremba:
## Proposal

In last two upgrades we had / have a sequence of rc releases, each of which is a breaking change, and require new upgrade handler. This is not relevant to the mainnet. Instead I propose:
1. Use `betaX` rather than `rcX`.
1. keep single handler and rename until we have successful, stable RC

### DIFF
--- a/app/upgrades.go
+++ b/app/upgrades.go
@@ -35,13 +35,12 @@ func (app UmeeApp) RegisterUpgradeHandlers(bool) {
 	app.registerUpgrade3_3to4_0(upgradeInfo)
 	app.registerUpgrade4_0_1(upgradeInfo)
 	app.registerUpgrade4_1(upgradeInfo)
-	app.registerUpgrade4_1_0rc3(upgradeInfo)
 	app.registerUpgrade4_2(upgradeInfo)
 }
 
 // performs upgrade from v4.1 to v4.2
 func (app *UmeeApp) registerUpgrade4_2(upgradeInfo upgradetypes.Plan) {
-	const planName = "v4.2"
+	const planName = "v4.2-beta1"
 	app.UpgradeKeeper.SetUpgradeHandler(planName, onlyModuleMigrations(app, planName))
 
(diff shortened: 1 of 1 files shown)


app/upgrades.go
```diff
@@ -35,13 +35,12 @@ func (app UmeeApp) RegisterUpgradeHandlers(bool) {
...
 	app.registerUpgrade4_0_1(upgradeInfo)
 	app.registerUpgrade4_1(upgradeInfo)
-	app.registerUpgrade4_1_0rc3(upgradeInfo)
 	app.registerUpgrade4_2(upgradeInfo)
 }
 
 // performs upgrade from v4.1 to v4.2
 func (app *UmeeApp) registerUpgrade4_2(upgradeInfo upgradetypes.Plan) {
-	const planName = "v4.2"
+	const planName = "v4.2-beta1"
```
robert-zaremba, line 43: @gsk967 for the IBC-quota testnet, the upgrade name has to match this string.

Later we will rename it as described in the PR details.

app/upgrades.go
```diff
@@ -51,12 +50,6 @@ func (app *UmeeApp) registerUpgrade4_2(upgradeInfo upgradetypes.Plan) {
 	})
 }
 
-// performs upgrade from v4.1.0-rc2 to v4.1.0-rc3
-func (app *UmeeApp) registerUpgrade4_1_0rc3(_ upgradetypes.Plan) {
-	const planName = "v4.1.0-rc3"
-	app.UpgradeKeeper.SetUpgradeHandler(planName, onlyModuleMigrations(app, planName))
-}
```
robert-zaremba, line 58: @toteki , this is not relevant to the main branch, should be only in the release branch, as the main branch is already ahead.

gsk967 (approved):

toteki:
This works, escpecially if we're doing testnets where the upgrade is run then later updated in main

PR #1865, opened 2023-02-22
chore(uibc): update quota -> outflows

robert-zaremba:


x/uibc/module/abci.go
```diff
@@ -14,7 +14,7 @@ func BeginBlock(ctx sdk.Context, keeper keeper.Keeper) {
 
 	// reset quotas
 	if quotaExpires == nil || quotaExpires.Before(ctx.BlockTime()) {
-		util.Panic(keeper.ResetQuota(ctx))
+		util.Panic(keeper.ResetAllQuotas(ctx))
```
gsk967, line 17: May be can you rename to `ResetAllOutflows` instead of `ResetAllQuotas`

x/uibc/quota/keeper/grpc_query.go
```diff
@@ -28,23 +28,23 @@ func (q Querier) Params(goCtx context.Context, _ *uibc.QueryParams) (
 	return &uibc.QueryParamsResponse{Params: params}, nil
 }
 
-// Quota returns quotas of denoms.
-func (q Querier) Quota(goCtx context.Context, req *uibc.QueryQuota) (
-	*uibc.QueryQuotaResponse, error,
+// Outflows returns quotas of denoms.
```
gsk967, line 31:
```suggestion
// Outflows returns quota outflow of denoms.
```

gsk967 (changes requested):

x/uibc/module/abci.go
robert-zaremba: Actually, here `ResetAllQuotas` makes more sense. Outflow is an aggregate.

x/uibc/quota/keeper/grpc_query.go
robert-zaremba: should be a bit different. I will update

robert-zaremba:
@gsk967 you should not block a PR if your suggestions are only about updating code comments.

gsk967 (approved):
LGTM

Issue #1878, opened 2023-02-27
Move client code to a new Go module


robert-zaremba:
## Summary

Many parts of our application stack require a grpc client. Doing this "manually" can be either tedious and lot of copy-pasta spaghetti. And we have that across other chains.

## Goal

Extract e2e client code and make it universal.

PR #1888, opened 2023-03-05
feat: Client implementation

robert-zaremba:
## Description

Closes: https://github.com/umee-network/umee/issues/1878 

Concept:
* have a common client which can be used in e2e tests and external tools
* provide separate modules:
  * first providing SDK client methods, and 
  * second: providing umee specific methods
* the first package is a separate Go module and can be imported by other chains.


toteki:
Should fix those missing error checks - they'll lead to panics on (for example) `resp.Medians` when `resp == nil`

robert-zaremba:
Yes, will look at this nil checks tomorrow. However this should not block the review.

robert-zaremba:
up, waiting for review

client/client.go
```diff
@@ -0,0 +1,36 @@
...
+// Initializes a cosmos sdk client context and transaction factory for
+// signing and broadcasting transactions
+func NewClient(
+	chainID,
+	tmrpcEndpoint,
+	grpcEndpoint,
+	accountName,
+	accountMnemonic string,
+	gasAdjustment float64,
+	ecfg sdkparams.EncodingConfig,
```
zarazan, line 23: Naming this variable `encodingConfig` would be a lot more readable.

sdkclient/client.go
```diff
@@ -0,0 +1,49 @@
...
+	"time"
+
+	sdkparams "github.com/cosmos/cosmos-sdk/simapp/params"
+	"github.com/rs/zerolog"
+	rpcclient "github.com/tendermint/tendermint/rpc/client"
+	"github.com/umee-network/umee/v4/sdkclient/query"
+	"github.com/umee-network/umee/v4/sdkclient/tx"
+)
+
+// OjoClient is a helper for initializing a keychain, a cosmos-sdk client context,
```
zarazan, line 14: `SdkClient`

client/client.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"context"
+
+	sdkparams "github.com/cosmos/cosmos-sdk/simapp/params"
+	"github.com/umee-network/umee/v4/sdkclient"
+)
+
+type Client struct {
+	sdkclient.Client
```
zarazan, line 11: I like this idea, I think this is one of the better ways in Golang to mimic inheritance.

client/client.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// Initializes a cosmos sdk client context and transaction factory for
+// signing and broadcasting transactions
```
zarazan, line 15: You could add a little something about how this Client inherits the queries and transactions of a normal Cosmos SDK chain with additional tx/queries specific to the Umee chain.

zarazan (commented):
I like what you did with cleaning up the encoding and separating the two Clients. Just a couple naming nit picks. I look forward to pulling this into its own package!

zarazan (approved):

client/client.go
```diff
@@ -0,0 +1,36 @@
+package client
```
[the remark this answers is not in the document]
robert-zaremba: it's already in the umee repo. The generic client pkg name is `sdkclient`. `Ojo` client will be `ojo/client`. What do you think?

client/client.go
robert-zaremba: yes!

client/client.go
robert-zaremba: OK

client/client.go
zarazan: Because for someone reviewing or modifying the code for the first time seeing "ecfg" 100 lines down from the declaration doesn't mean anything to me. I would have to scroll up to check what that variable is supposed to be. If there are a couple more variables like that it becomes more unreadable. I wouldn't call it "repeating". I'd call it "not abbreviating past the point of readability". IMO the extra few characters are worth readability and maintainability but this is not a blocker.

zarazan (approved):

client/client.go
robert-zaremba: Using `encCfg` which is also used in Cosmos SDK

zarazan (approved):

PR #1894, opened 2023-03-06
feat: incentive module non-gov messages

toteki:
Implements all remaining message types on incentive module, and a great deal of internal workings.

Still missing:
- endblocker
- hooks from leverage module on liquidate, reduce collateral amount
- reward math implementation (and expected exponent fix)
- some iterators
- extensive testing


proto/umee/incentive/v1/incentive.proto
```diff
@@ -10,7 +10,8 @@ option go_package = "github.com/umee-network/umee/v4/x/incentive";
 message Params {
   option (gogoproto.equal) = true;
 
-  // max_unbondings defines the maximum amount of concurrent unbondings an address can have.
+  // max_unbondings defines the maximum amount of concurrent unbondings an address can have
+  // on each unbonding tier of each bonded uToken denom.
```
toteki, line 14: This change will improve both UX and implementation requirements, but results in a lower default value.

x/incentive/expected_keepers.go
```diff
@@ -20,4 +20,5 @@ type LeverageKeeper interface {
 	GetCollateralAmount(ctx sdk.Context, borrowerAddr sdk.AccAddress, denom string) sdk.Coin
 	IsAcceptedUToken(ctx sdk.Context, uTokenDenom string) bool
 	AssertNotBlacklisted(ctx sdk.Context, denom string) error
+	// TODO: hooks
```
toteki, line 23: Similar to how oracle module registers leverage hooks to be run after adding tokens, incentive module will need to register leverage hooks to be run after liquidating or decollateralizing assets.

x/incentive/keeper/bond.go
```diff
@@ -0,0 +1,99 @@
...
+	time := k.GetLastRewardsTime(ctx)
+	storedUnbondings := k.GetUnbondings(ctx, addr)
+	for _, u := range storedUnbondings {
+		if tier == incentive.BondTierUnspecified || u.Tier == uint32(tier) {
+			// match both tier (if specified) and denom
+			if u.Amount.Denom == denom {
+				if u.End > time {
+					// this unbonding is still ongoing, and can be counted normally
+					unbondings = append(unbondings, u)
+					unbonding = unbonding.Add(u.Amount)
```
toteki, line 45: User-facing queries need to reflect bonded amounts when unbondings end, while at the same time not actually clearing those unbondings from state (would break consensus).

Unbondings are removed then next time a message affects account state.

x/incentive/keeper/bond.go
```diff
@@ -0,0 +1,99 @@
...
+		return err
+	}
+	// update module total
+	total := k.GetTotalBonded(ctx, bond.Denom, tier)
+	return k.SetTotalBonded(ctx, total.Add(bond), tier)
+}
+
+// decreaseBond decreases the bonded uToken amount for an account at a given tier, and updates the module's total.
+// it also clears reward trackers for any bond tiers which have reached zero for the account, to save storage.
+// This token must be called when unbondings are completed (not created) and when the leverage module forcefully
```
toteki, line 81:
```suggestion
// This function must be called when unbondings are completed (not created) and when the leverage module forcefully
```

x/incentive/keeper/bond.go
toteki: Also useful for functions which want to count bonded _but not currently unbonding_ tokens (since in state, unbonding tokens are considerd bonded until the unbonding _completes_)

x/incentive/keeper/update.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/umee-network/umee/v4/x/incentive"
+)
+
+// updateAccount finishes any unbondings associated with an account which have ended and claims any pending rewards.
+// It returns the amount of rewards claimed.
+//
+// REQUIREMENT: This function must be called during any message or hook which creates an unbonding or updates
+// bonded amounts. Leverage hooks which decrease borrower collateral must also call this before acting.
+// This ensures that between any two consecutive claims by a single account, bonded amounts were constant
+// for each bond tier and collateral uToken denom.
```
toteki, line 15:
```suggestion
// on that account for each bond tier and collateral uToken denom.
```

x/incentive/params.go
```diff
@@ -9,7 +9,7 @@ import (
 const secondsPerDay = 86400
 
 var (
-	defaultMaxUnbondings           = uint32(20)
+	defaultMaxUnbondings           = uint32(5)
```
toteki, line 12: This is not per-denom-and-tier, rather than account-wide. Allows smaller default value, and also scales better with large number of listed assets.

x/incentive/keeper/bond.go
```diff
@@ -0,0 +1,101 @@
...
+// accountBonds gets the total bonded and unbonding for a given account, as well as a list of ongoing unbondings,
+// for a single bond tier and uToken denom. Using incentive.BondTierUnspecified returns all tiers instead.
+// It reduces the returned bonded amount by completed unbondings without actually clearing those unbondings in state,
+// so it is safe for queries and any parts of Msg functions which are not intended to alter state.
+func (k Keeper) accountBonds(ctx sdk.Context, addr sdk.AccAddress, denom string, tier incentive.BondTier) (
+	bonded sdk.Coin, unbonding sdk.Coin, unbondings []incentive.Unbonding,
+) {
+	bonded = sdk.NewCoin(denom, sdk.ZeroInt())
+	unbonding = sdk.NewCoin(denom, sdk.ZeroInt())
+	unbondings = []incentive.Unbonding{}
```
gsk967, line 27: can you make 
`unbondings = make([]incentive.Unbonding{},0)`

x/incentive/keeper/bond.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// accountBonds gets the total bonded and unbonding for a given account, as well as a list of ongoing unbondings,
+// for a single bond tier and uToken denom. Using incentive.BondTierUnspecified returns all tiers instead.
+// It reduces the returned bonded amount by completed unbondings without actually clearing those unbondings in state,
+// so it is safe for queries and any parts of Msg functions which are not intended to alter state.
+func (k Keeper) accountBonds(ctx sdk.Context, addr sdk.AccAddress, denom string, tier incentive.BondTier) (
+	bonded sdk.Coin, unbonding sdk.Coin, unbondings []incentive.Unbonding,
+) {
+	bonded = sdk.NewCoin(denom, sdk.ZeroInt())
```
gsk967, line 25: for zero coin can you use 
`coin.Zero(denom)` from util package, 
same for all zero valued coins

gsk967 (approved):
pre approving

x/incentive/keeper/bond.go
robert-zaremba: Why? I think the `[]incentive.Unbonding{}` is more idiomatic.

x/incentive/expected_keepers.go
robert-zaremba: Is this to remove user from a plan?

x/incentive/keeper/bond.go
```diff
@@ -0,0 +1,101 @@
...
+	unbonding = sdk.NewCoin(denom, sdk.ZeroInt())
+	unbondings = []incentive.Unbonding{}
+
+	// sum all bonded tokens for this denom, for the specified tier or across all tiers
+	for _, t := range []incentive.BondTier{incentive.BondTierLong, incentive.BondTierMiddle, incentive.BondTierShort} {
+		if tier == incentive.BondTierUnspecified || t == tier {
```
robert-zaremba, line 31:
```suggestion
		if tier == incentive.BondTierUnspecified || tier == t {
```

robert-zaremba: for consistency

x/incentive/keeper/bond.go
```diff
@@ -0,0 +1,101 @@
...
+			bonded = bonded.Add(k.GetBonded(ctx, addr, denom, t))
+		}
+	}
+
+	time := k.GetLastRewardsTime(ctx)
+	storedUnbondings := k.GetUnbondings(ctx, addr)
+	for _, u := range storedUnbondings {
+		if tier == incentive.BondTierUnspecified || u.Tier == uint32(tier) {
```
robert-zaremba, line 39: Maybe tire should be a byte? We won't have more than 256 tires

toteki: It's being kept as a byte in KVStore, but I seem to have found uint32 easier to use in proto. I'll go back and check to see if that still makes sense though

x/incentive/keeper/bond.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	// if the new bond for this account + tier + denom has reached zero, it is safe to stop tracking rewards
+	if bonded.IsZero() {
+		err := k.clearRewardTracker(ctx, addr, tier, unbond.Denom)
+		if err != nil {
+			return err
+		}
+	}
+	// update module total
+	total := k.GetTotalBonded(ctx, unbond.Denom, tier)
```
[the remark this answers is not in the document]
toteki: Shouldn't be, since total bonded includes unbonding and bonded tokens for all users combined

x/incentive/expected_keepers.go
toteki: Possibly - it can either instantly unbond bonded tokens, partially complete an ongoing unbonding by reducing its amount, or completely finish some ongoing unbondings.

In general, bonded tokens must be liquidated instantly regardless of their status in the incentive module so this will need to keep any numbers related to bonds or unbodings up to date.

x/incentive/keeper/reward.go
```diff
@@ -0,0 +1,46 @@
...
+func (k Keeper) updateRewardTracker(ctx sdk.Context, addr sdk.AccAddress, tier incentive.BondTier, bondDenom string,
+) error {
+	trackers := k.getFullRewardTracker(ctx, addr, bondDenom, tier)
+	accumulators := k.getFullRewardAccumulator(ctx, bondDenom, tier)
+	for _, rewardCoin := range trackers {
+		accumulator := sdk.NewDecCoinFromDec(rewardCoin.Denom, accumulators.AmountOf(rewardCoin.Denom))
+		if err := k.SetRewardTracker(ctx, addr, bondDenom, accumulator, tier); err != nil {
+			return err
+		}
+	}
```
robert-zaremba, line 34: this is hard to review. Not sure why do we need it: the `getFullRewardAccumulator` implementation is missing.

robert-zaremba (commented):
I think we will need to make another algorithm and implementation review once the implementation is ready. Few key functions is missing, so hard to review the algorithm

x/incentive/keeper/bond.go
toteki: fixed

x/incentive/keeper/reward.go
EbrahimUmee: Moser is aware, he will implement and made new pr.

EbrahimUmee (approved):

robert-zaremba (approved):
Thanks for updating the PR. We will need to do another review once the functionality is completed.

PR #1896, opened 2023-03-08
feat(uibc): support total outflows in outflows query

robert-zaremba:
## Description
* Change OutflowsResponse to return a single dec, rather than a list of coins
* when denom is not specified, total outflows is returned.


proto/umee/uibc/v1/query.proto
```diff
@@ -17,10 +16,10 @@ service Query {
     option (google.api.http).get = "/umee/uibc/v1/params";
   }
 
-  // Outflow returns IBC denom outflows in the current epoch.
-  // If denom is empty, returns outflows of all tokens in the current epoch.
+  // Outflow returns IBC denom outflows in the current quota period.
+  // If denom is not specified, returns sum of all registered outflows.
   rpc Outflows(QueryOutflows) returns (QueryOutflowsResponse) {
-    option (google.api.http).get = "/umee/uibc/v1/outflows";
+    option (google.api.http).get = "/umee/uibc/v1/outflows/{denom}";
```
gsk967, line 22: Can you remove `denom` from req path, because we are storing the outflow value with `ibc/HASH` so it will fail 

that denom will be req query , req is  `/umee/uibc/v1/outflows?denom=ibc/HASH`

robert-zaremba: Good catch. Yes!

gsk967 (approved):
LGTM

PR #1903, opened 2023-03-10
DNM: chore: reduce not used exported consts

robert-zaremba:
## Description

Reduce boilerplate of consts and defaults which are used only once , or private.


gsk967 (approved):
pre approving 
Can you do same for oracle and leverage

robert-zaremba:
>  Can you do same for oracle and leverage
done

toteki:
Looks good on incentive & leverage side

PR #1930, opened 2023-03-18
fix: codecov file types

toteki:
We were including `pb.gw.go` files in our coverage.

Note that I had to add some code changes to trigger the coverage report, so I grabbed some trivial ones with tests from an incentive module PR.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -46,6 +46,7 @@ ignore:
   - "**/*.json"
   - "**/*.toml"
   - "**/*.rst"
+  - "**/*.md"
 # auto-generated files
   - "**/*.pb.go"
   - "**/*.pb.gw.go"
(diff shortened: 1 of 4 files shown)


.codecov.yml
```diff
@@ -40,14 +40,18 @@ flags:
       - "x/**/client/"
 
 ignore:
-  - "docs"
-  - "*.md"
```
robert-zaremba, line 44: docs and .md files should be excluded as well

robert-zaremba (approved):
pre-approving

.codecov.yml
```diff
@@ -40,14 +40,18 @@ flags:
...
-  - "*.rst"
+# ignore all files of these types
+  - "**/*.proto"
+  - "**/*.yml"
+  - "**/*.json"
+  - "**/*.toml"
+  - "**/*.rst"
```
toteki, line 48:
```suggestion
  - "**/*.rst"
  - "**/*.md"
```

PR #1946, opened 2023-03-23
Robert/testing sdk

robert-zaremba:
## Description

* added helper functions to create a test Cosmos SDK Context and KVStore
* demonstrated the usage in:
  * converted uibc quota test to an unit test
  * missing test for the `Iterate` function.

### DIFF
--- a/tests/tsdk/store.go
+++ b/tests/tsdk/store.go
@@ -28,12 +28,17 @@ func NewCommitMultiStore(t *testing.T, keys []types.StoreKey, tkeys []types.Stor
 	return cms
 }
 
-// NewCtx creates new context with store and mounted store keys.
+// NewCtx creates new context with store and mounted store keys and transient store keys.
 func NewCtx(t *testing.T, keys []types.StoreKey, tkeys []types.StoreKey) (sdk.Context, types.CommitMultiStore) {
 	cms := NewCommitMultiStore(t, keys, tkeys)
 	ctx := sdk.NewContext(cms, tmproto.Header{}, false, log.NewNopLogger())
 
 	return ctx, cms
+}
+
+// NewCtxq creates new context with only one store key
+func NewCtxOneStore(t *testing.T, keys types.StoreKey) (sdk.Context, types.CommitMultiStore) {
+	return NewCtx(t, []types.StoreKey{keys}, nil)
 }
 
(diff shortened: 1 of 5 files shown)


tests/tsdk/store.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/cosmos/cosmos-sdk/store"
+	types "github.com/cosmos/cosmos-sdk/store/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/tendermint/tendermint/libs/log"
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
+	dbm "github.com/tendermint/tm-db"
+)
+
+// NewCommitMultiStore creats SDK Multistore
```
toteki, line 16: Is there a way to document (or link to docs) the purpose of tkeys and keys for those unfamiliar with the SDK?

robert-zaremba: the type already provides that. Also, we are having it in every module we are implementing.

util/store/iter_test.go
```diff
@@ -0,0 +1,37 @@
...
+		{[]byte{1, 8}, []byte{2}},
+		{[]byte{2, 4}, []byte{1}},
+	}
+
+	db := tsdk.NewKVStore(t)
+	db.Set(pairs[0].K, pairs[0].V)
+	db.Set(pairs[4].K, pairs[4].V)
+	db.Set(pairs[3].K, pairs[3].V)
+	db.Set(pairs[1].K, pairs[1].V)
+	db.Set(pairs[2].K, pairs[2].V)
```
toteki, line 27: for / range loop might make sense here, or a comment indicating why it won't (e.g. `pairs` is sorted but we want to `db.Set` the keys out of order to properly test iterator)

robert-zaremba: BTW, would be great if you add more edge cases and also add a test to the reverse iterator. I'm also planning to write a generic fetcher, but that will be a new feature.

gsk967 (approved):
LGTM

util/store/iter_test.go
robert-zaremba: @toteki I will create a task to add missing tests. This PR is meant to provide the helper functions I mentioned with example how to use it.

toteki (approved):

PR #1956, opened 2023-03-27
feat(leverage): liquidation threshold must be bigger than collateral_weight

robert-zaremba:
## Description

* We must not accept `liquidation_threshold == collateral_weight` because that will trigger immediate liquidation.


x/leverage/keeper/msg_server.go
```diff
@@ -195,10 +195,6 @@ func (s msgServer) Collateralize(
 		return nil, err
 	}
 
-	if err := s.keeper.checkCollateralLiquidity(ctx, types.ToTokenDenom(msg.Asset.Denom)); err != nil {
-		return nil, err
-	}
-
```
toteki, line 201: We cannot remove this: `MinCollateralLiquidity` must prevent adding collateral when that would reduce the module's collateral liquidity below required value.

toteki (changes requested):

toteki (approved):
pre-approving

PR #1957, opened 2023-03-28
fix: amount reserved per block rounds up

toteki:
This will help with reserves in cases where interest accrued is very small, and would otherwise be rounded down to zero per block.


CHANGELOG.md
```diff
@@ -49,6 +49,7 @@ Ref: https://keepachangelog.com/en/1.0.0/
 ### Fixes
 
 - [1929](https://github.com/umee-network/umee/pull/1929) Leverage: `MaxWithdraw` now accounts for `MinCollateralLiquidity`
+- [XXXX](https://github.com/umee-network/umee/pull/XXXX) Leverage: Reserved amount per block now rounds up.
```
toteki, line 52:
```suggestion
- [1957](https://github.com/umee-network/umee/pull/1957) Leverage: Reserved amount per block now rounds up.
```

x/leverage/keeper/interest.go
```diff
@@ -133,11 +133,14 @@ func (k Keeper) AccrueAllInterest(ctx sdk.Context) error {
 			interestAccrued.TruncateInt(),
 		))
 
-		// calculate new reserves accrued for this denom
-		newReserves = newReserves.Add(sdk.NewCoin(
-			token.BaseDenom,
-			interestAccrued.Mul(token.ReserveFactor).TruncateInt(),
-		))
+		// if interest accrued on this denom is at least one base token
+		if interestAccrued.GT(sdk.OneDec()) {
```
robert-zaremba, line 137: shouldn't be `GT 0`?

robert-zaremba:
```suggestion
		if interestAccrued.GT(sdk.ZeroDec()) {
```

x/leverage/keeper/interest.go
```diff
@@ -133,11 +133,14 @@ func (k Keeper) AccrueAllInterest(ctx sdk.Context) error {
 			interestAccrued.TruncateInt(),
```
robert-zaremba, line 133: here we truncate (take a floor). Potentially `totalInterest` won't much accumulated

x/leverage/keeper/interest.go
```diff
@@ -133,11 +133,14 @@ func (k Keeper) AccrueAllInterest(ctx sdk.Context) error {
...
-		newReserves = newReserves.Add(sdk.NewCoin(
-			token.BaseDenom,
-			interestAccrued.Mul(token.ReserveFactor).TruncateInt(),
-		))
+		// if interest accrued on this denom is at least one base token
+		if interestAccrued.GT(sdk.OneDec()) {
+			// calculate new reserves gained for this denom, rounding up
+			newReserves = newReserves.Add(sdk.NewCoin(
+				token.BaseDenom,
+				interestAccrued.Mul(token.ReserveFactor).Ceil().TruncateInt(),
```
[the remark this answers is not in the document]
toteki: `Ciel()` operates first (but returns `sdk.Dec` so it must be truncated). I'll change from `Ceil().TruncateInt()` to `Ceil().RoundInt()` to be more clear

x/leverage/keeper/interest.go
toteki: This decimal represents base tokens, so if it's `0.5` (i.e. `0.5uumee`) it must be treated as `0uumee`, not `1uumee` to prevent more reserves from being allocated than interest.

(in that case, uToken exchange rate would decrease which we should not permit)

x/leverage/keeper/interest.go
toteki: This line specifically is for logging, so it's probably ok, but there is still a danger of interest per block being less than one base token, in which case it must be treated as zero.

x/leverage/keeper/interest.go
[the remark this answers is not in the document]
robert-zaremba: ah, I didn't notice. Let's keep `Ciel().TruncateInt()` (it's more efficient) and add a comment why we have truncate int.

robert-zaremba (approved):
pre-approving. Added few responses.

PR #1971, opened 2023-04-03
fix: add beta to releaser workflow

robert-zaremba:
## Description

beta releases are skipped from auto release.

### DIFF
--- a/.github/workflows/release-umee.yml
+++ b/.github/workflows/release-umee.yml
@@ -7,6 +7,7 @@ on:
     tags:
       - "v[0-9]+\\.[0-9]+\\.[0-9]+" # Official release version tags e.g. v2.0.5
       - "v[0-9]+\\.[0-9]+\\.[0-9]+-rc[0-9]+" # Release candidate tags e.g. v1.0.3-rc4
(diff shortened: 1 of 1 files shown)


.github/workflows/release-umee.yml
```diff
@@ -7,6 +7,7 @@ on:
     tags:
       - "v[0-9]+\\.[0-9]+\\.[0-9]+" # Official release version tags e.g. v2.0.5
       - "v[0-9]+\\.[0-9]+\\.[0-9]+-rc[0-9]+" # Release candidate tags e.g. v1.0.3-rc4
+      - "v[0-9]+\\.[0-9]+\\.[0-9]+-beta[0-9]+" # Release candidate tags e.g. v4.3.0-beta1
```
robert-zaremba, line 10:
```suggestion
      - "v[0-9]+\\.[0-9]+\\.[0-9]+-beta[0-9]+" # Beta releases tags e.g. v4.3.0-beta1
```

toteki (approved):
Approved. Also, would be nice if we could have these default as prereleases instead of taking up the "latest" spot in the repository

PR #1972, opened 2023-04-04
docs: UIBC and IBC Quota

robert-zaremba:
## Description

Adding module docs.


x/uibc/README.md
```diff
@@ -1 +1,89 @@
...
+
+- `Params.TokenQuota`: upper limit of a sum of all outflows per token. Initially it's set to 0.6m USD per token. It limits the outflows value for each token.
+  NOTE: we measure per token as defined in the x/leverage, not the IBC Denom Path (there can be multiple paths). Since creating a channel is permission less, we want to use same quota token.
+- `Params.TotalQuota`: upper limit of a sum of all token outflows combined. Initially it's set to 1m USD. Example of IBC outflows reaching the total quota: 300k USD worth of ATOM, 200k USD worth of STATOM, 250k USD worth of UMEE and 250k USD worth JUNO.
+
+If a quota parameter is set to zero then we consider it as unlimited.
+
+All quotas are reset in `BeginBlocker` whenever a time difference between the new block, and the previous reset is more than `Params.QuotaDuration` in seconds (initially set to 24h).
+
+Transfer is reverted whenever it breaks the any quota.
```
gsk967, line 54:
```suggestion
Transfer is reverted whenever it breaks the any quota before the end of QuotaDuration.
```

gsk967 (approved):
LGTM

x/uibc/README.md
robert-zaremba: After quota duration it will also break, for example when a single transfer is bigger then the whole quota.

x/uibc/README.md
```diff
@@ -1 +1,89 @@
...
+- ENABLED: inflow and quota outflow checks are enabled (default value).
+- TRANSFERS_PAUSED: all ICS-20 transfers are disabled.
+
+### State
+
+In the state we store:
+
+- Module [parameters](../../proto/umee/uibc/v1/quota.proto#L11).
+- Running sum of total outflow values, serialized as uint32 big-endian (no need for `sdk.Int`).
+- Running sum of per token outflow values, serialized as uint32 big-endian (no need for `sdk.Int`).
```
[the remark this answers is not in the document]
robert-zaremba: in the spec it was to use "serialized as uint32 big-endian (no need for `sdk.Int`". We were supposed to round up and use normal integers.

Now it's too late...

EbrahimUmee (approved):
removed an extra word, otherwise LGTM

PR #1978, opened 2023-04-05
tests: add e2e tests for ibc_status params

gsk967:
## Description
1. fix the gov proposal voting
2. add tests to ibc-transfer when ibc_status param is disabled to quota check
3. add gov proposal to enable ibc_status to quota check

closes: #XXXX

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


tests/grpc/client/tx/client.go
```diff
@@ -25,8 +27,8 @@ type Client struct {
 
 	ClientContext *client.Context
 
-	keyringKeyring keyring.Keyring
-	keyringRecord  *keyring.Record
+	keyringKeyring []keyring.Keyring
```
robert-zaremba, line 30: we can have many keys in a single Keyring, so we don't need a list here.

gsk967: Ok

tests/e2e/e2e_ibc_test.go
```diff
@@ -68,6 +70,39 @@ func (s *IntegrationTestSuite) TestIBCTokenTransfer() {
 		// ibc hash of uatom token
 		uatomIBCHash := "ibc/27394FB092D2ECCD56123C74F36E4C1F926001CEADA9CA97EA622B25F41E5EB2"
 
+		// Get the uibc params for quota checking
+		uibcParams, err := s.umeeClient.QueryClient.QueryUIBCParams()
+		s.Require().NoError(err)
+		if uibcParams.IbcStatus != uibc.IBCTransferStatus_IBC_TRANSFER_STATUS_QUOTA_ENABLED {
```
robert-zaremba, line 76: Why don't do something like this:
1. Start test with `s.Require.Equal(uibc.IBCTransferStatus_IBC_TRANSFER_STATUS_QUOTA_ENABLED, uibcParams.IbcStatus)`  -- this is the default, so should be set like that in the test
2. make a transfer in both (atom from gaia from umee, and uumee from umee to gaia)
3. make sure the balances got updated.
4. Make gov propo to change the IBC status to paused
5. make again transfers in both ways, and assure that neither balance got updated.

tests/e2e/e2e_setup_test.go
```diff
@@ -300,8 +301,9 @@ func (s *IntegrationTestSuite) initGenesis() {
 	var govGenState govtypesv1.GenesisState
 	s.Require().NoError(cdc.UnmarshalJSON(appGenState[govtypes.ModuleName], &govGenState))
 
-	votingPeroid := 5 * time.Second
+	votingPeroid := 15 * time.Second
```
robert-zaremba, line 304: why we are increasing the voting period?

robert-zaremba (commented):
This PR updates the `Client`.  This should go into `https://github.com/umee-network/umee/pull/1888  For some reason the latter PR has a timeout problem.  Ideally we firstly merge #1888  and then rebase this PR.
@gsk967 , could you look at #1888 and help to check what is timing out there?

tests/e2e/e2e_ibc_test.go
gsk967: Ok. I will change the tests flow

client/client.go
```diff
@@ -9,26 +9,24 @@ import (
 
 // Client sdkclient.Client and provides umee chain specific transactions and queries.
 type Client struct {
-	sdkclient.Client
+	*sdkclient.Client
```
robert-zaremba, line 12: why do we need a pointer here? The embedded struct already has pointers. I don't think we need another level of pointers.

client/client.go
```diff
@@ -9,26 +9,24 @@ import (
...
 }
 
 // NewClient constructs Client object.
 func NewClient(
 	chainID,
 	tmrpcEndpoint,
-	grpcEndpoint,
-	accountName,
```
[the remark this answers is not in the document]
robert-zaremba: maybe we should have a list of:
```
type ImportedAccount {
  name string
  mnemonic string
}
```

client/client.go
```diff
@@ -9,26 +9,24 @@ import (
...
-		accountName, accountMnemonic, gasAdjustment, encCfg)
+) (*Client, error) {
+	c, err := sdkclient.NewClient(chainID, tmrpcEndpoint, grpcEndpoint, mnemonics, gasAdjustment, encCfg)
 	if err != nil {
-		return Client{}, err
+		return nil, err
 	}
-	return Client{
-		Client: c,
+	return &Client{
```
robert-zaremba, line 28: I don't think we need to return a pointer here

client/gov.go
```diff
@@ -0,0 +1,20 @@
+package client
+
+import (
+	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types/v1"
+)
+
+func (c *Client) GovClient() govtypes.QueryClient {
```
robert-zaremba, line 7:
```suggestion
func (c Client) GovClient() govtypes.QueryClient {
```

client/client.go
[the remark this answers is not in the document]
robert-zaremba: Oh, I see that you are generating names as `val%d`. I'm OK with both solutions. If you want to stick with `val%d`, then please add a comment to the both `NewClient` constructors (umee, and sdk) that the account name is generated as `val%d`.

sdkclient/tx/client.go
```diff
@@ -110,5 +120,46 @@ func (c *Client) initTxFactory() {
 }
 
 func (c *Client) BroadcastTx(msgs ...sdk.Msg) (*sdk.TxResponse, error) {
+	c.ClientContext.From = c.keyringRecord[0].Name
+	c.ClientContext.FromName = c.keyringRecord[0].Name
+	c.ClientContext.FromAddress, _ = c.keyringRecord[0].GetAddress()
 	return BroadcastTx(*c.ClientContext, *c.txFactory, msgs...)
 }
+
+func (c *Client) BroadcastTxVotes(proposalID uint64) error {
```
robert-zaremba, line 129: Please move this function to `gov.go`

robert-zaremba, line 129:
```suggestion
// TxGovVoteYesAll creates transactions (one for each registered account) to approve a given proposal.
func (c *Client) TxGovVoteYesAll(proposalID uint64) error {
```

sdkclient/tx/client.go
```diff
@@ -110,5 +120,46 @@ func (c *Client) initTxFactory() {
...
+		c.ClientContext.FromName = c.keyringRecord[index].Name
+		c.ClientContext.FromAddress, _ = c.keyringRecord[index].GetAddress()
+
+		for retry := 0; retry < 5; retry++ {
+			// retry if txs fails, because sometimes account sequence mismatch occurs due to txs pending
+			if _, err = BroadcastTx(*c.ClientContext, *c.txFactory, []sdk.Msg{msg}...); err == nil {
+				break
+			}
+			time.Sleep(time.Second * 1)
+		}
```
robert-zaremba, line 157: let's check if err is sequence missmatch. If yes, then we continue the loop, otherwise, we break.

sdkclient/tx/gov.go
```diff
@@ -2,33 +2,19 @@ package tx
...
-		return nil, err
-	}
-
-	msg := govtypes.NewMsgVote(
-		voter,
-		proposalID,
-		voteType,
-	)
-	return c.BroadcastTx(msg)
+func (c *Client) GovVoteYes(proposalID uint64) error {
```
robert-zaremba, line 10: let's remove this function

robert-zaremba (commented):
to all transaction functions (except the one which uses all account to vote for a proposal), let's add a new parameters: `accIdx: int` -- the account index used to make transaction.

sdkclient/tx/client.go
gsk967: Sometimes we are getting other than mismatch also (ex: timeout for tx include in the block).
We will keep this loop like this, so if we got any errors we wait and retry

client/client.go
```diff
@@ -13,17 +13,16 @@ type Client struct {
 }
 
 // NewClient constructs Client object.
+// Note: For signing the transactions accounts are created by names like this val0, val1....
```
robert-zaremba, line 16:
```suggestion
// Accounts are generated using the list of mnemonics. Each string must be a sequence of words, 
// eg: `["w11 w12 w13", "w21 w22 w23"]`. Keyring names for created accounts will be: val1, val2....
```

sdkclient/client.go
```diff
@@ -15,6 +15,7 @@ import (
 // transactions and queries. The object should be extended by another struct to provide
 // chain specific transactions and queries. Example:
 // https://github.com/umee-network/umee/blob/main/client
+// For signing the transactions accounts are created by names like this val0, val1....
```
robert-zaremba, line 18:
```suggestion
// Accounts are generated using the list of mnemonics. Each string must be a sequence of words, 
// eg: `["w11 w12 w13", "w21 w22 w23"]`. Keyring names for created accounts will be: val1, val2....
```

robert-zaremba (approved):

PR #2009, opened 2023-04-25
chore(ci): use DavidAnson/markdownlint

robert-zaremba:
1. we already lint go files with golangci-linter and proto files with proto linter
2. we don't have other programming language
3. whenever PR check pass, the super linter takes anther run (about 2min) in the merge queue (most of the time is the super linter setup time).

Diff in a normal flow (PR): 2min -> down to 20s

Let's optimize the linter job by using better action.

**This PR also fixes**:

+ outstanding dead links
+ default formatting issues (markdown config was not correctly check by the super-linter)


toteki (approved):

toteki:
We can remove super linter from required statuses too

robert-zaremba:
yes, will do it, and change to `markdown-lint`

PR #2004, opened 2023-04-24
chore: update security patch to cosmwasm

gsk967:
## Description

Adding security patch to cosmwasm (https://forum.cosmos.network/t/upcoming-cosmwasm-security-patch-codename-cherry/10474)

---

### Author Checklist

_All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues._

I have...

- [ ] included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] added `!` to the type prefix if API or client breaking change
- [ ] added appropriate labels to the PR
- [ ] targeted the correct branch (see [PR Targeting](https://github.com/umee-network/umee/blob/main/CONTRIBUTING.md#pr-targeting))
- [ ] provided a link to the relevant issue or specification
- [ ] added a changelog entry to `CHANGELOG.md`
- [ ] included comments for [documenting Go code](https://blog.golang.org/godoc)
- [ ] updated the relevant documentation or specification
- [ ] reviewed "Files changed" and left comments if necessary
- [ ] confirmed all CI checks have passed

### Reviewers Checklist

_All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items._

I have...

- [ ] confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] confirmed all author checklist items have been addressed
- [ ] reviewed state machine logic
- [ ] reviewed API design and naming
- [ ] reviewed documentation is accurate
- [ ] reviewed tests and test coverage
- [ ] manually tested (if applicable)

PR #2011, opened 2023-04-26
chore: bump cosmwasm

robert-zaremba:
## Description

Replaces: #2004

* updating release notes
* this is in draft, and waiting for notional for a new release from their side (confirmed) - we don't need to do `replace` in go.mod


gsk967:
@robert-zaremba could you check once the integration When i tried  to build the binary locally it is failing

robert-zaremba:
hmm, strange that this update breaks API

gsk967 (changes requested):
Can you replace https://github.com/umee-network/umee/pull/2011/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L343


```github.com/tendermint/tendermint => github.com/cometbft/cometbft v0.34.28```

gsk967 (approved):

PR #2024, opened 2023-05-05
feat: better upgrade registration

robert-zaremba:
Added an util function to simplify upgrade registration even more.


gsk967 (approved):
LGTM

toteki (approved):

PR #2030, opened 2023-05-10
feat: simplify store generics

robert-zaremba:
## Description

We had 8 functions with wrappers to store int64, int32...
This PR reduces it to two functions

I've also removed the check prohibiting saving big `Int` or `Dec` negative values. Store should not make such assumptions. Such validation should happen layer above. On the other hand, `Coin` prohibits negative amounts and have it's `Validate` method which checks that, together with all arithmetic operations `Add`, `Sub`.


util/store/store.go
```diff
@@ -138,14 +132,9 @@ func SetInt(store sdk.KVStore, key []byte, val sdkmath.Int, errField string) err
...
 	val := GetValue[*sdk.Dec](store, key, errField)
 	if val == nil {
-		// No stored bytes at key: return zero
+		// Not found: return zero
 		return sdk.ZeroDec()
 	}
-
-	if val.IsNegative() {
-		panic(fmt.Sprintf("%s: retrieved negative %s", val, errField))
-	}
```
robert-zaremba, line 147: this check is already done in `SetDec` , so not needed.
In fact, we should do validity check in the "layer" above.

toteki: The philosophy behind checking for negative/invalid value in getters was explained in the comment at the top of the file:

Two main reasons
- Edge cases. Basically, even if we botch a migration, or load a store that was made with different Set methods (e.g. existing modules), or if some bad code doesn't use `SetDec` and instead stores a negative dec manually, this function will detect it. This is further enhanced by `errField`: if we run into this error unexpectedly, the cause will always be known in the error message (even if it's reported by a user or in a context where we don't have stack trace)
- Simplifies higher layers. If we check for negative/nil in this one file, all modules above it do not have to reproduce the same validation logic after every time they use `GetDec` It also makes it impossible to forget to include that validation.

toteki: I do see your point about not making assumptions in store - I am not sure if there is any context in our existing modules where a store negative number would be correct. It's a tradeoff between purity (no assumptions) and standardization (second point above)

robert-zaremba: Validation must happen on value operations (input, request processing, arithmetics), not on the store value, where we are out of the context.

robert-zaremba:
Note: pushing validation down to the stack, can create very bad patterns, where we are too late to handle errors properly.

kosegor (approved):

util/store/store.go
```diff
@@ -1,31 +1,21 @@
...
-//   - Incorrect binary swap (same vectors as incomplete migration)
-//
-// Setters return errors on bad input. For getters (which should only see bad data in exotic cases)
-// panics are used instead for simpler function signatures.
-//
 // Numerical getters and setters have a minimum value of zero, which will be interpreted as follows:
-//   - Setters reject negative input, and clear a KVStore entry when setting to exactly zero.
-//   - Getters panic on unmarshaling a negative value, and interpret empty data as zero.
+//   - Setters clear a KVStore entry when setting to exactly zero.
+//   - Getters panic on unmarshaling, and interpret empty data as zero.
```
toteki, line 5:
```suggestion
//   - Getters panic on unmarshal error, and interpret empty data as zero.
```

util/store/store.go
```diff
@@ -106,155 +96,101 @@ func SetObject(store sdk.KVStore, cdc codec.Codec, key []byte, object codec.Prot
...
+	switch any(v).(type) {
+	case int64, uint64:
+		v2 := binary.LittleEndian.Uint64(bz)
+		return Num(v2)
+	case int32, uint32:
+		return Num(binary.LittleEndian.Uint32(bz))
+	case byte:
+		return Num(bz[0])
+	}
+	panic("not possible!")
```
toteki, line 195: Error message could be more specific here

util/store/store.go
```diff
@@ -106,155 +96,101 @@ func SetObject(store sdk.KVStore, cdc codec.Codec, key []byte, object codec.Prot
...
+		binary.LittleEndian.PutUint64(bz, uint64(v))
+	case uint64:
+		bz = make([]byte, 8)
+		binary.LittleEndian.PutUint64(bz, v)
+	case int32:
+		bz = make([]byte, 8)
+		binary.LittleEndian.PutUint32(bz, uint32(v))
+	case uint32:
+		bz = make([]byte, 8)
+		binary.LittleEndian.PutUint32(bz, v)
```
toteki, line 173:
```suggestion
	case int32:
		bz = make([]byte, 4)
		binary.LittleEndian.PutUint32(bz, uint32(v))
	case uint32:
		bz = make([]byte, 4)
		binary.LittleEndian.PutUint32(bz, v)
```
32 bit

toteki, line 173:
```suggestion
	case int32:
		bz = make([]byte, 4)
		binary.LittleEndian.PutUint32(bz, uint32(v))
	case uint32:
		bz = make([]byte, 4)
		binary.LittleEndian.PutUint32(bz, v)
```
32 bit

util/store/store.go
robert-zaremba: updated the message

PR #2031, opened 2023-05-11
test: incentive CLI tests and url simplifications

toteki:
Adds some CLI tests.

Also, some changes:
- shortens some urls and CLI commands to reduce stutter
  - `umeed q incentive incentive-program 2` ->`umeed q incentive program 2`
  - `/umee/incentive/v1/incentive_program/2` -> `/umee/incentive/v1/program/2`
- removes pagination from complete programs
  - not necessary (it's impossible to spam programs since they have to pass gov), and makes things easier to cache
  - avoids fixing [this error](https://api.incentives-canon-3.network.umee.cc/umee/incentive/v1/incentive_programs/completed) which is probably something to do with the pagination object
 - adds missing CLI for `actual-rates` query


x/incentive/client/cli/query.go
```diff
@@ -142,6 +143,37 @@ func GetCmdQueryCurrentRates() *cobra.Command {
...
+			if len(args) > 0 {
+				denom = args[0]
+			}
+
+			queryClient := incentive.NewQueryClient(clientCtx)
+			resp, err := queryClient.ActualRates(cmd.Context(), &incentive.QueryActualRates{UToken: denom})
+			if err != nil {
+				return err
+			}
+
```
robert-zaremba, line 168:
```suggestion
```

robert-zaremba: `PrintOrErr` already handles that (as the name suggest)

x/incentive/client/tests/cli_test.go
```diff
@@ -0,0 +1,49 @@
...
+			Validator:    "umeevaloper1kqh6nt4f48vptvq4j5cgr0nfd2x4z9ulvrtqrh",
+			Orchestrator: "umee1kqh6nt4f48vptvq4j5cgr0nfd2x4z9ulv8v0ja",
+			EthAddress:   "0xddfda961410b2815b48679377baa0009ace173a2",
+		},
+	}
+
+	bz, err := cfg.Codec.MarshalJSON(&gravityGenState)
+	assert.NilError(t, err)
+
+	cfg.GenesisState[gravitytypes.ModuleName] = bz
```
robert-zaremba, line 40: I think we have a helper function for the setup. @gsk967 , do you remember?

x/incentive/client/tests/tests.go
```diff
@@ -0,0 +1,234 @@
...
+	s.RunTestQueries(zeroQueries...)
+
+	s.RunTestTransactions(
+		supplyCollateral,
+		bond,
+		beingUnbonding,
+		emergencyUnbond,
+		claim,
+		sponsor,
+	)
```
robert-zaremba, line 233: shall we run some queries after the transactions?

robert-zaremba (commented):
pre-approving.
- [ ] Let's add few queries after the transaction execution.

Also, a note: probably would be better to focus the energy on integration & unit tests to improve the coverage.

x/incentive/client/tests/tests.go
toteki: for now, we can skip that.

There are minimal tests designed to ensure the commands worked, rather than scenarios where the results of these actions are being verified (that happens in msg_server_test and scenario_test since they run faster.)

x/incentive/client/cli/query.go
```diff
@@ -142,6 +143,37 @@ func GetCmdQueryCurrentRates() *cobra.Command {
 	return cmd
 }
 
+// GetCmdQueryActualRates creates a Cobra command to query current annual rewards for a reference amount
+// of a given bonded uToken.
+func GetCmdQueryActualRates() *cobra.Command {
```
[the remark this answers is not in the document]
toteki: made an issue - seems like a good idea

x/incentive/client/tests/suite.go
```diff
@@ -0,0 +1,18 @@
...
+
+type IntegrationTests struct {
+	*itestsuite.E2ESuite
+}
+
+func NewIntegrationTestSuite(cfg network.Config, t *testing.T) *IntegrationTests {
+	return &IntegrationTests{
+		&itestsuite.E2ESuite{Cfg: cfg, T: t},
+	}
+}
```
robert-zaremba, line 18: this is not an integration test suite. It's an e2e test suite, and should be moved to `tests/e2e/incentive`

x/ugov/msg_test.go
```diff
@@ -25,7 +25,11 @@ func TestMsgGovUpdateMinGasPrice(t *testing.T) {
 	require.NoError(msg.ValidateBasic())
 
 	// TODO: probably we should use Dec string
-	require.Equal(`authority:"umee10d07y265gmmuvt4z0w9aw880jnsr700jg5w6jp" min_gas_price:<denom:"ibc/C4CFF46FD6DE35CA4CF4CE031E643C8FDC9BA4B99AE598E9B0ED98FE3A2319F9" amount:"1250000000000000000" > `, msg.String())
+	require.Equal(
+		`authority:"umee10d07y265gmmuvt4z0w9aw880jnsr700jg5w6jp"`+
+			` min_gas_price:<denom:"ibc/C4CFF46FD6DE35CA4CF4CE031E643C8FDC9BA4B99AE598E9B0ED98FE3A2319F9"`+
+			` amount:"1250000000000000000" > `,
```
robert-zaremba, line 31: why changing this?

robert-zaremba (commented):
overall looks good. One nit: the CLI tests are e2e tests, so let's move them under e2e.

x/ugov/msg_test.go
toteki: line length linter was giving me a warning I think

x/incentive/client/tests/suite.go
toteki: Should we also move the matching `x/leverage/client/tests/suite.go` to e2e? It looks like the makefile won't need to be changed, so this seems doable.

x/ugov/msg_test.go
toteki: (outdated after merging main)

robert-zaremba:
Why this PR was merged?
1. followups were not added to linear
2. e2e tests are mixed in the module tests

x/incentive/client/tests/cli_test.go
robert-zaremba: The problem is that often this followups are not done, and we cumulate this copy pastes which are less and less readable.

PR #2035, opened 2023-05-12
chore: use cli.PrintOrErr

robert-zaremba:
## Description

small cleanups in the client code


gsk967 (approved):
Pre Approving 

Can you check the comment

PR #2037, opened 2023-05-12
chore: speedup e2e by removing docker cache

robert-zaremba:
## Description

cache is not reused in other actions, and adding it increases the setup time.


toteki (approved):

toteki:
seems ok but we can always revert if this break something unexpected

robert-zaremba:
the e2e test is running almost 2min faster now.

PR #2040, opened 2023-05-16
fix: incentive module EndBlocker

toteki:


x/incentive/keeper/unit_test.go
```diff
@@ -137,13 +138,20 @@ func (k *testKeeper) advanceTime(duration int64) {
...
-	require.Nil(k.t, k.updatePrograms(k.ctx), "update programs")
+	// update block time in testkeeper's context, so endBlock will read it
+	utcTime := time.Unix(blockTime, 0)
+	k.ctx = k.ctx.WithBlockTime(utcTime)
+
+	// ensure that last rewards time has actually updated without errors
+	err, skipped := k.EndBlock(k.ctx)
+	require.Nil(k.t, err, "endBlock")
+	require.False(k.t, skipped, "endBlock skipped")
 	require.Equal(k.t, blockTime, k.GetLastRewardsTime(k.ctx))
```
toteki, line 155: This is why the test was not catching the problem - it called the two `update` functions but the bug was in the logic around them in `EndBlock`

x/incentive/keeper/update.go
```diff
@@ -156,27 +156,32 @@ func (k Keeper) updatePrograms(ctx sdk.Context) error {
...
 		k.Logger(ctx).Error(
 			"incentive module LastRewardTime was not initialized",
 			"prev", prevTime,
+			"blockTime", blockTime,
 		)
-		return k.setLastRewardsTime(ctx, blockTime)
+		if err := k.setLastRewardsTime(ctx, blockTime); err != nil {
+			return err, true
+		}
+		prevTime = blockTime
```
toteki, line 184: This change is unrelated to the bug, but should be marginally better

x/incentive/keeper/update.go
```diff
@@ -190,16 +195,16 @@ func (k Keeper) EndBlock(ctx sdk.Context) error {
...
 	}
 
 	// Implications of updating reward accumulators and PrevRewardTime before starting/completing incentive programs:
 	// - if an incentive program starts this block, it does not disburse any rewards this block
 	// - if an incentive program completes this block, it distributes its remaining rewards before being completed
 	// - in the case of a chain halt which misses a program's start time, rewards before its late start are skipped
 	// - in the case of a chain halt which misses a program's end time, remaining rewards are correctly distributed
-	if err := k.updateRewards(ctx, prevTime); err != nil {
-		return err
+	if err := k.updateRewards(ctx, blockTime); err != nil {
```
toteki, line 206: This was the bug

x/incentive/keeper/update.go
```diff
@@ -156,27 +156,32 @@ func (k Keeper) updatePrograms(ctx sdk.Context) error {
 
 // EndBlock updates incentive programs and reward accumulators, then sets LastRewardTime
 // to the current block time. Also protects against negative time elapsed (without causing chain halt).
-func (k Keeper) EndBlock(ctx sdk.Context) error {
+// In addition to regular error, returns a boolean indicating whether the main logic was skipped due
+// to a blockTime issue. These situations are accompanied by error logs.
+func (k Keeper) EndBlock(ctx sdk.Context) (err error, skipped bool) {
```
robert-zaremba, line 161:
```suggestion
func (k Keeper) EndBlock(ctx sdk.Context) (skipped bool, err error, ) {
```

x/incentive/keeper/update.go
```diff
@@ -156,27 +156,32 @@ func (k Keeper) updatePrograms(ctx sdk.Context) error {
...
 	blockTime := ctx.BlockTime().Unix()
 	if blockTime < 0 {
 		k.Logger(ctx).Error(
 			incentive.ErrDecreaseLastRewardTime.Error(),
 			"negative block time", blockTime,
 		)
-		return nil
+		return nil, true
```
robert-zaremba, line 168: I think this should be an error

x/incentive/module/abci.go
```diff
@@ -10,6 +10,7 @@
 
 // EndBlocker implements EndBlock for the x/incentive module.
 func EndBlocker(ctx sdk.Context, k keeper.Keeper) []abci.ValidatorUpdate {
-	util.Panic(k.EndBlock(ctx))
+	err, _ := k.EndBlock(ctx)
+	util.Panic(err)
```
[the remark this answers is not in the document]
robert-zaremba: can we inspect the error to panic only when really necessary?

x/incentive/keeper/scenario_test.go
```diff
@@ -171,6 +179,20 @@ func TestBasicIncentivePrograms(t *testing.T) {
 		rewards,
 		"alice pending rewards at time 300",
 	)
+	// actually claim the rewards (same amount)
+	rewards, err = k.UpdateAccount(k.ctx, alice)
+	require.NoError(k.t, err)
+	require.Equal(
+		k.t,
+		sdk.NewCoins(sdk.NewInt64Coin(umee, 100000+7_920000+8_000000)),
```
robert-zaremba, line 187: 1. this numbers are repeated, let's save it in a variable
2. could you calculate this numbers, instead of hardcoding the sum of values?

robert-zaremba (commented):
The tests has lot of ad hock values, would be better to actually calculate them in the test.

x/incentive/keeper/update.go
toteki: This logs an error, but does not return one. Returned errors become panic (chain halt)

x/incentive/keeper/update.go
toteki: fixed

x/incentive/keeper/scenario_test.go
toteki: Saved to a var. I do want to keep the sum hardcoded, since this test is for a specific scenario and it should fail if we make any unintentional changes

EbrahimUmee (approved):

x/incentive/keeper/update.go
robert-zaremba: well, I would say that this is an impossible scenario, to have ctx.BlockTime before 1971

PR #2045, opened 2023-05-16
feat: add ugov/MsgGovUpdateMinGasPrice String implementation

robert-zaremba:
## Description

add proper `String` implementation MsgGovUpdateMinGasPrice


x/ugov/msg.go
```diff
@@ -23,3 +25,8 @@ func (msg *MsgGovUpdateMinGasPrice) ValidateBasic() error {
 func (msg *MsgGovUpdateMinGasPrice) GetSigners() []sdk.AccAddress {
 	return checkers.Signers(msg.Authority)
 }
+
+// GetSignBytes implements Stringer method
+func (msg *MsgGovUpdateMinGasPrice) String() string {
```
toteki, line 30:
```suggestion
// String implements Stringer interface
func (msg *MsgGovUpdateMinGasPrice) String() string {
```

x/ugov/msg.go
```diff
@@ -23,3 +25,8 @@ func (msg *MsgGovUpdateMinGasPrice) ValidateBasic() error {
...
+	return fmt.Sprintf("<authority: %s, min_gas_price: %s>", msg.Authority, msg.MinGasPrice.String())
```
toteki, line 31: Is the `<` `>` necessary? I notice before it was only appearing in the dec

toteki (approved):

x/ugov/msg.go
robert-zaremba: No, it's just a convention for composed objects.

robert-zaremba: In the future we will have better, automated generator, which doesn't require yaml etc...

PR #2058, opened 2023-05-19
feat: move cosmwasm into stable and x/incentive into experiemntal

robert-zaremba:
We are finalizing the cosmwams release, so need to swap the experiemntal features.


toteki (approved):

toteki:
make sure to test the upgrade again before canon-2, to make sure we're not missing a line

app/app.go
```diff
@@ -628,7 +631,34 @@ func New(
...
+	var err error
+	wasmDir := filepath.Join(homePath, "wasm")
+	app.wasmCfg, err = wasm.ReadWasmConfig(appOpts)
+	if err != nil {
+		panic(fmt.Sprintf("error while reading wasm config: %s", err))
+	}
+
+	// The last arguments can contain custom message handlers, and custom query handlers,
+	// if we want to allow any custom callbacks
+	availableCapabilities := "iterator,staking,stargate,cosmwasm_1_1"
```
gsk967, line 643:
```suggestion
	availableCapabilities := "iterator,staking,stargate,cosmwasm_1_1,umee"
```
can you add `umee` also

app/app.go
```diff
@@ -477,7 +476,11 @@ func New(
 		app.LeverageKeeper,
 	)
 	app.LeverageKeeper.SetTokenHooks(app.OracleKeeper.Hooks())
-	app.LeverageKeeper.SetBondHooks(app.IncentiveKeeper.BondHooks())
+	if Experimental {
```
gsk967, line 479: Can move `app.IncentiveKeeper` inside experimental

app/experimental_appconfig.go
```diff
@@ -3,93 +3,8 @@
...
-	chunks := strings.Split(EnableSpecificWasmProposals, ",")
-	proposals, err := wasm.ConvertToProposals(chunks)
-	if err != nil {
-		panic(err)
-	}
-
-	return proposals
-}
-
-func (app *UmeeApp) customKeepers(
```
gsk967, line 61: Can you keep customKeepers as it is, so we can use for leverageKeeper or any future experimental module keepers

gsk967 (approved):
LGTM, Pre Approving 
Can you check the comments

app/experimental_appconfig.go
robert-zaremba: I don't think this is needed, we can do if in the app.go, it will be more clear.

app/app.go
robert-zaremba: we can initialize the keeper. It doesn't hurt. There is no code running when you initialize the keeper.
It's important to not initialize the module, and not register the hooks nor the begin/end/init blockers

gsk967: Okay

PR #2062, opened 2023-05-22
feat: ugov wiring & cli

robert-zaremba:


robert-zaremba:
grpc-web gateway works: http://127.0.0.1:1317/umee/ugov/v1/min-gas-price
But CLI doesn't:
```
Error: rpc error: code = Unknown desc = unknown query path: unknown request
Usage:
  umeed query ugov min-gas-price [flags]
```

app/app.go
```diff
@@ -482,6 +483,8 @@ func New(
 
 	app.UGovKeeperB = ugovkeeper.NewKeeperBuilder(appCodec, keys[ugov.ModuleName])
 
+	app.UGovKeeperB = ugovkeeper.NewKeeperBuilder(appCodec, keys[ugov.ModuleName])
```
gsk967, line 486: duplicate code , it is already exists in above line

gsk967 (approved):

gsk967:
@robert-zaremba  there is cli queries are missing

PR #2071, opened 2023-05-30
bump GB and set GB burn

robert-zaremba:
## Description

+ bumps GB version
+ set Burn mechanism to true

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -502,6 +502,7 @@ func New(
 		&app.IBCTransferKeeper,
 		&app.bech32IbcKeeper,
 	)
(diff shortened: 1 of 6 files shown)


app/app.go
```diff
@@ -502,6 +502,7 @@ func New(
 		&app.IBCTransferKeeper,
 		&app.bech32IbcKeeper,
 	)
+	app.GravityKeeper.BurnValset(true)
```
facundomedica, line 505: This is not necessary, it's already set to true by default when creating the keeper

robert-zaremba: let's keep it for clarity

PR #2085, opened 2023-06-06
feat: add inspector queries

toteki:
Contains a query that combines the risk data query as specified by risk team, and the inspector query which allows a user to filter all active borrower accounts.

`umeed query leverage inspect [symbol] [borrowed] [collateral [danger] [ltv]`

Example of web usage:
```
http://localhost:1317/umee/leverage/v1/inspect?symbol=ALL&borrowed=100000&ltv=0.7
```

Example of new CLI usage:

(show all borroweds with `>$100k USD borrowed value` and `>0.7 LTV`

```
umeed q leverage inspect all 100000 0 0 0.7

borrowers:
- address: umee1rphd0j66lpnzchpw956ejyhg237rffy9k9zvph
  analysis:
    Borrowed: 1007000
    Liquidation: 1065000
    Value: 1365000
  position:
    borrowed:
    - amount: "1008946.943914000000000000"
      denom: USDC
    collateral:
    - amount: "3032261.017830785773337404"
      denom: OSMO
- address: umee1pjuufs7jk6sa5avs36uec8u8spf2hqq02xdyyy
  analysis:
    Borrowed: 375791
    Liquidation: 388241
    Value: 507004
  position:
    borrowed:
    - amount: "376496.493107000000000000"
      denom: USDC
    collateral:
    - amount: "23432.816342540682021296"
      denom: ATOM
    - amount: "0.000000000100017791"
      denom: DOT
    - amount: "192246.705675326605412238"
      denom: OSMO
    - amount: "5185.135472357015580737"
      denom: STOSMO
    - amount: "0.000000010010631796"
      denom: WBTC
    - amount: "22473.172598202930492651"
      denom: stATOM
```


proto/umee/leverage/v1/query.proto
```diff
@@ -65,6 +67,26 @@ service Query {
...
+  rpc Inspect(QueryInspect)
+      returns (QueryInspectResponse) {
+    option (google.api.http).get = "/umee/leverage/v1/inspect";
+  }
+
+  // InspectNeat is the inspector query, with a different response structure
+  // optimized for human readability on the CLI.
+  rpc InspectNeat(QueryInspectNeat)
+      returns (QueryInspectNeatResponse) {
+    option (google.api.http).get = "/umee/leverage/v1/inspect_neat";
```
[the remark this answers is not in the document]
toteki: When I was building inspector query and using it during the SVB collapse, having this must more concise return helped tremendously when sorting through a large number of accounts visually

proto/umee/leverage/v1/inspector.proto
```diff
@@ -0,0 +1,152 @@
...
+  // X is borrowed value as a portion of borrow limit.
+  double X = 3;
+  // Y is borrowed value as a portion of liquidation threshold.
+  double Y = 4;
+  // Z is the borrowed value as a portion of collateral value.
+  double Z = 5;
+}
+
+// AccountSymbolBalances contains an account's position denoted in symbol denom tokens.
+message AccountSymbolBalances {
```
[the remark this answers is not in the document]
toteki: changed to `AccountDecBalances`

proto/umee/leverage/v1/inspector.proto
```diff
@@ -0,0 +1,140 @@
...
+message BorrowerSummaryNeat {
+  string Account = 1;
+  // Borrowed is account's borrowed value, either in total or just a specified denom.
+  double Borrowed = 2;
+  // LTB is borrowed value as a portion of borrow limit.
+  double LTB = 3;
+  // LTL is borrowed value as a portion of liquidation threshold.
+  double LTL = 4;
+  // LTV is the borrowed value as a portion of collateral value.
+  double LTV = 5;
```
robert-zaremba, line 119: what `LT` stands for in the queries above?

x/leverage/keeper/inspector.go
```diff
@@ -0,0 +1,363 @@
...
+			filters = append(filters, withMinDanger(sortMin))
+		}
+	case "ltv":
+		sorting = moreLTV()
+		if sortMin.IsPositive() {
+			filters = append(filters, withMinLTV(sortMin))
+		}
+	default:
+		// if no sort mode is specified, return all borrowers sorted by total collateral value
+		sorting = moreCollateral(false)
```
robert-zaremba, line 80: why we don't add `sortMin` filter here?
I think the `if sortMin.isPositive()` .... should be outside of the switch, then being repeated.

x/leverage/keeper/inspector.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+	k, ctx := q.Keeper, sdk.UnwrapSDKContext(goCtx)
+	borrowers := k.riskData(ctx)
+
+	return &types.QueryRiskDataResponse{Borrowers: borrowers}, nil
+}
+
+// neat truncates an sdk.Dec to a common-sense precision based on its size and converts it to float.
+// This makes a big difference in readability when using the CLI.
+func neat(num sdk.Dec) float64 {
```
robert-zaremba, line 160: could you add few tests to this function?

x/leverage/keeper/inspector.go
```diff
@@ -0,0 +1,363 @@
...
+	// Truncate the float at a certain precision (can be negative)
+	x := math.Pow(10, float64(precision))
+	return float64(int(n*x)) / x
+}
+
+// filteredSortedBorrowers returns a list of borrower addresses and their account summaries sorted and filtered
+// by selected methods. Sorting is ascending in X if the sort function provided is "less X", and descending in
+// X if the function provided is "more X"
+func (k Keeper) filteredSortedBorrowers(
+	ctx sdk.Context, filters []inspectorFilter, sorting inspectorSort, symbol string,
```
robert-zaremba, line 187: let's document the `symbol`.

x/leverage/keeper/inspector.go
```diff
@@ -0,0 +1,363 @@
...
+	borrowers := k.unsortedBorrowers(ctx, symbol)
+
+	// filters the borrowers
+	filteredBorrowers := []*types.BorrowerSummary{}
+	for _, bs := range borrowers {
+		ok := true
+		for _, f := range filters {
+			if !f(bs) {
+				ok = false
+			}
```
robert-zaremba, line 198:
```suggestion
			ok = ok && f(bs)
```

x/leverage/keeper/inspector.go
```diff
@@ -0,0 +1,363 @@
...
+	symbols map[string]string,
+	exponents map[string]uint32,
+	exchangeRates map[string]sdk.Dec,
+) sdk.DecCoins {
+	symbolCoins := sdk.NewDecCoins()
+
+	for _, c := range coins {
+		if _, ok := symbols[c.Denom]; !ok {
+			// unregistered tokens cannot be converted, but can be returned as base denom
+			symbolCoins = symbolCoins.Add(sdk.NewDecCoinFromDec(c.Denom, sdk.NewDecFromInt(c.Amount)))
```
robert-zaremba, line 347: why we don't apply exchange rate here?

x/leverage/keeper/inspector_sort.go
```diff
@@ -0,0 +1,58 @@
...
+// which must return true if a should come before b using custom logic for sorts.
+type inspectorSort func(a, b *types.BorrowerSummary) bool
+
+func moreLTV() inspectorSort {
+	return func(a, b *types.BorrowerSummary) bool {
+		ha := a.BorrowedValue.Quo(a.CollateralValue)
+		hb := b.BorrowedValue.Quo(b.CollateralValue)
+		return ha.GT(hb)
+	}
+}
```
robert-zaremba, line 28:
```suggestion
func moreLTV(a, b *types.BorrowerSummary) bool {
		ha := a.BorrowedValue.Quo(a.CollateralValue)
		hb := b.BorrowedValue.Quo(b.CollateralValue)
		return ha.GT(hb)
}
```

robert-zaremba: and then you pass it as `moreLTV` instead `moreLTV()`.
Same below.

robert-zaremba (commented):
few more comments. Would be good to add a bit more tests.

proto/umee/leverage/v1/inspector.proto
toteki: "loan to" - for example, LTV is "loan to value" and is a term that appears in the literature. 

The other two are not based on any official terms - they stand for "loan to borrow limit" and loan to liquidation threshold".

Not 100% informative as they are now, but if I give them verbose names like `LoanToValue` then the output of the query is crowded by the same words repeated over and over each entry, and the numbers are no longer as prominent. This makes it harder to read

x/leverage/keeper/inspector.go
toteki: an unregistered denom will not have a uToken or a valid utoken exchange rate

(in practice, unregistered tokens shouldn't appear here at all, but if they do I want them to be displayed as-is rather than causing an error)

x/leverage/keeper/inspector_sort.go
toteki: fixed

x/leverage/keeper/inspector.go
toteki: `if sortMin.isPositive()` does a different thing depending on which case it's in, so it can't be moved outside

In the default (no sorting specified) case, we currently ignore the minimum value to filter from sorted collateral.

proto/umee/leverage/v1/inspector.proto
```diff
@@ -0,0 +1,152 @@
...
+
+import "gogoproto/gogo.proto";
+import "cosmos/base/v1beta1/coin.proto";
+
+option go_package = "github.com/umee-network/umee/v5/x/leverage/types";
+
+option (gogoproto.goproto_getters_all) = false;
+
+// QueryInspect defines the request structure for the Inspect gRPC service handler.
+message QueryInspect {
```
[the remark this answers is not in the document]
toteki: finished combining proto files and some go files

proto/umee/leverage/v1/inspector.proto
```diff
@@ -0,0 +1,152 @@
...
+  // Mode is a string which selects the logic behind the inspector query.
```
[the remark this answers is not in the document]
toteki: replaced `mode, modemin, sort, sortmin` with parameters `borrow, collateral, danger, LTV` which all act as minimum values to filter positions. Each one is documented and there is no longer ambiguity in which modes might be available.

proto/umee/leverage/v1/inspector.proto
toteki: removed unclear field names in response struct refactor

x/leverage/keeper/inspector.go
toteki: simplified

x/leverage/keeper/inspector.go
toteki: added

x/leverage/keeper/inspector.go
toteki: now documented in proto (and this sub function with `symbol` param no longer exists)

proto/umee/leverage/v1/query.proto
```diff
@@ -320,3 +326,56 @@ message QueryMaxBorrowResponse {
     (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"
   ];
 }
+
+// QueryInspect defines the request structure for the Inspect gRPC service handler.
+message QueryInspect {
+  // Symbol selects a symbol denom to sort accounts by borrowed value. Use "all" to ignore.
```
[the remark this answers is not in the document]
toteki: I'll clarify that either empty string or all will ignore. (for CLI, empty string for an arg in the middle of the list is not possible, so the "all" option is required)

x/leverage/keeper/inspector.go
```diff
@@ -0,0 +1,191 @@
...
+	// collect all accounts (unsorted)
+	_ = k.iterate(ctx, prefix, iterator)
+
+	// filters the borrowers
+	filteredBorrowers := []*types.InspectAccount{}
+	for _, account := range borrowers {
+		ok := account.Analysis.Borrowed > req.Borrowed
+		ok = ok && account.Analysis.Value > req.Collateral
+		ok = ok && account.Analysis.Borrowed/account.Analysis.Liquidation > req.Danger
+		ok = ok && account.Analysis.Borrowed/account.Analysis.Value > req.Ltv
```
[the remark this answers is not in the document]
toteki: fixed

x/leverage/keeper/inspector.go
```diff
@@ -0,0 +1,191 @@
...
+		if ok {
+			filteredBorrowers = append(filteredBorrowers, account)
+		}
+	}
+
+	// sorts the borrowers
+	sort.Sort(byCustom{
+		bs:   filteredBorrowers,
+		less: moreBorrowed(req.Symbol),
+	})
```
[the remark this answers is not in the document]
toteki: simplified

proto/umee/leverage/v1/query.proto
```diff
@@ -65,6 +65,12 @@ service Query {
       returns (QueryMaxBorrowResponse) {
     option (google.api.http).get = "/umee/leverage/v1/max_borrow";
   }
+
+  // Inspect is the customizable inspector query.
```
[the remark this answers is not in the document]
toteki: added

PR #2122, opened 2023-06-29
v5.1-beta docs

robert-zaremba:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,8 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


gsk967 (approved):

app/upgrades.go
```diff
@@ -57,17 +57,16 @@ func (app UmeeApp) RegisterUpgradeHandlers(bool) {
 }
 
 func (app *UmeeApp) registerUpgrade5_1(upgradeInfo upgradetypes.Plan) {
-	planName := "v5.1"
+	planName := "v5.1-beta" // TODO: decide about the final release
```
toteki, line 60: Should we call this "v5.1" regardless (or "v5.1.0" if we need it to be different) and if canon-3 needs two releases, only give the second one a special name?

toteki (approved):
pre-approve, some discussion on name remains

PR #2140, opened 2023-07-11
fix: rename ugov EventMinTxFees to EventMinGasPrice

robert-zaremba:
## Description

The event name was incorrect, we are changing the price, not the min fee.


gsk967 (approved):
Approving , can you resolve the merge conflicts

PR #2147, opened 2023-07-14
refactore SetObject and GetObject usage

robert-zaremba:
## Description

There is a confusion weather to use `store.GetValue` or `store.GetObject`.

We can't really deprecate `GetObject`, because there are few useful cases, where we want to reuse the object and pass it as a pointer.
Moreover if an object contains a `codec.Any` field, then it can't be used in `store.GetValue`.

Idea:
* update docs,
* simplify usage
* rename `store.GetObject` to `store.GetValueCdc` , and similarly `store.SetObject`
* add helper function for GetObject and SetObject in the x/incentive, which currently doesn't use the `KeeperBuilder` pattern.


x/incentive/keeper/store.go
```diff
@@ -9,22 +9,20 @@ import (
...
 }
 
 // setParams validates and sets the incentive module parameters
 func (k Keeper) setParams(ctx sdk.Context, params incentive.Params) error {
-	kvs := k.KVStore(ctx)
 	if err := params.Validate(); err != nil {
 		return err
 	}
-	return store.SetObject(kvs, k.cdc, keyPrefixParams, &params, "params")
+	return store.SetValue(k.KVStore(ctx), keyPrefixParams, &params, "params")
```
toteki, line 25: Incentive module is not currently storing these objects in state on mainnet (they're default since they haven't been set). However, they have been set on canon-3 (and could possibly be set on mainnet before we ship this).

So, do we need to worry about changing which get/set encoding is used here?

x/incentive/keeper/store.go
```diff
@@ -116,9 +112,8 @@ func (k Keeper) setIncentiveProgram(
 		return sdkerrors.ErrInvalidRequest.Wrapf("program %d already exists with status %d", program.ID, s)
 	}
 
-	kvStore := k.KVStore(ctx)
 	key := keyIncentiveProgram(program.ID, status)
-	return store.SetObject(kvStore, k.cdc, key, &program, "incentive program")
+	return k.setObject(ctx, key, &program, "incentive program")
```
toteki, line 116: same concern - we have some incentive programs in state already, so does this require a migration?

util/store/store.go
```diff
@@ -20,7 +20,9 @@ import (
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 )
 
-// GetValue loads value from the store using default Unmarshaler
+// GetValue loads value from the store using default Unmarshaler. Panics on failure to decode.
+// Returns nil if the key is not found in the store.
+// If the value contains codec.Any filed, then SetObject MUST be used instead.
```
toteki, line 25:
```suggestion
// If the value contains codec.Any field, then SetObject MUST be used instead.
```

x/incentive/keeper/store.go
```diff
@@ -86,10 +83,7 @@ func (k Keeper) getIncentiveProgram(ctx sdk.Context, id uint32) (
 		return program, incentive.ProgramStatusNotExist, sdkerrors.ErrNotFound.Wrapf("program id %d", id)
 	}
 
-	kvStore := k.KVStore(ctx)
-
-	// Looks for an incentive program with the specified ID and status
-	if !store.GetObject(kvStore, k.cdc, keyIncentiveProgram(id, status), &program, "incentive program") {
+	if k.getObject(&ctx, keyIncentiveProgram(id, status), &program, "incentive program") {
```
toteki, line 86:
```suggestion
	if !k.getObject(&ctx, keyIncentiveProgram(id, status), &program, "incentive program") {
```

x/oracle/keeper/historic_avg.go
```diff
@@ -143,10 +139,22 @@ func (k AvgKeeper) GetCurrentAvg(denom string) (sdk.Dec, error) {
...
+	av := store.GetValue[*types.AvgCounter](k.store, key, "avg counter")
+	if av == nil {
+		return types.AvgCounter{}, sdkerrors.ErrNotFound.Wrap("avg counter")
 	}
-	var av types.AvgCounter
-	return av, av.Unmarshal(bz)
+	return *av, nil
+}
+
+// SetAvgPeSetHistoricAvgCounterParams sets avg period and avg shift time duration
```
toteki, line 149:
```suggestion
// SetHistoricAvgCounterParams sets avg period and avg shift time duration
```

kosegor (approved):
LGTM after fixing Adam's comments

PR #2154, opened 2023-07-18
improve(oracle): don't use pointer type for Price slice alias

robert-zaremba:
## Description

`type Prices = *[]Price` being a type alias for pointer is smells bad. The slice usually doesn't have to be a pointer, because it already contains pointer. Also, based on usage, this often requires dereference to iterate (for range ...) or to access elements (index). 
So it's better to simplify and remove the pointer.

### DIFF
--- a/x/oracle/keeper/grpc_query.go
+++ b/x/oracle/keeper/grpc_query.go
@@ -301,7 +301,7 @@ func (q querier) MedianDeviations(
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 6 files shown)


gsk967 (approved):
LGTM, Pre Approving Small Suggeation  
I think we can remove pointer return for `NewPrice` func also , because lot of places they are using pointer to access the value of this 
```
func NewPrice(exchangeRate sdk.Dec, denom string, blockNum uint64) *Price {
}
```

toteki (approved):

PR #2155, opened 2023-07-19
feat: basis points math package

robert-zaremba:
## Description

New package to optimize arithmetic computation when we need to deal with fraction number of basis points.
There is no precision lost for operations like `a * b` where `a` or `b` is basis points, and the other variable is `Int`.

### DIFF
--- a/util/bpmath/bp.go
+++ b/util/bpmath/bp.go
@@ -4,11 +4,21 @@ import (
 	"math"
 
 	cmath "cosmossdk.io/math"
+	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
 // BP represents values in basis points. Maximum value is 2^32-1.
 // Note: BP operations should not be chained - this causes precision losses.
 type BP uint32
+
+func (bp BP) ToDec() sdk.Dec {
+	return sdk.NewDecWithPrec(int64(bp), 4)
+}
+
+// Mul return a*bp rounding towards zero.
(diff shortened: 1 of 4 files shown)


gsk967 (approved):

util/bpmath/bp.go
```diff
@@ -0,0 +1,49 @@
+package bpmath
+
+import (
+	"math"
+
+	cmath "cosmossdk.io/math"
+)
+
+// BP represents values in basis points. Maximum value is 2^32-1.
+// Note: BP operations should not be chained - this causes precision loses.
```
toteki, line 10:
```suggestion
// Note: BP operations should not be chained - this causes precision losses.
```

util/bpmath/types.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+type Rounding uint
+
+const (
+	DOWN = iota
+	UP
+)
+
+const ONE = 10000
```
toteki, line 14:
```suggestion
const One = 10000
```
Probably don't need all caps

util/bpmath/types.go
```diff
@@ -0,0 +1,19 @@
+package bpmath
+
+import (
+	"cosmossdk.io/math"
+)
+
+type Rounding uint
+
+const (
+	DOWN = iota
```
toteki, line 10:
```suggestion
	DOWN Rounding = iota
```
keeps the constants defined under a type

kosegor (approved):

PR #2164, opened 2023-07-21
feat: implement special asset pairs

toteki:
Extension of #2150 (second try, since the first got a little messy with merge commits from main)


toteki:
Implementation of AssertBorrowerHealth is R4R here.

Still need to create MaxBorrow, MaxWithdraw (with helpers on the `types.AccountPosition` struct)

x/leverage/README.md
```diff
@@ -200,13 +200,14 @@ For example, an account using a single collateral token with `CollateralWeight 0
 
 The leverage module can define pairs of assets which are advantaged when one is used as collateral to borrow the other.
 
-They are defined in the form `[Asset A, Asset B, Special Collateral Weight]`. In effect, this means that
+They are defined in the form `[Asset A, Asset B, Special Collateral Weight, Special Liquidation Threshold]`. In effect, this means that
```
toteki, line 203: Necessary change, it turns out, to prevent special asset pairs allowing users to instantly become eligible for liquidation.

x/leverage/keeper/iter.go
```diff
@@ -45,14 +47,22 @@ func (k Keeper) GetAllRegisteredTokens(ctx sdk.Context) []types.Token {
...
+	sort.SliceStable(pairs, func(i, j int) bool {
+		return pairs[i].CollateralWeight.LT(pairs[j].CollateralWeight)
+	})
+	return pairs
 }
 
 // GetSpecialAssetPairs returns all the special asset pairs from the x/leverage
 // module's KVStore which match a single asset.
 func (k Keeper) GetSpecialAssetPairs(ctx sdk.Context, denom string) []types.SpecialAssetPair {
 	prefix := types.KeySpecialAssetPairOneDenom(denom)
```
toteki, line 60: This is why we needed directional store - it is now possible to load only relevant special asset pairs instead of all of them (as sets don't get the denom prefix)

x/leverage/types/position.go
```diff
@@ -0,0 +1,230 @@
...
+	position := AccountPosition{
+		specialPairs:    WeightedSpecialPairs{},
+		collateral:      WeightedDecCoins{},
+		borrowed:        WeightedDecCoins{},
+		tokens:          map[string]Token{},
+		collateralValue: sdk.ZeroDec(),
+		borrowedValue:   sdk.ZeroDec(),
+	}
+
+	// cache all registered tokens
```
toteki, line 55: This doesn't go beyond the query or transaction that is using `AccountPosition`, so it's not the kind of cache that causes gas nondeterminism.

We would get only tokens represented in the existing position, but this struct must be capable of computing max borrow for assets that aren't initially present as well, so we fetch all.

x/leverage/types/position.go
```diff
@@ -0,0 +1,230 @@
...
+		)
+	}
+
+	return position
+}
+
+// BorrowedValue returns an account's total USD value borrowed
+func (ap *AccountPosition) BorrowedValue() sdk.Dec {
+	return ap.borrowedValue
+}
```
toteki, line 148: Keeping all the fields on this struct private to prevent accidental misuse: this value should be immutable once the struct is initialized, so we don't expose it to be set from the keeper package (or others)

x/leverage/types/position.go
```diff
@@ -0,0 +1,230 @@
...
+		return ap.borrowedValue.Sub(remainingBorrowValue)
+	}
+
+	// if no borrows remain after matching, user may have additional borrow limit available
+	borrowLimit := ap.borrowedValue
+	for _, c := range remainingCollateral {
+		// the borrow limit calculation assumes no additional limitations based on borrow factor
+		borrowLimit = borrowLimit.Add(c.Asset.Amount.Mul(c.Weight))
+	}
+	return borrowLimit
```
toteki, line 225: If the borrowed asset had a lower collateral weight than remaining collateral, the actual max borrow will be lower than is suggested here. We will have a separate helper function for MaxBorrow, or add parameters to this one to enable that calculation.

toteki:
Converting back to draft - the AccountPosition struct is being moved to #2178 and will be tested extensively before I resume my work here, since I've been running into some unexpected complexity.

toteki:
Superseded by #2178

PR #2169, opened 2023-07-26
Update numeric store getters to return bool if value is missing

robert-zaremba:
In Numeric getters (Big Int, Dec, int, int64...) returned zero value if the value was not present in the store, without letting know if the value is there. That can cause unexpected behavior when zero is a proper value, but we want to know if it was present in the store or not.

This PR adds `bool` return type to numeric getters and the time getter.

### DIFF
--- a/proto/umee/metoken/v1/query.proto
+++ b/proto/umee/metoken/v1/query.proto
@@ -111,5 +111,5 @@ message QueryIndexPrice {
 
 // QueryIndexPriceResponse defines the response structure for the IndexPrice gRPC service handler.
 message QueryIndexPriceResponse {
-  repeated Price prices = 1 [(gogoproto.nullable)   = false];
+  repeated Price prices = 1 [(gogoproto.nullable) = false];
 }
--- a/swagger/proto-config-gen.json
+++ b/swagger/proto-config-gen.json
@@ -45,6 +45,14 @@
           "Params": "IncentiveParams"
         }
       }
+    },
+    {
+      "url": "./tmp-swagger-gen/umee/metoken/v1/query.swagger.json",
+      "operationIds": {
+        "rename": {
+          "Params": "MeTokenParams"
+        }
+      }
     }
   ]
 }
--- a/swagger/swagger.yaml
+++ b/swagger/swagger.yaml
@@ -2288,6 +2288,102 @@ paths:
                       format: byte
       tags:
         - Query
+  /umee/ugov/v1/inflation_cycle_end:
+    get:
+      summary: InflationCycleEnd returns inflation cycle end time
+      operationId: InflationCycleEnd
+      responses:
+        '200':
+          description: A successful response.
+          schema:
+            type: object
+            properties:
+              end:
+                type: string
+                format: date-time
+            description: QueryInflationCycleEndResponse response type.
+        default:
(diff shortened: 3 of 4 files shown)


util/store/store.go
```diff
@@ -163,12 +162,9 @@ func SetAddress(store sdk.KVStore, key []byte, val sdk.AccAddress) {
 
 // GetTimeMs retrieves time saved as Unix time in Miliseconds.
 // Returns sdkerrors.NotFound error if the value is not there, hence time = 0 is not supported.
```
kosegor, line 164: Update the doc please

x/metoken/keeper/metoken.go
```diff
@@ -43,8 +43,10 @@ func (k Keeper) GetAllIndexesBalances() []metoken.IndexBalances {
 }
 
 // getNextRebalancingTime returns next x/metoken re-balancing time in Milliseconds.
-func (k Keeper) getNextRebalancingTime() (time.Time, error) {
-	return store.GetTimeMs(k.store, keyPrefixNextRebalancingTime)
+// Returns 0 unix time if the time was not set before.
+func (k Keeper) getNextRebalancingTime() time.Time {
+	t, _ := store.GetTimeMs(k.store, keyPrefixNextRebalancingTime)
```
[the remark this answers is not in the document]
kosegor: Yep, we are safe. After the first execution the time will be updated

kosegor (approved):

util/store/store.go
```diff
@@ -162,13 +161,10 @@ func SetAddress(store sdk.KVStore, key []byte, val sdk.AccAddress) {
...
 
 // GetTimeMs retrieves time saved as Unix time in Miliseconds.
-// Returns sdkerrors.NotFound error if the value is not there, hence time = 0 is not supported.
-func GetTimeMs(store sdk.KVStore, key []byte) (time.Time, error) {
-	t := GetInteger[int64](store, key)
-	if t == 0 {
-		return time.Time{}, sdkerrors.ErrNotFound
-	}
-	return time.UnixMilli(t), nil
+// Returns if the value is not in the store, returns (0 unix time, false).
```
toteki, line 164:
```suggestion
// If the value is not in the store, returns (0 unix time, false).
```

x/incentive/keeper/store.go
```diff
@@ -147,7 +148,8 @@ func (k Keeper) setNextProgramID(ctx sdk.Context, id uint32) error {
 // getLastRewardsTime gets the last unix time incentive rewards were computed globally by EndBlocker.
 // panics if it would return a negative value.
```
toteki, line 149:
```suggestion
```

x/incentive/keeper/update.go
```diff
@@ -191,7 +191,7 @@ func (k Keeper) EndBlock(ctx sdk.Context) (skipped bool, err error) {
 
 	if blockTime <= prevTime {
 		// Avoids this and related issues: https://github.com/tendermint/tendermint/issues/8773
-		k.Logger(ctx).Error(
+		k.Logger(ctx).Debug(
```
toteki, line 194: This situation is relevant enough that it should be an error log

toteki (approved):

PR #2185, opened 2023-08-03
fix: x/oracle end_blocker panic

kosegor:
This PR contains a fix for the chain halting during the EndBlocker function. The scenario to reproduce it is as follows:
 * Start a fresh chain with `median_stamp_period < historic_stamp_period`.
 * Populate `x/leverage` registry with tokens and wait until `x/oracle` EndBlocker executes.
 * A panic will raise halting the chain with the following error: `panic: denom: ETH: empty price list passed in`


adamewozniak:
This doesn’t look like a bug in endblocker, rather a condition we shouldn’t expect to be in the genesis file. Imo we shouldn’t merge this in

toteki:
Maybe we can fix one layer lower in the functions - missing medians shouldn't be a chain halt, just like missing prices aren't.

We could do:
```
func (k Keeper) CalcAndSetHistoricMedian(
	ctx sdk.Context,
	denom string,
) error {
	historicPrices := k.historicPrices(ctx, denom, k.MaximumPriceStamps(ctx))
	if len(historicPrices) == 0 {
		return nil
	}
	median, err := decmath.Median(historicPrices)
	if err != nil {
		return errors.Wrap(err, "denom: "+denom)
	}

	block := uint64(ctx.BlockHeight())
	k.SetHistoricMedian(ctx, denom, block, median)
	return k.calcAndSetHistoricMedianDeviation(ctx, denom, median, historicPrices)
}
```
(the return nil check is the only change)

@adamewozniak

adamewozniak:
> Maybe we can fix one layer lower in the functions - missing medians shouldn't be a chain halt, just like missing prices aren't.
> 
> 
> 
> We could do:
> 
> ```
> 
> func (k Keeper) CalcAndSetHistoricMedian(
> 
> 	ctx sdk.Context,
> 
> 	denom string,
> 
> ) error {
> 
> 	historicPrices := k.historicPrices(ctx, denom, k.MaximumPriceStamps(ctx))
> 
> 	if len(historicPrices) == 0 {
> 
> 		return nil
> 
> 	}
> 
> 	median, err := decmath.Median(historicPrices)
> 
> 	if err != nil {
> 
> 		return errors.Wrap(err, "denom: "+denom)
> 
> 	}
> 
> 
> 
> 	block := uint64(ctx.BlockHeight())
> 
> 	k.SetHistoricMedian(ctx, denom, block, median)
> 
> 	return k.calcAndSetHistoricMedianDeviation(ctx, denom, median, historicPrices)
> 
> }
> 
> ```
> 
> (the return nil and its if statement are the only changes)
> 
> 
> 
> @adamewozniak 

Seems like a much more appropriate fix imo

kosegor:
I applied @toteki solution. @adamewozniak check it out.

x/oracle/keeper/historic_price.go
```diff
@@ -34,6 +34,11 @@ func (k Keeper) CalcAndSetHistoricMedian(
 	denom string,
 ) error {
 	historicPrices := k.historicPrices(ctx, denom, k.MaximumPriceStamps(ctx))
+	// this should never happen
+	if len(historicPrices) == 0 {
+		return nil
```
adamewozniak, line 39: Let's add a log here with the err saying this should never be the case + a test for it

kosegor: done

adamewozniak: Maybe a more descriptive msg here

adamewozniak (approved):
Lgtm, one nit

x/oracle/keeper/historic_price.go
kosegor: Could you provide some inside on why this shouldn't happen never? I'm not so related with the x/oracle module, that's why the message it's not so descriptive.

adamewozniak: this only happens when the historic stamp period is less than the median stamp period; this doesn't make any sense since the medians need to be calculated based off the existing historic stamps. 

It should never happen since there should always be historic stamps before we attempt to make any median stamps

kosegor: gotchu, improved the log

PR #2198, opened 2023-08-16
feat: x/metoken balances and prices query improvements

kosegor:
* Amplifying both queries with some additional information was requested from the FE: https://linear.app/umee/issue/UMEE-808/metoken-query-changes-for-frontend


gsk967 (approved):
LGTM, Pre Approving 
Small nit can you look into those comments

PR #2203, opened 2023-08-23
feat: rework proposal messages

robert-zaremba:
## Description

+ remove title and assure to use x/gov metatadata if x/gov authority is used
+ assure to set description otherwise.

## Considerations

* use `memo` instead of description
  * pro: no need to keep description
  * cons: people usually don't use memo
  * cons: not possible to enforce `memo` for non x/gov proposals.

### DIFF
--- a/tests/grpc/gov.go
+++ b/tests/grpc/gov.go
@@ -59,7 +59,7 @@ func OracleParamChanges(
 func UIBCIBCTransferSatusUpdate(umeeClient client.Client, status uibc.IBCTransferStatus) error {
 	msg := uibc.MsgGovSetIBCStatus{
 		Authority:   checkers.GovModuleAddr,
-		Description: "Update the ibc transfer status",
+		Description: "",
 		IbcStatus:   status,
 	}
 


gsk967 (commented):
LGTM 
Only one suggestion like , if you want remove title and description then remove it , no need for description again for emergency txn , anyway we can retrive from metadata field

app/upgradev3x3/migrate.go
```diff
@@ -5,93 +5,20 @@ import (
...
-}
-
 type migrator struct {
 	gov govkeeper.Keeper
 }
 
 // Creates migration handler for gov leverage proposals to new the gov system
 // and MsgGovUpdateRegistry type.
 func Migrator(gk govkeeper.Keeper, registry cdctypes.InterfaceRegistry) module.MigrationHandler {
-	registerInterfaces(registry)
```
toteki, line 32: check lint

proto/umee/leverage/v1/tx.proto
```diff
@@ -257,12 +257,14 @@ message MsgGovUpdateRegistry {
...
+
+  // authority is the address of the governance account or the Emergency Group.
   string authority = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
-  // Deprecated: use proposal metadata instead.
-  string title = 2 [deprecated = true];
-  // Deprecated: use proposal metadata instead.
-  string description = 3 [deprecated = true];
+  // description motivating the change. Should be used only when executing by the
+  // Emergency Group. Otherwise the x/gov Proposal metadata should be used.
+  string description = 3;
```
toteki, line 267: What other options do we have for recording emergency group intentions? transaction memo field?

Would description be any more visible / permanent in storage?

toteki: Would be ideal if we could eliminate this field, as Sai said

util/checkers/proposal.go
```diff
@@ -70,34 +69,28 @@ func EmergencyGroupAuthority(authority string, eg WithEmergencyGroup) (bool, err
...
 	}
 
-	if len(description) == 0 {
-		return errors.New("proposal description cannot be blank")
-	}
-	if len(description) > gov1b1.MaxDescriptionLength {
+	return nil
+}
+
+func ProposalDescription(description string) error {
```
toteki, line 87:
```suggestion
func ValidateProposalDescription(description string) error {
```
if we keep this, make sure name includes full meaning

app/upgradev3x3/migrate.go
```diff
@@ -5,93 +5,20 @@ import (
...
-	)
-}
-
 type migrator struct {
 	gov govkeeper.Keeper
 }
 
 // Creates migration handler for gov leverage proposals to new the gov system
 // and MsgGovUpdateRegistry type.
 func Migrator(gk govkeeper.Keeper, registry cdctypes.InterfaceRegistry) module.MigrationHandler {
```
toteki, line 16: Will we run into "unknown field: title" when attempting to load historic proposals for gov queries? Perhaps we need another migration similar to how this one worked. (Unless store load safely ignores unknown fields)

x/incentive/keeper/unit_test.go
```diff
@@ -132,7 +132,6 @@ func (k *testKeeper) initCommunityFund(funds ...sdk.Coin) {
 // and requires no errors. Use when setting up incentive scenarios.
 func (k *testKeeper) addIncentiveProgram(uDenom string, start, duration int64, reward sdk.Coin, fromCommunity bool,
 ) {
-	// govAccAddr := s.app.GovKeeper.GetGovernanceAccount(ctx).GetAddress().String()
```
gsk967, line 135: govAccAddr := checkers.GovModuleAddr
// govAccAddr := "govAcct"

gsk967 (approved):
LGTM

@robert-zaremba  I need one final confirmation , are you sure all explorers (pingpub/mintscan) showing proposal title and description from `metadata` feild  ?

toteki (approved):
I have a preference for using memo for emergency groups instead of sometimes using description and sometimes not. It's not a strong preference though

We will have to make sure (during a testnet) that when running this code against existing networks that we don't get problems such as `unknown field: title` when querying previous `MsgGovUpdateRegistry` which have that field.

gsk967:
@toteki  `title` is not deleted, it is reserved, so there will be no issues like `unknown field: title` on Marshal/Unmarshaling 

and `description` field make sense more than `memo` field because we already have `memo` param on txn , so sometimes it will make confuse

PR #2208, opened 2023-08-28
feat: v6.0 proposal migration

robert-zaremba:
## Description

* we still need to update the emergency group address

### DIFF
--- a/app/upgrades.go
+++ b/app/upgrades.go
@@ -68,9 +68,7 @@ func (app *UmeeApp) registerUpgrade6(upgradeInfo upgradetypes.Plan) {
 			if err := upgradev6x0.Migrate(ctx, app.GovKeeper); err != nil {
 				return fromVM, err
 			}
-			if err := app.LeverageKeeper.SetParams(leveragetypes.DefaultParams()); err != nil {
-				return fromVM, err
-			}
+			app.LeverageKeeper.SetParams(ctx, leveragetypes.DefaultParams())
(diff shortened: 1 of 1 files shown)


toteki (approved):
Looks very straightforward

gsk967:
@robert-zaremba  @toteki  Instead of upgradev6 , we can follow this folder struct 
```
- app
 - upgrades
   - v6 
   - v3 
```

app/upgrades.go
```diff
@@ -56,32 +57,20 @@ func (app UmeeApp) RegisterUpgradeHandlers(bool) {
...
-
 func (app *UmeeApp) registerUpgrade6(upgradeInfo upgradetypes.Plan) {
 	planName := "v6.0"
 	gravityModuleName := "gravity" // hardcoded to avoid dependency on GB module
 
 	app.UpgradeKeeper.SetUpgradeHandler(planName,
 		func(ctx sdk.Context, plan upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
+			if err := upgradev6x0.Migrate(ctx, app.GovKeeper); err != nil {
+				return fromVM, err
+			}
```
toteki, line 70:
```suggestion
			}
			if err := app.LeverageKeeper.SetParams(leveragetypes.DefaultParams()); err != nil {
				return fromVM, err
			}
```
We'll need this after my params PR (but I don't want to create a merge conflict in this function by changing too much there)

PR #2211, opened 2023-08-28
leverage: require params to be initialized

robert-zaremba:
## Description
follow up to : https://github.com/umee-network/umee/pull/2209/files#r1307683345

All data should be in storage to make it verifiable.


toteki (dismissed):
This unnecessarily removes a safeguard and creates a panic for no gain

gsk967:
I think don't do `panic` just as you say just return empty param struct not `defaultParams` because `defaultParams`  are not stored in the state

toteki (approved):

PR #2214, opened 2023-08-29
refactore(ugov): don't expose store internals

robert-zaremba:
## Description

1. Store internals should not be exposed 
2. Ideally Keeper interface should be used in migrations, rather than a store directly.


gsk967 (approved):
LGTM

PR #2215, opened 2023-08-30
fix: proto query fixed

kosegor:
Query params don't accept sdk.Coin: https://github.com/grpc-ecosystem/grpc-gateway/blob/094a6fe78b3ca888297d090185cdf30f0e42e157/runtime/query.go#L395-L404


kosegor:
> I ran `make proto-all` which updated the swagger docs

Thanks, `make proto-all` never worked for me so I'm using a custom cmd to generate the files.

PR #2229, opened 2023-09-06
chore: set test emergency group in the upgrade handler

robert-zaremba:
## Description

Last thing before `beta-2`


toteki (approved):
Testnet address I presume?

PR #2247, opened 2023-09-15
fix: gov special assets fix for 6.0

toteki:


gsk967 (commented):
Can you add `invalid` MsgGovUpdateSpecialAssets just like existed code missing some fields

PR #2243, opened 2023-09-13
feat(WIP): adding timestamp to exchange rates of denoms

gsk967:
## Description

closes: #XXXX

---

### Author Checklist

_All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues._

I have...

- [ ] included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] added `!` to the type prefix if API or client breaking change
- [ ] added appropriate labels to the PR
- [ ] targeted the correct branch (see [PR Targeting](https://github.com/umee-network/umee/blob/main/CONTRIBUTING.md#pr-targeting))
- [ ] provided a link to the relevant issue or specification
- [ ] added a changelog entry to `CHANGELOG.md`
- [ ] included comments for [documenting Go code](https://blog.golang.org/godoc)
- [ ] updated the relevant documentation or specification
- [ ] reviewed "Files changed" and left comments if necessary
- [ ] confirmed all CI checks have passed

### Reviewers Checklist

_All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items._

I have...

- [ ] confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] confirmed all author checklist items have been addressed
- [ ] reviewed state machine logic
- [ ] reviewed API design and naming
- [ ] reviewed documentation is accurate
- [ ] reviewed tests and test coverage
- [ ] manually tested (if applicable)

PR #2249, opened 2023-09-15
feat(oracle): include timestamp into save exchange rates

robert-zaremba:
## Description

Exchange rates should be stored with a timestamp, not separately as it's done in https://github.com/umee-network/umee/pull/2243

1. today the end blocker removes ALL prices before setting the new ones
2. instead, we don't want to remove ALL prices , but overwrite the old ones with a new ones. 
3. If a new one price is missing, an old one will be kept 

With timestamp we can assess if the old prices is not too old.

### DIFF
--- a/x/oracle/abci_test.go
+++ b/x/oracle/abci_test.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/cosmos/cosmos-sdk/simapp"
 	sdk "github.com/cosmos/cosmos-sdk/types"
-	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	minttypes "github.com/cosmos/cosmos-sdk/x/mint/types"
 	"github.com/cosmos/cosmos-sdk/x/staking"
 	"github.com/cosmos/cosmos-sdk/x/staking/teststaking"
@@ -177,8 +176,8 @@ func (s *IntegrationTestSuite) TestEndBlockerVoteThreshold() {
 
 	for _, denom := range app.OracleKeeper.AcceptList(ctx) {
 		rate, err := app.OracleKeeper.GetExchangeRate(ctx, denom.SymbolDenom)
-		s.Require().ErrorIs(err, sdkerrors.Wrap(types.ErrUnknownDenom, denom.SymbolDenom))
-		s.Require().Equal(sdk.ZeroDec(), rate)
+		s.Require().ErrorIs(err, types.ErrUnknownDenom.Wrap(denom.SymbolDenom))
+		s.Require().Equal(types.ExchangeRate{}, rate)
 	}
 
 	// Test: val2 and val3 votes.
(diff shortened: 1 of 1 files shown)


x/leverage/keeper/oracle.go
```diff
@@ -56,15 +57,17 @@ func (k Keeper) TokenPrice(ctx sdk.Context, baseDenom string, mode types.PriceMo
 		}
 	}
 
+	// TODO: need to use spotPrice.Timestamp to make a decision about the price
+
 	switch mode {
 	case types.PriceModeSpot:
-		price = spotPrice
+		price = spotPrice.Rate
```
toteki, line 64: This way of doing it is straightforward on the leverage side

x/leverage/keeper/oracle.go
```diff
@@ -56,15 +57,17 @@ func (k Keeper) TokenPrice(ctx sdk.Context, baseDenom string, mode types.PriceMo
 		}
 	}
 
+	// TODO: need to use spotPrice.Timestamp to make a decision about the price
```
toteki, line 60: Leverage will probably have a param which sets the maximum age for spot prices

x/oracle/keeper/keeper.go
```diff
@@ -110,16 +107,16 @@ func (k Keeper) GetExchangeRateBase(ctx sdk.Context, denom string) (sdk.Dec, err
...
 // SetExchangeRate sets the consensus exchange rate of USD denominated in the
 // denom asset to the store.
-func (k Keeper) SetExchangeRate(ctx sdk.Context, denom string, exchangeRate sdk.Dec) {
-	store := ctx.KVStore(k.storeKey)
-	bz := k.cdc.MustMarshal(&sdk.DecProto{Dec: exchangeRate})
-	denom = strings.ToUpper(denom)
-	store.Set(types.KeyExchangeRate(denom), bz)
+func (k Keeper) SetExchangeRate(ctx sdk.Context, denom string, rate sdk.Dec) {
+	key := types.KeyExchangeRate(denom)
+	val := types.ExchangeRate{Rate: rate, Timestamp: ctx.BlockTime()}
```
toteki, line 117: ctx.BlockTime() is still vulnerable to odd bugs, such as having zero value at times (which is why I store `LastInterestTime` and `LastRewardTime` with a safety layer in `leverage` and `incentive` modules)

toteki: we have run into bugs on mainnet and testnets where ctx.BlockTime() has been wildly inaccurate - either zero or years in the past - for a single block. That's why I have "ignore zero / negative / decrease block time" when doing `LastInterestTime` for example.

Seems to come from tebndermint layer so we can't fix it

x/oracle/types/keys.go
```diff
@@ -33,7 +34,7 @@ var (
 // KeyExchangeRate - stored by *denom*
 func KeyExchangeRate(denom string) []byte {
 	// append 0 for null-termination
-	return util.ConcatBytes(1, KeyPrefixExchangeRate, []byte(denom))
+	return util.ConcatBytes(1, KeyPrefixExchangeRate, []byte(strings.ToUpper(denom)))
```
gsk967, line 37: this key will override previous stored price of denom

we need to store the prices with timestamps for storing previous prices
```
tbz, _  := timestamp.MarshalBinary()
key = util.ConcatBytes(1, KeyPrefixExchangeRate, []byte(strings.ToUpper(denom)),tbz)
```

gsk967 (approved):
LGTM

x/oracle/types/keys.go
gsk967: Ok

PR #2256, opened 2023-09-20
feat(client): unify and refactore client package

robert-zaremba:
## Description

+ better fromAccount management
+ remove code duplication
+ msg checks
+ metoken queries in the umee/client
+ added logger to Tx and Query Client
+ use RPC queries rather than REST


kosegor (approved):
LGTM!

PR #2260, opened 2023-09-21
fix(oracle): avg params storage.

robert-zaremba:
## Description

+ v6.1 upgrade handler with code to set AvgParams
+ remove the v6.0.2 fallback to use default AvgParams
+ fix the key scheme by moving the AvgParams store key from `params.go` (which is dedicated for the x/param spaces) to `keys.go`.

### DIFF
--- a/client/metoken.go
+++ b/client/metoken.go
@@ -1,0 +1,58 @@
+package client
+
+import (
+	"github.com/umee-network/umee/v6/util/sdkutil"
+	"github.com/umee-network/umee/v6/x/metoken"
+)
+
+func (c Client) MetokenQClient() metoken.QueryClient {
+	return metoken.NewQueryClient(c.Query.GrpcConn)
+}
+
+func (c Client) QueryMetokenParams() (metoken.Params, error) {
+	ctx, cancel := c.NewQCtx()
+	defer cancel()
+
+	resp, err := c.MetokenQClient().Params(ctx, &metoken.QueryParams{})
+	if err != nil {
+		return metoken.Params{}, err
+	}
+	return resp.Params, err
+}
+
+func (c Client) QueryMetokenIndexBalances(denom string) (*metoken.QueryIndexBalancesResponse, error) {
+	ctx, cancel := c.NewQCtx()
+	defer cancel()
+
+	msg := &metoken.QueryIndexBalances{MetokenDenom: denom}
+	if err := sdkutil.ValidateProtoMsg(msg); err != nil {
+		return nil, err
+	}
+	return c.MetokenQClient().IndexBalances(ctx, msg)
+}
+
+func (c Client) QueryMetokenIndexes(denom string) (*metoken.QueryIndexesResponse, error) {
+	ctx, cancel := c.NewQCtx()
+	defer cancel()
+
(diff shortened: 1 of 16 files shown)


x/oracle/types/keys.go
```diff
@@ -19,16 +19,17 @@ const (
...
+	KeyPrefixExchangeRate                 = []byte{1}  // prefix for each key to a rate
+	KeyPrefixFeederDelegation             = []byte{2}  // prefix for each key to a feeder delegation
+	KeyPrefixMissCounter                  = []byte{3}  // prefix for each key to a miss counter
+	KeyPrefixAggregateExchangeRatePrevote = []byte{4}  // prefix for each key to a aggregate prevote
+	KeyPrefixAggregateExchangeRateVote    = []byte{5}  // prefix for each key to a aggregate vote
+	KeyPrefixMedian                       = []byte{6}  // prefix for each key to a price median
+	KeyPrefixMedianDeviation              = []byte{7}  // prefix for each key to a price median standard deviation
+	KeyPrefixHistoricPrice                = []byte{8}  // prefix for each key to a historic price
+	KeyPrefixAvgCounter                   = []byte{9}  // prefix for each key to a historic avg price counter
+	KeyLatestAvgCounter                   = []byte{10} // key where we store the latest avg price counter
```
gsk967, line 31: `[]byte{0x10} != []byte{10}` , so you have keep `[]byte{0x10}` because it is already stored previous AvgCounter

x/oracle/keeper/migrations.go
```diff
@@ -21,6 +21,8 @@ func (m Migrator) Migrate1to2(ctx sdk.Context) error {
 	m.keeper.SetMedianStampPeriod(ctx, 1)
 	m.keeper.SetMaximumPriceStamps(ctx, 1)
 	m.keeper.SetMaximumMedianStamps(ctx, 1)
+	// NOTE: call to m.SetAvgPeriodAndShift is missing here, and caused a chain halt
```
gsk967, line 24: there `SetAvgPeriodAndShift` is not part of the `Migrate1to2`

gsk967 (changes requested):

app/upgrades.go
```diff
@@ -57,6 +57,21 @@ func (app UmeeApp) RegisterUpgradeHandlers() {
 	app.registerUpgrade5_1(upgradeInfo)
 	app.registerUpgrade("v5.2", upgradeInfo) // v5.2 migration is not compatible with v6, so leaving default here.
 	app.registerUpgrade6(upgradeInfo)
+	app.registerUpgrade6_1("v6.1", upgradeInfo)
+}
+
+func (app *UmeeApp) registerUpgrade6_1(planName string, _ upgradetypes.Plan) {
+	app.UpgradeKeeper.SetUpgradeHandler(planName,
+		func(ctx sdk.Context, plan upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
+			ctx.Logger().Info("-----------------------------\n-----------------------------")
```
toteki, line 66: todo: remove

gsk967 (approved):
LGTM

PR #2261, opened 2023-09-21
chore: use go-1.21

robert-zaremba:


gsk967 (approved):
LGTM

gsk967:
I have small suggestion can we do after the release of umee  with cosmos-sdk 47

PR #2263, opened 2023-09-22
feat: account summary to display both spot and leverage prices

toteki:


x/leverage/keeper/inspector.go
```diff
@@ -88,6 +88,7 @@ func (q Querier) Inspect(
 				Collateral: symbolDecCoins(collateral, exchangeRates),
 				Borrowed:   symbolDecCoins(borrowed, exchangeRates),
 			},
+			Info: position.String(),
```
toteki, line 91:
```suggestion
			Info: "",
```

x/leverage/keeper/inspector.go
```diff
@@ -171,4 +231,4 @@ func neat(num sdk.Dec) float64 {
 	// Truncate the float at a certain precision (can be negative)
 	x := math.Pow(10, float64(precision))
 	return float64(int(n*x)) / x
-}
+}
```
toteki, line 234:
```suggestion
}

```

x/leverage/keeper/inspector.go
```diff
@@ -121,6 +122,65 @@ func (q Querier) Inspect(
...
+		Analysis: &types.RiskInfo{
+			Borrowed:    neat(borrowedValue),
+			Liquidation: neat(liquidationThreshold),
+			Value:       neat(collateralValue),
+		},
+		Position: &types.DecBalances{
+			Collateral: symbolDecCoins(collateral, exchangeRates),
+			Borrowed:   symbolDecCoins(borrowed, exchangeRates),
+		},
+		Info: position.String() + "\n\n" + borrowPosition.String(),
```
toteki, line 178: This string field is meant to be unstable, but also easier to manipulate in patch releases or modified binaries without changing proto.

Currently, when inspecting a single account, it will show the account's positions arranged both by borrow logic and liquidation logic, for a good amount of visibility.

Since this query is single target, it will be available on any node without causing iteration-based DDoS-like problems.

proto/umee/leverage/v1/query.proto
```diff
@@ -289,6 +297,21 @@ message QueryAccountSummaryResponse {
...
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
+    (gogoproto.nullable)   = false
+  ];
+  // Spot Collateral Value is collateral value but always uses spot prices.
+  string spot_collateral_value = 7 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
+    (gogoproto.nullable)   = false
+  ];
+  // Spot Borrowed Value is borrowed value but always uses spot prices.
+  string spot_borrowed_value = 8 [
```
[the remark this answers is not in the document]
toteki: Value denotes USD, but borrowed would denote tokens, at least if we follow the rest of the leverage messages

PR #2267, opened 2023-09-27
feat: recent price mode support

toteki:
Adjust account summary query to display leverage logic prices (considering the more cautious of spot and historic generally) but allow expired prices.

Leverage transactions, and transaction-simulating queries like max borrow, reject expired prices.

Hard-codes max price age (for now) at 3 minutes for transactions' sake. Compare to current 30 second price voting window, but 3 hour historic median period


proto/umee/leverage/v1/query.proto
```diff
@@ -286,28 +286,31 @@ message QueryAccountSummaryResponse {
   // Borrow Limit is the maximum Borrowed Value the account is allowed to reach through direct borrowing.
   // The lower of spot or historic price for each collateral token is used when calculating borrow limits.
   // Computation skips collateral which is missing an oracle price, potentially resulting in a lower borrow
-  // limit than if prices were all available.
+  // limit than if prices were all available. Will be null if an oracle price required for computation is
+  // missing.
   string borrow_limit = 4 [
     (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
-    (gogoproto.nullable)   = false
+    (gogoproto.nullable)   = true
```
toteki, line 293: API change

x/leverage/keeper/oracle.go
```diff
@@ -12,6 +12,9 @@ import (
 	oracletypes "github.com/umee-network/umee/v6/x/oracle/types"
 )
 
+// TODO: parameterize this
+const MaxSpotPriceAge = 180 // 3 minutes
```
toteki, line 16: We can add this ad a parameter at some point (along with MinimumBorrowFactor)

toteki: imported into keeper_test package so can't lowercase it. Added comment though

gsk967:
@adam can you check why we are clearning exg_rates  if we clear exg_rates then we will get `unknown denom` error in `GetExchangeRate` function, so resolve this we need keep the old prices 

@robert-zaremba  WDYT about this ?  

We are getting this results after stoping the price-feeder but have to get old results instead of `unknown denom` error

```json
http://localhost:1316/umee/leverage/v1/market_summary?denom=uumee
{
  "symbol_denom": "UMEE",
  "exponent": 6,
  "oracle_price": null,
  "uToken_exchange_rate": "1.000000005967583058",
  "supply_APY": "0.000454358176141041",
  "borrow_APY": "0.051000989990100099",
  "supplied": "100000",
  "reserved": "0",
  "collateral": "100000",
  "borrowed": "1001",
  "liquidity": "99000",
  "maximum_borrow": "90000",
  "maximum_collateral": "100000",
  "minimum_liquidity": "10000",
  "uToken_supply": "100000",
  "available_borrow": "88999",
  "available_withdraw": "89000",
  "available_collateralize": "0",
  "oracle_historic_price": null,
  "errors": "oracle: UMEE: unknown denomoracle: UMEE: unknown denom"
}```

**Review by gsk967 — APPROVED** [review]
@adam Please check my comments 
 I am per-approving this pr
  you can remove https://github.com/umee-network/umee/blob/adam/recent/x/oracle/abci.go#L57 this line in next your pr

File: x/leverage/keeper/oracle.go [hunk 3]
```diff
@@ -39,9 +43,20 @@ func (k Keeper) TokenPrice(ctx sdk.Context, baseDenom string, mode types.PriceMo
 		if err != nil {
 			return sdk.ZeroDec(), t.Exponent, errors.Wrap(err, "oracle")
 		}
+		if !mode.AllowsExpired() {
```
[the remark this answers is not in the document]
toteki: This is required for frontend. Right now, when a price goes down (for example, every day for 10 minutes during the osmosis epoch for some assets), users are seeing their positions values go to zero on the web app, which is not a good experience.

Even though this does not affect transactions (which require a non-expired price to function, both on mainnet and in this feature), allowing queries to use old prices instead of zero during outages is the improvement we've been requested to make.

x/leverage/keeper/oracle.go
```diff
@@ -12,6 +12,9 @@ import (
 	oracletypes "github.com/umee-network/umee/v6/x/oracle/types"
 )
 
+// TODO: parameterize this
+const MaxSpotPriceAge = 180 // 180 seconds = 3 minutes
```
toteki, line 16: @robert-zaremba This is the time limit on prices used for transactions.

Prices older than this are considered "Expired", and can only be used for queries.

x/leverage/keeper/errors.go
```diff
@@ -21,6 +21,7 @@ func nonOracleError(err error) bool {
 	if errors.IsOf(err,
 		leveragetypes.ErrInvalidOraclePrice,
 		leveragetypes.ErrNoHistoricMedians,
+		leveragetypes.ErrExpiredOraclePrice,
 		oracletypes.ErrUnknownDenom,
```
toteki, line 25: @robert-zaremba This is where having registered error types is currently helpful.

We'd have to do this check manually by making an `OracleError` type in both the leverage and oracle modules, and then only including speciic cases, if we wanted to delete that one line registering a new error type.

x/leverage/types/errors.go
```diff
@@ -40,6 +40,7 @@ var (
 	ErrUndercollateralized   = errors.Register(ModuleName, 402, "borrow positions are undercollateralized")
 	ErrLiquidationIneligible = errors.Register(ModuleName, 403, "borrower not eligible for liquidation")
 	ErrNoHistoricMedians     = errors.Register(ModuleName, 405, "insufficient historic medians available")
+	ErrExpiredOraclePrice    = errors.Register(ModuleName, 406, "no recent oracle price")
```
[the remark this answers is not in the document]
toteki: We'll need some kind of cross-module error typing system then. See my comment on `nonOracleError` function

x/leverage/types/oracle.go
```diff
@@ -4,12 +4,42 @@ package types
...
+	PriceModeQueryHigh
 	// Low mode uses the lower of either Spot or Historic prices
 	PriceModeLow
+	// QueryLow mode uses the lower of either Spot or Historic prices, allowing expired spot prices.
+	PriceModeQueryLow
 )
+
+// IgnoreHistoric transforms a price mode in a way that uses spot prices instead of historic. This
+// happens when historic prices would normally be used by a token has zero required historic medians.
+func (mode PriceMode) IgnoreHistoric() PriceMode {
```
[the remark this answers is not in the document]
toteki: added

PR #2288, opened 2023-10-17
chore: change consensus.block.max_size to 4 MB

robert-zaremba:
## Description

As per Amulet team analysis, the existing 21MB block size may expose security risk and teams should consider reducing it. In this PR we reduce it to 4 MB, giving enough room for Cosmowasm deployments and big transactions. Blocks full of oracle txs are less than 40kB.


gsk967 (approved):
LGTM

toteki (approved):

PR #2316, opened 2023-11-07
feat: feed oracle with meTokens price every endBlock

kosegor:
* Sends every meToken price each endBlock to x/oracle
 * Avoid adding meTokens to x/oracle accept list when they are added to x/leverage registry. This is done to exclude validators from reporting and voting on the price, as well as being slashing for not reporting it.


toteki (commented):
Check to see if GetExchangeRate works in oracle module.

(It might not know the symbol denom <-> base denom relationship for metokens)

contrib/scripts/mockgen.sh
```diff
@@ -2,7 +2,7 @@
...
 mockgen_cmd="go run github.com/golang/mock/mockgen"
 
-$mockgen_cmd -source ./x/ugov/keeper_interfaces.go -package mocks -destination ./x/ugov/mocks/keepers.go
-$mockgen_cmd -source ./x/uibc/expected_keepers.go -package mocks -destination ./x/uibc/mocks/keepers.go
-$mockgen_cmd -source ./x/metoken/expected_keepers.go -package mocks -destination ./x/metoken/mocks/keepers.go
-$mockgen_cmd -source ./app/inflation/expected_keepers.go -package mocks -destination ./app/inflation/mocks/keepers.go
+$mockgen_cmd -source ./../../x/ugov/keeper_interfaces.go -package mocks -destination ./../..//x/ugov/mocks/keepers.go
+$mockgen_cmd -source ./../../x/uibc/expected_keepers.go -package mocks -destination ./../..//x/uibc/mocks/keepers.go
+$mockgen_cmd -source ./../../x/metoken/expected_keepers.go -package mocks -destination ./../../x/metoken/mocks/keepers.go
+$mockgen_cmd -source ./../../app/inflation/expected_keepers.go -package mocks -destination ./../..//app/inflation/mocks/keepers.go
```
gsk967, line 8: Keep old one only because when we run `make mocks` , it will run from main directory

kosegor:
> utACK. Did you check if we need to check slashing condition in oracle? We don't want oracle module to both:
> 
> * require validators to submit meToken prices
> * slash them if they don't submit

Yes, I checked it with Adam and oracle team. After the token is added to leverage registry we are NOT adding it to oracle accept list.

PR #2328, opened 2023-11-16
chore: rename UMEE token to UX token part1

robert-zaremba:
## Description

Changes:
* changed umee denom Name, Display to `UX`
* but does NOT change 
  * the stake denom itself, it's still `uumee`  (this will cause problems with IBC and transfers)
  * Symbol, it's still `UMEE` (this will cause errors with Oracle)
* added aliases to the stake denom: `ux` and `uux`.

This won't allow to query nor send tokens by alias, but will allow to show it in the metadata.

Changing base name creates many problems, which I will document in Notion.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new search function for the native staking token metadata.
  - Implemented token denomination migration from UMEE to UX.
  - Initialized new governance parameter for minimum initial deposit ratio.
  - Added migration functions for IBC outflow sum and parameters.

- **Refactor**
  - Extracted token metadata creation into a separate function for better modularity.

- **Documentation**
  - Added notes to emphasize the immutability of the `AccountAddressPrefix`.

- **Tests**
  - Added new tests to validate token denominations.

- **Style**
  - Updated constants to reflect new token denomination UX and preserve legacy UMEE.

- **Chores**
  - Updated regex to support new "ux" token denomination.

### DIFF
--- a/app/modules.go
+++ b/app/modules.go
@@ -44,13 +44,12 @@ func (BankModule) DefaultGenesis(cdc codec.JSONCodec) json.RawMessage {
 				Denom:    appparams.BondDenom,
 				Exponent: 0,
 				Aliases: []string{
-					"microumee",
+					"microumee", appparams.DefaultBaseDenom,
 				},
-			},
-			{
+			}, {
 				Denom:    appparams.DisplayDenom,
 				Exponent: 6,
-				Aliases:  []string{},
+				Aliases:  []string{appparams.LegacyDisplayDenom},
 			},
 		},
 	}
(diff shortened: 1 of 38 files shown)


app/modules.go
```diff
@@ -44,17 +43,20 @@ func (BankModule) DefaultGenesis(cdc codec.JSONCodec) json.RawMessage {
 				Denom:    appparams.BondDenom,
 				Exponent: 0,
 				Aliases: []string{
-					"microumee",
+					"microumee", appparams.BaseExtraDenom,
```
toteki, line 46: Since we're already been running with an alias, I don't have any more concerns about adding another one here

toteki (approved):

gsk967 (approved):
LGTM

PR #2337, opened 2023-11-27
chore(oracle): update miss counter docs

robert-zaremba:
## Description

Oracle docs update: miss counter increase  based on ballot support.

### DIFF
--- a/x/oracle/README.md
+++ b/x/oracle/README.md
@@ -50,7 +50,7 @@ Validators must first pre-commit to a set of exchange rates, then in the subsequ
 
   For each exchange rate, if the total voting power of submitted votes exceeds 50%, the weighted median of the votes is recorded on-chain as the effective rate for that denomination against USD for the following `VotePeriod` `P_t+1`.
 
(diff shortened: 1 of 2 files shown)


x/oracle/README.md
```diff
@@ -78,8 +78,8 @@ The reward pool is not distributed all at once, but instead over a period of tim
 A `VotePeriod` during which either of the following events occur is considered a "miss":
 
 - The validator fails to submits a vote for **each and every** exchange rate specified in `AcceptList`.
-
 - The validator fails to vote within the `reward band` around the weighted median for one or more denominations.
+- Validators don't provide enough price votes for a given denom (ballot denom support is less than the required `VoteThreshold`) - in that case all validators will receive a miss.
```
gsk967, line 82: Can you make this point in **BOLD Text**

```suggestion
- Validators don't provide enough price votes for a given denom (ballot denom support is less than the required `VoteThreshold`) - **in that case all validators will receive a miss.**
```

robert-zaremba: OK, good idea

PR #2369, opened 2023-12-27
feat: query RegisteredTokenMarkets

toteki:
## Summary by CodeRabbit

- **New Features**
  - Introduced a new service to query registered tokens and their market summaries.
  - Users can now view token market information within the app.

- **Documentation**
  - Added new message types for better querying and response structure in token market services.

- **Tests**
  - Implemented tests for the new token market querying functionality.


gsk967 (approved):
LGTM

gsk967:
@toteki  WDYT  may be we can add this query as optional integration (customizable query) to binary

toteki:
Since this doesn't iterate over users, it doesn't need to be enabled by flag like the `-l` queries. Should be safe as-is

proto/umee/leverage/v1/query.proto
```diff
@@ -24,6 +24,12 @@ service Query {
     option (google.api.http).get = "/umee/leverage/v1/registered_tokens";
   }
 
+  // RegisteredTokenMarkets queries for all the registered tokens and their market summaries.
+  rpc RegisteredTokenMarkets(QueryRegisteredTokenMarkets)
+      returns (QueryRegisteredTokenMarketsResponse) {
+    option (google.api.http).get = "/umee/leverage/v1/registered_token_markets";
```
robert-zaremba, line 30:
```suggestion
  // RegisteredTokensWithMarkets queries for all the registered tokens and their market summaries.
  rpc RegisteredTokensWithMarkets(QueryRegisteredTokensWithMarkets)
      returns (QueryRegisteredTokensWithMarketsResponse) {
    option (google.api.http).get = "/umee/leverage/v1/registered_tokens_with_markets";
```

x/leverage/keeper/grpc_query.go
```diff
@@ -62,6 +62,35 @@ func (q Querier) RegisteredTokens(
...
+
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	tokens := q.Keeper.GetAllRegisteredTokens(ctx)
+	markets := []types.TokenMarket{}
+
+	for _, token := range tokens {
+		marketSumnmary, err := q.MarketSummary(goCtx, &types.QueryMarketSummary{Denom: token.BaseDenom})
+		if err != nil {
+			// absorb overall query error into struct, which may be empty, but proceed with this query
+			marketSumnmary.Errors += err.Error()
```
robert-zaremba, line 81: why do we return errors as a String as a part of the "normal" object response?

robert-zaremba (approved):
utACK. Left one comment to discuss the API

x/leverage/keeper/grpc_query.go
toteki: Mimics `MarketSummary` behavior.

This was before `errors.Join` existed, and it runs multiple calculations where we need to return the fields we can even if some of the others error. The string field kept the API simple (always a known, readable type for debugging)

PR #2370, opened 2023-12-27
feat(uibc): add missing params to MsgGovUpdateQuota

robert-zaremba:
## Summary by CodeRabbit

- **New Features**
  - Expanded `MsgGovUpdateQuota` to include additional quota management fields.
  - Introduced `ICS20Memo` message with support for multiple message types.

- **Improvements**
  - Enhanced validation logic in `MsgGovUpdateQuota` for better error handling.
  - Refined quota management with new parameters and validation checks.

- **Bug Fixes**
  - Adjusted execution order for specific processes to ensure proper handling.

- **Documentation**
  - Updated CHANGELOG.md to reflect new features, improvements, and bug fixes.


util/checkers/number.go
```diff
@@ -52,3 +52,19 @@ func DecInZeroOne(a sdk.Dec, name string, oneInclusive bool) error {
 	}
 	return nil
 }
+
+// DecNotNegative checks if a is defined and a >= 0
+func DecNotNegative(a sdk.Dec, paramName string, errs []error) []error {
```
toteki, line 57: For these functions, they should probably just return `error` and any `append(errs, err)` can be kept outside of them. This will keep them more standard in case we want to use them anywhere else

x/uibc/msg.go
```diff
@@ -26,21 +25,24 @@ func (msg *MsgGovUpdateQuota) String() string {
...
+		errs = append(errs, err)
 	}
 
-	if msg.Total.IsNil() || !msg.Total.IsPositive() {
-		return sdkerrors.ErrInvalidRequest.Wrap("total quota must be positive")
-	}
-	if msg.PerDenom.IsNil() || !msg.PerDenom.IsPositive() {
-		return sdkerrors.ErrInvalidRequest.Wrap("quota per denom must be positive")
-	}
+	errs = checkers.DecPositive(msg.Total, "total quota", errs)
```
toteki, line 33: To keep this "one line per check" feature while implementing the comment above, you could make

```
errs = checkers.AppendErr(errs, checkers.DecPositive(msg.Total, "total quota")
```

where `AppendErr` appends an error if non-nil

x/uibc/msg.go
```diff
@@ -26,21 +25,24 @@ func (msg *MsgGovUpdateQuota) String() string {
...
+	errs = checkers.DecPositive(msg.PerDenom, "per_denom quota", errs)
+	errs = checkers.DecNotNegative(msg.InflowOutflowQuotaBase, "inflow_outflow_quota_base", errs)
+	errs = checkers.DecNotNegative(msg.InflowOutflowTokenQuotaBase, "inflow_outflow_token_quota_base", errs)
+	errs = checkers.DecNotNegative(msg.InflowOutflowQuotaRate, "inflow_outflow_quota_rate", errs)
 	if msg.Total.LT(msg.PerDenom) {
-		return sdkerrors.ErrInvalidRequest.Wrap("total quota must be greater than or equal to per_denom quota")
+		errs = append(errs, errors.New("total quota must be greater than or equal to per_denom quota"))
+	}
+	if msg.InflowOutflowQuotaBase.LT(msg.InflowOutflowTokenQuotaBase) {
+		errs = append(errs, errors.New("inflow_outflow_quota_base must be greater than or equal than inflow_outflow_token_quota_base"))
```
toteki, line 42: lint

x/uibc/msg_test.go
```diff
@@ -58,6 +64,10 @@ func TestMsgGovUpdateQuota(t *testing.T) {
 			name:   "invalid total quota with respect to per denom",
 			msg:    invalidTotalQuota,
 			errMsg: "total quota must be greater than or equal to per_denom quota",
+		}, {
+			name:   "invalid inflow outflow quota abse with respect to per denom",
```
toteki, line 68:
```suggestion
			name:   "invalid inflow outflow quota base with respect to per denom",
```

toteki (approved):
See comments / Otherwise approved

x/uibc/msg.go
robert-zaremba: I think we should use more often the aggregated errors which came to `go v1.20`. So adding `checkers.AppendErr` is additional code we don't need.

toteki: `errors.Join` works too

robert-zaremba: yes, that's why we can have one list of errors, and then do `errors.Join`. It reduces the code, without adding additional layers.

robert-zaremba: so alternative would be, instead of
```errs = checkers.XXX(...)
errs = checkers.XXX(...)
errs = checkers.XXX(...)
errs = checkers.XXX(...)
errs = checkers.XXX(...)
return errors.Join(errs...)
```
we would do
```err1 = checkers.XXX(...)
err2 = checkers.XXX(...)
err3 = checkers.XXX(...)
err4 = checkers.XXX(...)
err5 = checkers.XXX(...)
return errors.Join(err1, err2, err3, err4, err5)
```

The problem here is that we can  forget one of the err1, err2, ....

robert-zaremba: However if we would do
```
err = errors.Join(err, checkers.XXX(...))
err = errors.Join(err, checkers.XXX(...))
err = errors.Join(err, checkers.XXX(...))
err = errors.Join(err, checkers.XXX(...))
err = errors.Join(err, checkers.XXX(...))
return err
```
The code is more bloated, and creates lof of wrapping, so we have many nested errors.

gsk967: In my opinion this is fine because we don’t know how many errors we will get 
```go
errs = checkers.XXX(...) -> append errors
errs = checkers.XXX(...)
errs = checkers.XXX(...)
errs = checkers.XXX(...)
errs = checkers.XXX(...) -> append errors
return  errs 
```

kosegor: > However if we would do
> 
> ```
> err = errors.Join(err, checkers.XXX(...))
> err = errors.Join(err, checkers.XXX(...))
> err = errors.Join(err, checkers.XXX(...))
> err = errors.Join(err, checkers.XXX(...))
> err = errors.Join(err, checkers.XXX(...))
> return err
> ```
> 
> The code is more bloated, and creates lof of wrapping, so we have many nested errors.

This looks better to me

gsk967: But here if `checkers.XXX(...)` is nil , it will append `nil` to errors @kosegor

gsk967 (approved):
LGTM

x/uibc/msg.go
toteki: Join actually ignores nil (no-op), so it's fine

robert-zaremba: but `errors.Join` will make unnecessary wrappers, so under the hood we will have `wrapper{ wrapper { wrapper {...} } }`

PR #2376, opened 2024-01-02
feat: v6.3 upgrade handler and release notes

robert-zaremba:
## Summary by CodeRabbit

- **New Features**
	- Added symbol name to the x/uibc QueryAllOutflowsResponse `outflows` entry.
	- Added new queries `uibc/MissCounters` and `uibc/Inflows`.
	- Addition of query `leverage/RegisteredTokenMarkets` to fetch Registered Tokens and their Market Summaries for frontend in fewer queries.

- **Improvements**
	- Modification of parameters in `uic/MsgGovUpdateQuota` to `uibc/MsgGovUpdateQuota`.
	- Addition of ICS 20 memo handler (disabled) in UIBC.
	- Adjusted IBC Quota v2 mechanism for better quota management.

- **Bug Fixes**
	- Fix in inflow amount calculation.

- **Documentation**
	- Updated CHANGELOG to reflect the latest changes and improvements.

PR #2379, opened 2024-01-04
fix(sdkclient): bring back the acc sequence setting fix

robert-zaremba:
## Description

In release PR we had to [remove the fix ](https://github.com/umee-network/umee/pull/2376/commits/3a0099d89f11f0d7ecc6a50fee1b1a0b1b8ec946) because it was breaking the [e2e tests](https://github.com/umee-network/umee/actions/runs/7395266399).

Here I'm readding the fix , and adding local test.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the account sequence setting mechanism for transaction processing.

- **Tests**
  - Introduced tests to validate account sequence settings in transactions.

- **Documentation**
  - Added comments to clarify future code deprecations.


tests/e2e/e2e_leverage_test.go
```diff
@@ -110,21 +112,28 @@ func (s *E2ETest) TestLeverageBasics() {
 			s.leverageSupply(0, appparams.BondDenom, 100_000_000)
 		},
 	)
+	time.Sleep(time.Second * 5)
```
gsk967, line 115: I added `sleep 5` seconds here for making sure previous txn will succed

gsk967 (approved):

robert-zaremba (commented):
Instead of doing `BroadcastTxWithRetry` (which we already have), we should increase the sequence number after the broadcast

PR #2396, opened 2024-01-19
feat(ics20_module): check sender address

robert-zaremba:
## Description

* Sender address check was missing. In fact , we need to verify both sender and receiver after initial tests.
* fix: we should firstly record the transfer, before handling the memo.


gsk967 (approved):
LGTM, utACK

PR #2439, opened 2024-02-27
fix: ics20 memo handler denom

robert-zaremba:
## Description

When receiving an ibc denom, we need to correctly transform the `FungibleTokenPacketData.Denom` into the IBC Denom, as it is recorded in the `x/leverage`.

## Summary by CodeRabbit

- **New Features**
    - Enhanced the handling of ICS20 token packets to automatically extract and manage denominations based on source and destination information.
- **Refactor**
    - Simplified the logic for recording IBC inflow by utilizing a new utility function for denomination extraction.
    - Updated method signatures in the ICS20Module and MemoHandler to support enhanced packet processing capabilities.


gsk967 (approved):
LGTM

PR #2442, opened 2024-02-28
feat(ibc): memo fallback address

robert-zaremba:
## Description

1. Try to unpack and prepare memo. If memo has a correct structure, and fallback addr is defined but malformed, we cancel the transfer (otherwise would not be able to use it correctly)
2. If memo has a correct structure, but memo.messages can't be unpack or don't pass validation,
   then we continue with the transfer and overwrite the original receiver to fallback_addr if
   it's defined.
1. Execute the downstream middleware and the transfer app.
2. Execute hooks. If hook execution fails, we don't use the the fallback_addr nor ignore the transfer.
   This is because there could be other middlewares that are already executed.

NOTE: I'm not very convinced about the last point - personally I would always revert the transfer if the memo hook structure is correct but we can't handle it (validation problem, message not supported, or hook execution error). What do you think?

## Summary by CodeRabbit

- **New Features**
    - Introduced a fallback recipient address option for enhanced transaction flexibility.
    - Added memo processing functionality, including receiver overwrite based on specific conditions.
- **Enhancements**
    - Improved message unpacking capabilities for better data handling.
    - Enhanced event logging and error reporting for more transparent operation feedback.
- **Refactor**
    - Optimized memo handling logic for increased efficiency and reliability.
- **Tests**
    - Expanded testing suite to cover new functionalities and ensure robustness of memo processing.


x/uibc/uics20/ibc_module.go
```diff
@@ -92,12 +117,28 @@ func (im ICS20Module) onAckErr(ctx *sdk.Context, packet channeltypes.Packet) {
...
 	}
 	qk := im.kb.Keeper(ctx)
 	qk.IBCRevertQuotaUpdate(ftData.Amount, ftData.Denom)
 }
 
+func (im ICS20Module) emitEvents(em *sdk.EventManager, logger log.Logger, topic string, events []string) {
+	attributes := make([]sdk.Attribute, len(events))
+	key := topic + "-context"
+	for i, s := range events {
+		attributes[i] = sdk.NewAttribute(key, s)
```
gsk967, line 130: here all the attributes will have same key 

key => val1 
key => val2 ....

gsk967 (approved):
utAck

x/uibc/uics20/ibc_module.go
robert-zaremba: It's OK, this is not a map, but list of (key, value) pairs. This was influenced how ibc-apps deal with events.

PR #2443, opened 2024-03-01
feat: handle fallback address with memo execution and adjust message coins

robert-zaremba:
## Description

1. The amount of IBC tokens we sent might be different than the amount we receive (eg operator can take fees). So, if a user will set a maximum amount (eg in MsgSupply) then we may get an error when executing the message. In that case we need to adjust the amount in the message.
2. chore: refactore ibc memo handling to unify GMP and standard hooks
3. support receiver fallback address in case hook execution fails.

## Summary by CodeRabbit

- **Refactor**
	- Renamed and updated methods in `MemoHandler` for improved memo parsing and preparation.
	- Refactored `ICS20Module` to handle memo validation errors and manage fallback receivers efficiently.
	- Improved test coverage in `memo_handler_test.go` for enhanced functionality testing.
- **New Features**
	- Added new fields and methods to `MemoHandler` for advanced memo handling capabilities.
- **Bug Fixes**
	- Addressed issues related to memo message validation and handling.

### DIFF
--- a/x/uibc/uics20/ibc_module.go
+++ b/x/uibc/uics20/ibc_module.go
@@ -68,15 +68,15 @@ func (im ICS20Module) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet, 
 	// no other middleware that can change packet data or amounts.
 
 	mh := MemoHandler{cdc: im.cdc, leverage: im.leverage}
-	fallbackReceiver, events, err := mh.onRecvPacketPrepare(&ctx, packet, ftData)
+	events, err := mh.onRecvPacketPrepare(&ctx, packet, ftData)
 	if err != nil && err != errMemoValidation {
 		return channeltypes.NewErrorAcknowledgement(err)
 	}
 	var transferCtx = ctx
 	var ctxFlush func()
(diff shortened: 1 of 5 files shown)


x/uibc/uics20/ibc_module.go
```diff
@@ -67,32 +67,56 @@ func (im ICS20Module) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet,
...
 		}
 	}
+	execCtx, execCtxFlush := transferCtx.CacheContext()
 
 	// call transfer module app
-	ack := im.IBCModule.OnRecvPacket(ctx, packet, relayer)
+	ack := im.IBCModule.OnRecvPacket(transferCtx, packet, relayer)
 	if !ack.Success() {
-		return ack
+		goto end
```
gsk967, line 95: why we need `goto` functions , we can write here only

gsk967 (approved):
utACK

x/uibc/uics20/ibc_module.go
robert-zaremba: we can, but it will introduce copy-paste

PR #2480, opened 2024-03-28
chore(setup): add editorconfig

robert-zaremba:
## Summary by CodeRabbit

- **Chores**
	- Updated coding style rules to enhance code consistency across different file types.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,42 @@
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+# indent_style = space
+# indent_size = 4
+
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,42 @@
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+# indent_style = space
+# indent_size = 4
```
gsk967, line 9: can you enable this, so everyone will see same tab size

gsk967 (approved):

.editorconfig
robert-zaremba: I think we should leave default for language specific config

PR #2517, opened 2024-05-06
feature(auction) genesis

robert-zaremba:


x/auction/genesis.go
```diff
@@ -1,12 +1,43 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
 // DefaultGenesis creates a default genesis state
 func DefaultGenesis() *GenesisState {
-	// TODO
-	return &GenesisState{}
+	return &GenesisState{
+		RewardsParams:   RewardsParams{BidDuration: 14 * 24 * 3600}, // 14 days
```
gsk967, line 13:
```suggestion
		RewardsParams:   RewardsParams{BidDuration:  time.Second * 14 * 24 * 3600}, // 14 days
```

gsk967 (approved):
utACK, 
Small suggestion , can you add 
```go
func nextRewardsID(currentId int) int {
...
}

```

x/auction/genesis.go
robert-zaremba: This is correct, see the params docs: 
> // bid_duration is duration of the bid phase in seconds.

gsk967: Okay

Issue #2795, opened 2026-03-28
Proposal: Bad Debt Isolation & Repayment Mechanism


umee-man:
# Proposal: Bad Debt Isolation & Repayment Mechanism

> [!CAUTION]
> **The Problem:** In the current `x/leverage` architecture, when a user's collateral is fully liquidated but borrow balances remain, the remaining debt is flagged as "Bad Debt". However, this debt remains tracked as standard borrow shares. Because interest in Umee is calculated globally via an `InterestScalar` acting on all borrow shares, **bad debt continues to accrue interest indefinitely.** This infinitely increases the protocol's deficit and artificially inflates the utilization rate, driving up Borrow APY for healthy users.

## 1. Executive Summary

This proposal suggests an architectural shift to isolate bad debt. By completely removing uncollateralized debt from the standard borrow tracking system and moving it to a static, global ledger, the protocol can freeze the debt size (stopping interest accrual) while continuing to pay it down programmatically from newly accrued protocol reserves. 

## 2. Proposed Architecture

The solution requires two primary state transitions: isolating the debt to freeze its value, and updating the global sweep mechanism to pay it down efficiently.

### A. The Isolation Mechanism (Freezing the Debt)

Currently, the liquidation hook `postLiquidate` flags an `address|denom` when collateral hits zero. 

**Proposed Change:**
Instead of just flagging the address, the protocol should execute a **debt write-off** during the liquidation state transition:
1. **Calculate Final Debt:** Convert the user's remaining borrow shares into a static base token amount (e.g., `100 USDC`).
2. **Erase the Position:** Delete the user's borrow position entirely so they no longer hold borrow shares.
3. **Decrease Total Borrowed:** Subtract the static amount from the leverage module's `TotalBorrowed` state. *This immediately fixes the utilization rate calculation, ensuring healthy borrowers are not penalized with high interest rates caused by dead liquidity.*
4. **Move to Isolated State:** Add the static amount to a new global variable: `IsolatedBadDebt[denom]`.

### B. Modified Reserve Sweep Procedure (Automatic Repayment)

Currently, the `SweepBadDebts` function (called in `EndBlocker`) iterates over all flagged `address|denom` pairs across the network, attempting to repay each individually. This is highly inefficient.

**Proposed Change:**
The `SweepBadDebts` function should be refactored to operate globally rather than individually:
1. In `EndBlocker`, the protocol checks if `IsolatedBadDebt[denom] > 0`.
[...]

PR #2799, opened 2026-04-10
moving bad debts to isolate bad debts

gsk967:
## Description
closes: #2795 
moving bad debts to isolate bad debts
1. add tx cmd - `repay-isolated-bad-debt`
2. add cli query cmd - `isolated-bad-debts`, `last-interest-time` , `interest-scalars`

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


proto/umee/leverage/v1/genesis.proto
```diff
@@ -26,7 +26,26 @@ message GenesisState {
...
+    (gogoproto.nullable)   = false
+  ];
+  string interest_scalar = 4 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
+    (gogoproto.nullable)   = false
+  ];
+  string interest_accrued = 5 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+    (gogoproto.nullable)   = false
+  ];
```
toteki, line 48: I recommend adding comments to clarify these fields

For examples: to indicate if amout is before or after scalar is applied; if scalar stored here is the value of the scalar at the moment the debt was isolated or some other value; and if interest accrued is part of the amount field, separate from it, and before or after scalar.

x/leverage/keeper/isolated_bad_debts.go
```diff
@@ -0,0 +1,206 @@
...
+		interestScalar := k.getInterestScalar(ctx, borrowDenom)
+		// interestScalar always >=1
+		interestAccured := adjustedAmount.Mul(interestScalar.Sub(sdk.OneDec())).Ceil().TruncateInt()
+		b := types.NewBorrowWithInterest(borrowDenom, adjustedAmount.Ceil().TruncateInt(), interestScalar, interestAccured)
+		totalBorrowed = append(totalBorrowed, b)
+		return nil
+	}
+	util.Panic(k.iterate(ctx, prefix, iterator))
+	return totalBorrowed
+}
```
toteki, line 100: This function is where the danger is.

For context, `InterestScalar` for a given `denom` starts at `1.0` the first time that denom is registered with the leverage module, and then begins increasing as borrowers accrue interest. New borrows created after that do not start at `InterestScalar=1.0` but at a higher value, even before they accrue interest.

Therefore, this function incorrectly separates principal and interest.

> Example scenario:
>
> On `Feb 1`, we register denom `CoinA` with the leverage module
> - At that moment, `InterestScalar(CoinA) = 1.0`
>
> Unspecified borrowing and lending activity happens between `Feb 1 and Jul 31`
> - Due to borrow interest, `InterestScalar(CoinA) = 1.2` now
>
> On `Aug 1`, `borrower Z` borrows `180 CoinA`
> - This is happening at a time when `InterestScalar(CoinA) = 1.2`
> - At this instant, the leverage module stores `adjustedAmount=150` because 150*1.2 = `180 CoinA`
>
> Unspecified borrowing and lending activity happens between `Aug 1 and Dec 31`
> - Due to borrow interest, `InterestScalar(CoinA) = 1.4` now
> - The leverage module still stores `adjustedAmount=150` assuming `borrower Z` did nothing else
> - Therefore the current debt owed is 150*1.4 = `210 CoinA`. Principal is 180 and interest is 30.
>
> If we applied this `GetBorrowerBorrowsWithoutInterest` to `borrower Z` on `Dec 31` it would incorrectly state that principal is 150 and interest is 60.

toteki (changes requested):
See the problem with interest scalar.

We would need to crawl the chain with an archive node to separate interest from principal accurately. A migration couldn't do that.

We should clarify that what we're storing here is a _snapshot_ of the bad debts with the current interest scalar so it stops increasing for them. The real values of `(principal,interest)` will be `(snapshotPrincipal+X,snapshotInterest-X)` for some `X >= 0` for each borrower.

